PR #1, opened 2017-10-26
Add cli and run Etcd and Control Planes

galal-hussein:


cmd/cluster.go
```diff
@@ -0,0 +1,133 @@
...
+		k8shosts[i].TunnelUp(ctx)
+		defer k8shosts[i].DClient.Close()
+		if err != nil {
+			return err
+		}
+	}
+	etcdHosts, cpHosts, workerHosts := hosts.DivideHosts(k8shosts)
+	err = services.RunEtcdPlane(etcdHosts, servicesLookup.Services.Etcd)
+	if err != nil {
+		logrus.Errorf("[Etcd] Failed to bring up Etcd Plane: %v", err)
```
alena1108: same here. Do not log error

cmd/cluster.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/rancher/rke/hosts"
+	"github.com/rancher/rke/services"
+	"github.com/urfave/cli"
+)
+
+func ClusterCommand() cli.Command {
+	clusterUpFlags := []cli.Flag{
+		cli.StringFlag{
+			Name:   "cluster-file",
+			Usage:  "Specify an alternate cluster YAML file (default: cluster.yml)",
```
alena1108: You can set the default value directly in cli.flag, no need to hardcode

cmd/cluster.go
```diff
@@ -0,0 +1,133 @@
...
+	buf, err := ioutil.ReadAll(file)
+	if err != nil {
+		return "", fmt.Errorf("failed to read file: %v", err)
+	}
+	clusterFile = string(buf)
+
+	return clusterFile, nil
+}
+
+func parseClusterFile(clusterFile string) (services.Container, []hosts.Host, error) {
```
alena1108: Any reason we can't return a pointer to a services.Container?

cmd/cluster.go
galal-hussein: done

services/services.go
```diff
@@ -0,0 +1,68 @@
...
+	"io/ioutil"
+	"os"
+
+	"github.com/Sirupsen/logrus"
+	"github.com/docker/docker/api/types"
+	"github.com/rancher/rke/hosts"
+	"golang.org/x/net/context"
+)
+
+type Container struct {
```
alena1108: Why do we need this wrapper struct?

galal-hussein: because we need a container struct with name `yaml: services` to contain each service type

services/etcd.go
```diff
@@ -0,0 +1,106 @@
...
+		if isRunning {
+			logrus.Infof("[Etcd] Container is already running on host [%s]", host.Hostname)
+			return nil
+		}
+		err = runEtcdContainer(host, etcdService)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
```
alena1108: We should add the log on success for each of the planes

services/etcd.go
```diff
@@ -0,0 +1,106 @@
...
+	err := PullImage(host, etcdService.Image+":"+etcdService.Version)
+	if err != nil {
+		return err
+	}
+	logrus.Infof("[Etcd] Successfully pulled Etcd image on host [%s]", host.Hostname)
+	err = doRunEtcd(host, etcdService)
+	if err != nil {
+		return err
+	}
+	logrus.Infof("[Etcd] Successfully ran Etcd container on host [%s]", host.Hostname)
```
alena1108: Successfully started instead of successfully ran.

services/kubelet.go
```diff
@@ -0,0 +1,114 @@
...
+
+type Kubelet struct {
+	Version string `yaml:"version"`
+	Image   string `yaml:"image"`
+	ClusterDomain	string `yaml:"cluster_domain"`
+	InfraContainerImage string `yaml:"infra_container_image"`
+}
+
+func runKubelet(host hosts.Host, masterHost hosts.Host, kubeletService Kubelet, isMaster bool) error {
+	isRunning, err := IsContainerRunning(host, KubeletContainerName)
```
alena1108: It feels like the code IsContainerRunning/PullImage/doRun is being repeated in each component. See how you can make it more generic and reusable (including the logs).

alena1108: Only building the config is really component specific

services/services.go
```diff
@@ -0,0 +1,68 @@
...
+	WorkerRole                  = "worker"
+	KubeAPIContainerName        = "kube-api"
+	KubeletContainerName        = "kubelet"
+	KubeproxyContainerName      = "kube-proxy"
+	KubeControllerContainerName = "kube-controller"
+	SchedulerContainerName      = "scheduler"
+	EtcdContainerName           = "etcd"
+)
+
+func IsContainerRunning(host hosts.Host, containerName string) (bool, error) {
```
alena1108: I'd separate docker specific checks to some kind of docker utils class.

services/worker.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"github.com/Sirupsen/logrus"
+	"github.com/rancher/rke/hosts"
+)
+
+func RunWorkerPlane(masterHosts []hosts.Host, workerHosts []hosts.Host, workerServices Services) error {
+	logrus.Infof("[WorkerPlane] Building up Worker Plane..")
+	for _, host := range masterHosts {
+		// only one master for now
+		err := runKubelet(host, masterHosts[0], workerServices.Kubelet, true)
```
alena1108: Why do we run kubelet on master hosts? Isn't it supposed to run on worker hosts only?

galal-hussein: No kubelet need to run on master hosts as well, however it should be in non schedulable state, the node still should be monitored by kubernetes

cluster.yml
```diff
@@ -0,0 +1,38 @@
...
+services:
+  etcd:
+    version: latest
+    image: quay.io/coreos/etcd
+  kube-api:
+    version: v1.7.5_coreos.0
+    image: quay.io/coreos/hyperkube
+    service_cluster_ip_range: 10.233.0.0/18
+  kube-controller:
+    version: v1.7.5_coreos.0
```
alena1108: the version should be a part of image

cluster/certificates.go
```diff
@@ -0,0 +1,118 @@
...
+	"k8s.io/client-go/util/cert"
+)
+
+// SaveClusterCerts ...
+func SaveClusterCerts(crts map[string]pki.CertificatePKI) error {
+	logrus.Infof("[certificates] Save kubernetes certificates as secrets")
+	for crtName, crt := range crts {
+		err := saveCrtToKubernetes(crtName, crt)
+		if err != nil {
+			return fmt.Errorf("Failed to save certificate [%s] to kubernetes", crtName)
```
alena1108: We should return the exact error returned by saveCrtToKubernetes, along with the "Failed to save certificate" message

cluster/certificates.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+	}
+	logrus.Infof("[certificates] Successfuly saved certificates as kubernetes secret [%s]", pki.CertificatesSecretName)
+	return nil
+}
+
+func GetClusterCerts(kubeConfigPath string) (map[string]pki.CertificatePKI, error) {
+	logrus.Infof("[certificates] Getting Cluster certificates from Kubernetes")
+	k8sClient, err := k8s.NewClient(pki.KubeAdminConfigPath)
```
alena1108: Are we creating a k8s client for each an every k8s command we issue? It shoudl be done just once, and shared later on

pki/deploy.go
```diff
@@ -0,0 +1,142 @@
...
+	return name + "=" + string(encodedCrt)
+}
+
+func convertKeyToENV(name string, key *rsa.PrivateKey) string {
+	encodedKey := cert.EncodePrivateKeyPEM(key)
+	return name + "=" + string(encodedKey)
+}
+
+func convertConfigToENV(name string, config string) string {
+	return name + "=" + config
```
alena1108: Better use fmt.sprintf to concat strings

cluster/certificates.go
```diff
@@ -0,0 +1,110 @@
...
+			Key:         secretKey.(*rsa.PrivateKey),
+			Config:      secretConfig,
+		}
+	}
+	logrus.Infof("[certificates] Successfully fetched Cluster certificates from Kubernetes")
+	return certMap, nil
+}
+
+func DeployClusterCerts(kubeHosts []hosts.Host, clusterCerts map[string]pki.CertificatePKI, forceDeploy bool) error {
+	logrus.Infof("[certificates] Deploying kubernetes certificates to Cluster nodes")
```
alena1108: We should consider that there can be diff auth providers, and may be:

* create a generic method, something like "saveConfigToNode".
* Deploying admin config will be generic
* Saving certs (or other type of auth) will be based on the auth provider chosen

alena1108: This generic method should be in a dedicated file; each auth component will be in its separate file.

cmd/cluster.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+}
+
+var clusterCerts map[string]pki.CertificatePKI
+
+func clusterUp(ctx *cli.Context) error {
+	logrus.Infof("Building up Kubernetes cluster")
+	clusterFile, err := resolveClusterFile(ctx)
+	if err != nil {
+		return fmt.Errorf("Failed to bring cluster up: %v", err)
```
alena1108: Should the error be "Failed to resolve cluster file"?

cmd/cluster.go
```diff
@@ -0,0 +1,186 @@
...
+	kubeCluster, err := parseClusterFile(clusterFile)
+	if err != nil {
+		return fmt.Errorf("Failed to parse the cluster file: %v", err)
+	}
+	for i := range kubeCluster.Hosts {
+		// Set up socket tunneling
+		kubeCluster.Hosts[i].TunnelUp(ctx)
+		defer kubeCluster.Hosts[i].DClient.Close()
+		if err != nil {
+			return err
```
alena1108: More description to the error indicating that setting tunneling had failed.

hosts/hosts.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func DivideHosts(hosts []Host) ([]Host, []Host, []Host) {
+	etcdHosts := []Host{}
+	cpHosts := []Host{}
+	workerHosts := []Host{}
+	for _, host := range hosts {
+		for _, role := range host.Role {
+			logrus.Debugf("Host: " + host.Hostname + " has role: " + role)
+			if role == "etcd" {
```
alena1108: I believe you had a constants for the roles defined, better to use them.

hosts/hosts.go
```diff
@@ -0,0 +1,76 @@
...
+				etcdHosts = append(etcdHosts, host)
+			}
+			if role == "controlplane" {
```
alena1108: Should be either elseif, or use switch statement

hosts/hosts.go
```diff
@@ -0,0 +1,76 @@
...
+				cpHosts = append(cpHosts, host)
+			}
+			if role == "worker" {
+				workerHosts = append(workerHosts, host)
+			}
```
alena1108: If role is unrecognizable, log it

cluster/cluster.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/rancher/rke/hosts"
+	"github.com/rancher/rke/k8s"
+	"github.com/rancher/rke/services"
+	yaml "gopkg.in/yaml.v2"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/pkg/api/v1"
+)
+
+type Cluster struct {
+	Services services.Services `yaml:"services"`
```
alena1108: Lets think about defining all core structs in a separate file. This way it will be easier to refer to them. Also I see that you are calling divideHosts logic from many places. A config file passed by user, should be parsed and transformed to cluster data structure just once. If you think that there are some extra data structures are needed by your code - etcdHosts/controlPlaneHosts/etc - introduce them

cmd/cluster.go
```diff
@@ -0,0 +1,186 @@
...
+				Action: clusterUp,
+				Flags:  clusterUpFlags,
+			},
+		},
+	}
+}
+
+var clusterCerts map[string]pki.CertificatePKI
+
+func clusterUp(ctx *cli.Context) error {
```
alena1108: this method should return clusterState

PR #2, opened 2017-11-06
add network plugin support

moelsayed:


package/kubectl/entrypoint.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash -x
+
+SSL_CRTS_DIR=/etc/kubernetes/ssl
+mkdir -p $SSL_CRTS_DIR
+
+for i in $(env | grep -o KUBE_.*=); do
```
galal-hussein: i don't think this part is needed because the admin kubeconfig should include the certificate data

cluster/network.go
```diff
@@ -0,0 +1,59 @@
...
+func (c *Cluster) DeployNetworkPlugin() error {
+	deployerHost := c.ControlPlaneHosts[0]
+	logrus.Debugf("[Network] Using host [%s] for deployment", deployerHost.Hostname)
+	logrus.Debugf("[Network] Pulling network-deployer image..")
+	if err := docker.PullImage(deployerHost.DClient, deployerHost.Hostname, k8s.KubectlImage); err != nil {
+		return err
+	}
+	logrus.Debugf("[Network] Setting network plugin: %s", c.NetworkPlugin)
+	env := []string{
+		fmt.Sprintf("%s=%s", "KUBECFG_NET_DEPLOY", c.Certificates[pki.KubeAdminCommonName].Config),
```
galal-hussein: can you separate the kubectl run in a separate function like (RunKubectl) in k8s package that can receive the cmd string as in `/tmp/entrypoint.sh apply -f /network/ ....` which should be more generic and just call this function from the deploy network plugin

cluster/network.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"fmt"
+
+	"github.com/Sirupsen/logrus"
+	"github.com/rancher/rke/k8s"
+	"github.com/rancher/rke/pki"
+)
+
+func (c *Cluster) DeployNetworkPlugin() error {
+	logrus.Debugf("[Network] Setting network plugin: %s", c.NetworkPlugin)
```
alena1108: we should log it in info. And log successful/error update

cluster/network.go
```diff
@@ -0,0 +1,23 @@
...
+	deployerHost := c.ControlPlaneHosts[0]
```
alena1108: We should add a guard for validation the size of the control plane hosts. What if there are none? Up to you when you add it, may be in the place where the cluster.yml file is being parsed?

cluster/network.go
```diff
@@ -0,0 +1,23 @@
...
+	kubectlCmd := []string{"apply -f /network/" + c.NetworkPlugin + ".yaml"}
```
alena1108: Looks very hacky to me, the way we build the commands...Consider for example that there can be multiple files in the future, multiple services (dns). Lets think of more generic way of deploying all the addons. May be similar to kubernetes-package?

k8s/k8s.go
```diff
@@ -88,3 +101,47 @@ func GetSecret(k8sClient *kubernetes.Clientset, secretName string) (*v1.Secret,
...
+	resp, err := dClient.ContainerCreate(context.Background(), imageCfg, nil, nil, KubctlContainer)
+	if err != nil {
+		return fmt.Errorf("Failed to create kubectl container on host [%s]: %v", hostname, err)
+	}
+	logrus.Debugf("[Kubectl] Container %s created..", resp.ID)
+	if err := dClient.ContainerStart(context.Background(), resp.ID, types.ContainerStartOptions{}); err != nil {
+		return fmt.Errorf("Failed to start kubectl container on host [%s]: %v", hostname, err)
+	}
+	for {
+		isDeployerRunning, err := docker.IsContainerRunning(dClient, hostname, KubctlContainer)
```
alena1108: Can we explore other ways of checking if the container has done its job? Not only the container should stop, we should check the error code.

alena1108: Like subscribing to the container event. Also there should be a timeout so we don't end up waiting on things forever

cluster/network.go
moelsayed: The intention here is to create a general use kubectl container. It can be used to deploy any manifest on we add later or to run ad-hoc commands.

cluster/network.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"fmt"
+
+	"github.com/Sirupsen/logrus"
+	"github.com/rancher/rke/k8s"
+	"github.com/rancher/rke/pki"
+)
+
+func (c *Cluster) DeployNetworkPlugin() error {
+	logrus.Infof("[Network] Setting network plugin: %s", c.NetworkPlugin)
```
alena1108: I thought we we gonna handle addon creation in more generic manner? Imagine the same command will be called on the dns setup as well

cmd/cluster.go
```diff
@@ -77,6 +77,10 @@ func ClusterUp(clusterFile, authType string) (string, string, string, string, er
 		return ApiURL, caCrt, clientCert, clientKey, err
 	}
 
+	err = kubeCluster.DeployNetworkPlugin()
```
alena1108: Should be generic call to deploy the addons here. Then inside the addons deploy code you can distinguish between the plugins types.

cluster/network.go
moelsayed: No, addons will be added in a separate call.

cmd/cluster.go
moelsayed: I added env variable configuration. network plugin name will be passed in the cluster yaml file and we can add env variables add more plugins. We will also update this as we add plugins if needed.

cluster/addons.go
```diff
@@ -0,0 +1,42 @@
...
+
+func (c *Cluster) DeployK8sAddOns() error {
+	if err := c.deployKubeDNS(); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (c *Cluster) deployKubeDNS() error {
+	logrus.Infof("[Plugins] Setting up KubeDNS")
```
galal-hussein: can we change name Plugins to addons and to lowercase to be consistent with other log format in the code

cluster/network.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/rancher/rke/k8s"
+	"github.com/rancher/rke/pki"
+)
+
+const (
+	ClusterCIDREnvName = "RKE_CLUSTER_CIDR"
+)
+
+func (c *Cluster) DeployNetworkPlugin() error {
+	logrus.Infof("[Network] Setting up network plugin: %s", c.NetworkPlugin)
```
galal-hussein: same here, can this be lowercase

PR #3, opened 2017-11-07
unit tests and go lint fixes

galal-hussein:


alena1108:
We might think about adding tests for:

* validating that there is at least one host defined for etcd and control plane
* validating etcd scale

galal-hussein:
@alena1108 ^

PR #15, opened 2017-11-15
Add cluster upgrade

moelsayed:


cmd/cluster.go
```diff
@@ -136,3 +149,49 @@ func getClusterVersion(ctx *cli.Context) error {
...
+	err = kubeCluster.ClusterUpgrade()
+	if err != nil {
+		return APIURL, caCrt, clientCert, clientKey, err
+	}
+
+	APIURL = fmt.Sprintf("https://" + kubeCluster.ControlPlaneHosts[0].IP + ":6443")
+	caCrt = string(cert.EncodeCertPEM(kubeCluster.Certificates[pki.CACertName].Certificate))
+	clientCert = string(cert.EncodeCertPEM(kubeCluster.Certificates[pki.KubeAdminCommonName].Certificate))
+	clientKey = string(cert.EncodePrivateKeyPEM(kubeCluster.Certificates[pki.KubeAdminCommonName].Key))
+	return APIURL, caCrt, clientCert, clientKey, nil
```
galal-hussein: Can you add a log to indicate that upgrade succeeded

cmd/cluster.go
```diff
@@ -136,3 +149,49 @@ func getClusterVersion(ctx *cli.Context) error {
...
+		kubeCluster is the cluster.yaml definition. It should have updated configuration
+		currentCluster is the current state fetched from kubernetes
+		we add currentCluster certs to kubeCluster, kubeCluster would have the latest configuration from cluster.yaml and the certs to connect to k8s and apply the upgrade
+	*/
+	kubeCluster.Certificates = currentCluster.Certificates
+	err = kubeCluster.TunnelHosts()
+	if err != nil {
+		return APIURL, caCrt, clientCert, clientKey, err
+	}
+	logrus.Infof("Starting cluster upgrade")
```
galal-hussein: Can this be a debug info

k8s/node.go
```diff
@@ -0,0 +1,54 @@
...
+
+func DeleteNode(k8sClient *kubernetes.Clientset, nodeName string) error {
+	return k8sClient.CoreV1().Nodes().Delete(nodeName, &metav1.DeleteOptions{})
+}
+
+func GetNodeList(k8sClient *kubernetes.Clientset) (*v1.NodeList, error) {
+	return k8sClient.CoreV1().Nodes().List(metav1.ListOptions{})
+}
+
+func CordonUncordon(k8sClient *kubernetes.Clientset, nodeName string, cordoned bool) error {
```
galal-hussein: I don't think you are using this function anywhere

services/kubeapi.go
```diff
@@ -17,6 +18,33 @@ func runKubeAPI(host hosts.Host, etcdHosts []hosts.Host, kubeAPIService v1.KubeA
 	return docker.DoRunContainer(host.DClient, imageCfg, hostCfg, KubeAPIContainerName, host.AdvertisedHostname, ControlRole)
 }
 
+func upgradeKubeAPI(host hosts.Host, etcdHosts []hosts.Host, kubeAPIService v1.KubeAPIService) error {
+	logrus.Debugf("[upgrade/KubeAPI] Checking for deployed version")
+	containerInspect, err := docker.InspectContainer(host.DClient, host.AdvertisedHostname, KubeAPIContainerName)
```
galal-hussein: I am wondering if you can use some common function to include inspect/stopping/deploying/removing container for each service and just call it from inside upgrade* functions instead of repeating the code in each service

cmd/cluster.go
moelsayed: We have a similar one for the `up` command. Added this for consistency.

services/kubeapi.go
moelsayed: it would be clunky and would take a lot of parameters due to the way we have things structured..There are some code duplication here but easier to manage for now.

cmd/cluster.go
moelsayed: done.

k8s/node.go
moelsayed: Fixed, used for for worker upgrades.

PR #22, opened 2017-11-18
System and user addons based on k8s jobs and configmaps

moelsayed:


cluster/addons.go
```diff
@@ -2,32 +2,132 @@ package cluster
...
+
+	if err := k8s.ApplyK8sSystemJob(addonJob, c.LocalKubeConfigPath); err != nil {
+		fmt.Println(err)
+		return err
 	}
-	logrus.Infof("[plugins] kubeDNS deployed successfully..")
 	return nil
+}
 
+func (c *Cluster) GetAddonsExcuteJob(addonName string) string {
```
galal-hussein: Can this function be moved to addons package to have all the manifest functions in one package

cluster/addons.go
```diff
@@ -2,32 +2,132 @@ package cluster
...
+	timeout := make(chan bool, 1)
+	go func() {
+		for {
+			err := k8s.UpdateConfigMap(kubeClient, []byte(addonYaml), addonName)
+			if err != nil {
+				time.Sleep(time.Second * 5)
+				fmt.Println(err)
+				continue
+			}
+			logrus.Infof("[addons] Successfully Saved addon to	 Kubernetes ConfigMap: %s", addonName)
```
galal-hussein: I think this is an editor problem that added an extra tab to log info, can you remove that

k8s/job.go
```diff
@@ -0,0 +1,59 @@
...
+				logrus.Debugf("[k8s] Job %s completed successfully..", job.Name)
+				return nil
+			}
+		}
+
+	}
+	return fmt.Errorf("Failed to get job complete status: %v", err)
+}
+
+func ApplyK8sSystemJob(jobYaml, kubeConfigPath string) error {
```
galal-hussein: Can this function be removed since its not doing extra work than applyK8sJob

PR #23, opened 2017-11-20
Cluster Down/Remove

galal-hussein:
* Remove services' docker containers
* Clean up static mounted directories [/etc/kubernetes/ssl, /etc/cni, /opt/cni, /var/lib/etcd]

### DIFF
--- a/cluster/down.go
+++ b/cluster/down.go
@@ -1,0 +1,43 @@
+package cluster
+
+import (
+	"github.com/rancher/rke/hosts"
+	"github.com/rancher/rke/services"
+)
+
+func (c *Cluster) ClusterDown() error {
+	// Remove Worker Plane
+	if err := services.RemoveWorkerPlane(c.ControlPlaneHosts, c.WorkerHosts); err != nil {
+		return err
+	}
+
+	// Remove Contol Plane
+	if err := services.RemoveControlPlane(c.ControlPlaneHosts); err != nil {
+		return err
+	}
+
+	// Remove Etcd Plane
+	if err := services.RemoveEtcdPlane(c.EtcdHosts); err != nil {
+		return err
+	}
+
+	// Clean up all hosts
+	if err := cleanUpHosts(c.ControlPlaneHosts, c.WorkerHosts, c.EtcdHosts); err != nil {
+		return err
+	}
+	return nil
+}
+
+func cleanUpHosts(cpHosts, workerHosts, etcdHosts []hosts.Host) error {
+	allHosts := []hosts.Host{}
+	allHosts = append(allHosts, cpHosts...)
+	allHosts = append(allHosts, workerHosts...)
+	allHosts = append(allHosts, etcdHosts...)
+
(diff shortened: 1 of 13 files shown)


cluster/down.go
```diff
@@ -0,0 +1,43 @@
+package cluster
+
+import (
+	"github.com/rancher/rke/hosts"
+	"github.com/rancher/rke/services"
+)
+
+func (c *Cluster) ClusterDown() error {
```
moelsayed: We probably should tear down stuff in the revers order they were created.

cmd/cluster.go
```diff
@@ -14,45 +14,43 @@ import (
...
 		Subcommands: []cli.Command{
 			cli.Command{
 				Name:   "up",
 				Usage:  "Bring the cluster up",
 				Action: clusterUpFromCli,
-				Flags:  clusterUpFlags,
+				Flags:  clusterUpDownUpgradeFlags,
+			},
+			cli.Command{
+				Name:   "down",
```
moelsayed: We should ask the user for confirmation before removal..We also should provide a --force option to skip the confirmation and allow for automation and shoot-self-in-the-foot needs..

docker/docker.go
```diff
@@ -69,6 +69,32 @@ func DoRollingUpdateContainer(dClient *client.Client, imageCfg *container.Config
 	return err
 }
 
+func DoRemoveContainer(dClient *client.Client, containerName, hostname string) error {
+	logrus.Infof("[down/%s] Checking if container is running on host [%s]", containerName, hostname)
+	// not using the wrapper to check if the error is a NotFound error
+	_, err := dClient.ContainerInspect(context.Background(), containerName)
```
moelsayed: Why not use IsContainerRunning() ? or at least  InspectContainer() ?

services/kubeapi.go
```diff
@@ -43,6 +43,10 @@ func upgradeKubeAPI(host hosts.Host, etcdHosts []hosts.Host, kubeAPIService v1.K
 	return err
 }
 
+func removeKubeAPI(host hosts.Host) error {
+	return docker.DoRemoveContainer(host.DClient, KubeAPIContainerName, host.AdvertisedHostname)
```
moelsayed: Add logging to this and the similar ones ? Like you have with etcd ?

galal-hussein: Actually i do in https://github.com/rancher/rke/pull/23/files#diff-a1440968342b4da00653d6c74b21f48eR56 and in workerplane file

docker/docker.go
galal-hussein: This is because i need to inspect the error to make sure its not a notfound error

cluster/down.go
galal-hussein: Done

cmd/cluster.go
galal-hussein: I added force flag

moelsayed:
LGTM.

PR #33, opened 2017-11-22
Minor cluster.yaml updates

moelsayed:


cluster.yml
```diff
@@ -5,8 +5,12 @@ auth:
   options:
     foo: bar
 
+# supported plugins are:
+# flannel
+# calico
+# canal
 network:
-  plugin:
+  plugin flannel
```
galal-hussein: there is `:` missing

Issue #53, opened 2017-11-28
Should be allowed to create a k8s cluster by specifying only image parameter for all the service containers.


sangeethah:
Should be allowed to create a k8s cluster by specifying only image parameter for all the service containers.

sangeethah:
rke version v0.0.6-dev

Able to create K8s cluster successfully by using only the following parameters in services section of cluster.yml
```
services:
  etcd:
    image: quay.io/coreos/etcd:latest
  kube-api:
    image: rancher/k8s:v1.8.3-rancher2
  kube-controller:
    image: rancher/k8s:v1.8.3-rancher2
  scheduler:
    image: rancher/k8s:v1.8.3-rancher2
  kubelet:
    image: rancher/k8s:v1.8.3-rancher2
  kubeproxy:
    image: rancher/k8s:v1.8.3-rancher2
```

Issue #52, opened 2017-11-27
Provide relevant error message when unsupported auth strategy and network plugins are used in cluster.yml file 


soumyalj:
1. Create a cluster with one master [controlplane, etcd] node and the other worker node using cluster.yml file. Use the command ./rke_darwin-amd64 cluster up --cluster-file cluster.yml
Cluster.yml file example:
```
auth:
  strategy: x200
network:
  plugin: abcd
hosts:
  - advertised_hostname: server1
    ip: 1.1.1.1
    user: ubuntu
    role: [controlplane, etcd]
    docker_socket: /var/run/docker.sock
    advertise_address: 10.1.1.1
  - advertised_hostname: server2
    ip: 2.2.2.2
    user: ubuntu
    role: [worker]
    advertise_address: 10.2.2.2

services:
  etcd:
    image: quay.io/coreos/etcd:latest
  kube-api:
    image: quay.io/coreos/hyperkube:v1.7.5_coreos.0
    service_cluster_ip_range: 10.233.0.0/18
    extra_args:
      v: 4
  kube-controller:
    image: quay.io/coreos/hyperkube:v1.7.5_coreos.0
    cluster_cidr: 10.233.64.0/18
    service_cluster_ip_range: 10.233.0.0/18
  scheduler:
    image: quay.io/coreos/hyperkube:v1.7.5_coreos.0
  kubelet:
    image: quay.io/coreos/hyperkube:v1.7.5_coreos.0
    cluster_domain: cluster.local
    cluster_dns_server: 10.233.0.3
    infra_container_image: gcr.io/google_containers/pause-amd64:3.0
  kubeproxy:
    image: quay.io/coreos/hyperkube:v1.7.5_coreos.0
```
Cluster creation fails but does not provide any indication that the fields are unsupported/invalid.
It fails with the below error:
```
soumyas-MBP:rke soumya$ ./rke_darwin-amd64 cluster up --cluster-file clusterdigworking.yml
INFO[0000] Building Kubernetes cluster                  
INFO[0000] [ssh] Checking private key                   
INFO[0000] [ssh] Start tunnel for host [rketest1] 
INFO[0000] [ssh] Start tunnel for host [rketest1] 
INFO[0000] [ssh] Start tunnel for host [rketes2]  
INFO[0000] [state] Found local kube config file, trying to get state from cluster 
INFO[0000] [state] Fetching cluster state from Kubernetes 
INFO[0030] Timed out waiting for kubernetes cluster to get state 
[...]

Issue #50, opened 2017-11-27
Host ip should be able to accept fqdn names that are resolvable.


sangeethah:
rke version v0.0.5-dev

When host ip is provided as  fqdn name that is resolvable ,certificate generation has problems.

```
Sangeethas-MBP:~ sangeethahariharan1$ ./rke_darwin-amd64 cluster up
INFO[0000] Building Kubernetes cluster                  
INFO[0000] [ssh] Checking private key                   
INFO[0000] [ssh] Start tunnel for host [ip-172-31-38-78.us-east-2.compute.internal] 
INFO[0000] [ssh] Start tunnel for host [ip-172-31-39-84.us-east-2.compute.internal] 
INFO[0000] [ssh] Start tunnel for host [ip-172-31-37-222.us-east-2.compute.internal] 
INFO[0000] [ssh] Start tunnel for host [ip-172-31-38-78.us-east-2.compute.internal] 
INFO[0000] [ssh] Start tunnel for host [ip-172-31-39-84.us-east-2.compute.internal] 
INFO[0000] [state] Found local kube config file, trying to get state from cluster 
INFO[0000] [state] Fetching cluster state from Kubernetes 
INFO[0030] Timed out waiting for kubernetes cluster to get state 
INFO[0030] [certificates] Generating kubernetes certificates 
INFO[0030] [certificates] Generating CA kubernetes certificates 
INFO[0031] [certificates] Generating Kubernetes API server certificates 
FATA[0033] Failed to generate Kubernetes certificates: Failed to generate kube-apiserver certificate: x509: certificate contained IP address of length 0 
```

sangeethah:
rke version v0.0.6-dev

When host ip is provided as fqdn name that is resolvable ,certificate generation has problems.
Able to create K8s cluster successfully when specifying "address" parameter for the node to as fqdn.
```
nodes:
  - address: ec2-<ip>.us-east-2.compute.amazonaws.com
    user: ubuntu
    role: [controlplane]
```

Issue #41, opened 2017-11-23
Rename "ip" to "host" and make "advertised_hostname" optional.


sangeethah:
host config parameter requirements:

advertised_hostname: Optional - If provided ,  name should be resolvable
host:  Required - can be ip of fqdn that is fully resolvable externally
advertise_address: Optional - Internally resolvable ips

Default resolution order ( in kube api):
advertised_hostname
advertise_address
host

This can be altered by configuring the following in kube api:
```
  extra_args:
      kubelet-preferred-address-types: "InternalIP,ExternalIP,Hostname"
```

This will allow for hostname resolution to not take effect and hence allow for providing "advertised_hostname" to be non resolvable names.

sangeethah:
rke version v0.0.6-dev

Following configurations  works fine:
Able to create K8s cluster successfully.
Able to deploy pods.
Able to launch dashboard addon.
Able to exec/log to containers.


1. Specify only address for host:
```
nodes:
  - address: ec2-<ip>.us-east-2.compute.amazonaws.com
    user: ubuntu
    role: [controlplane]
```

In this case , nodes get registered with the "address".

2. Specify host , internal_address and hostname_override. 
In my case hostname_override is not resolvable:

```
 nodes:
  - address: ec2-<ip>.us-east-2.compute.amazonaws.com
    internal_address: <ip>
    hostname_override: mynode1
    user: ubuntu
    role: [controlplane]
```

In this case , nodes get registered with the "hostname_override" value.

PR #56, opened 2017-11-28
RKE host structure and config changes

galal-hussein:
https://github.com/rancher/rke/issues/53
https://github.com/rancher/rke/issues/52
https://github.com/rancher/rke/issues/50
https://github.com/rancher/rke/issues/41
https://github.com/rancher/rke/issues/36

- `IP` to `Address`
- `AdvertiseAddress` to `InternalAddress`
- `AdvertisedHostname` to `HostnameOverride`

Hostname will be optional, and if not set it will be equal to Address.


galal-hussein:
@alena1108 @moelsayed Can you please review?

hosts/hosts.go
```diff
@@ -90,7 +90,7 @@ func IsHostListChanged(currentHosts, configHosts []Host) bool {
 	for _, host := range currentHosts {
 		found := false
 		for _, configHost := range configHosts {
-			if host.AdvertisedHostname == configHost.AdvertisedHostname {
+			if host.Address == configHost.Address {
 				found = true
```
alena1108: we should break when found=true

cluster/validation.go
```diff
@@ -13,16 +15,79 @@ func (c *Cluster) ValidateCluster() error {
...
+			if role != services.ETCDRole && role != services.ControlRole && role != services.WorkerRole {
+				return fmt.Errorf("Role [%s] for host (%d) is not recognized", role, i+1)
+			}
+		}
+		bothRoles := make(map[string]bool)
+		for _, role := range host.Role {
+			if role == services.ControlRole {
+				bothRoles[services.ControlRole] = true
+			}
+			if role == services.WorkerRole {
```
alena1108: should be elseif

cluster/validation.go
```diff
@@ -13,16 +15,79 @@ func (c *Cluster) ValidateCluster() error {
...
+		}
+		if len(host.Role) == 0 {
+			return fmt.Errorf("Role for host (%d) is not provided", i+1)
+		}
+		for _, role := range host.Role {
+			if role != services.ETCDRole && role != services.ControlRole && role != services.WorkerRole {
+				return fmt.Errorf("Role [%s] for host (%d) is not recognized", role, i+1)
+			}
+		}
+		bothRoles := make(map[string]bool)
```
alena1108: any reason why we use a map here? there are only 2 values, workerRole and controlRole, may be can use bool instead to simplify the outlook?

cluster/validation.go
```diff
@@ -13,16 +15,79 @@ func (c *Cluster) ValidateCluster() error {
 	if len(c.EtcdHosts) == 0 {
```
alena1108: Should we validate number of etcd hosts (it should be an odd number)?

cluster/validation.go
```diff
@@ -13,16 +15,79 @@ func (c *Cluster) ValidateCluster() error {
...
-func validateServicesOption(c *Cluster) error {
+func validateAuthOptions(c *Cluster) error {
+	if c.Authentication.Strategy != DefaultAuthStrategy {
+		return fmt.Errorf("Authentication strategy [%s] is not supported", c.Authentication.Strategy)
+	}
+	return nil
+}
+
+func validateNetworkOptions(c *Cluster) error {
+	if c.Network.Plugin != "flannel" && c.Network.Plugin != "calico" && c.Network.Plugin != "canal" {
```
alena1108: I suggest you define the constants for the network plugins.

hosts/hosts.go
galal-hussein: fixed it

cluster/validation.go
```diff
@@ -3,26 +3,88 @@ package cluster
...
 )
 
 func (c *Cluster) ValidateCluster() error {
 	// make sure cluster has at least one controlplane/etcd host
 	if len(c.ControlPlaneHosts) == 0 {
 		return fmt.Errorf("Cluster must have at least one control plane host")
 	}
-	if len(c.EtcdHosts) == 0 {
-		return fmt.Errorf("Cluster must have at least one etcd host")
+	if len(c.EtcdHosts)%2 == 0 {
```
alena1108: what if len is 0?

Issue #54, opened 2017-11-28
Remove cluster from the cli command


galal-hussein:
Remove the `cluster` from cli command, and just use:
```
rke up/down/upgrade
```

sangeethah:
rke version v0.0.6-dev
command to do up and remove does not include cluster anymore.

```
./rke_darwin-amd64 up
./rke_darwin-amd64 remove
```

alarmdev:
if you are running multiple clusters, please note and don't forget to specify --config and cluster config yaml

PR #58, opened 2017-11-28
Command overhaul

moelsayed:
remove cluster command
merge upgrade and up commands
rename down command

https://github.com/rancher/rke/issues/54


cmd/config.go
```diff
@@ -179,7 +179,7 @@ func getHostConfig(reader *bufio.Reader, index int) (*v1.RKEConfigHost, error) {
 	}
 	host.User = sshUser
 
-	dockerSocketPath, err := getConfig(reader, fmt.Sprintf("Docker socker path on host (%s)", advertisedHostname), "/var/run/docker.sock")
+	dockerSocketPath, err := getConfig(reader, fmt.Sprintf("Docker docker path on host (%s)", advertisedHostname), "/var/run/docker.sock")
```
galal-hussein: socket not docker

docker/docker.go
```diff
@@ -50,6 +59,12 @@ func DoRollingUpdateContainer(dClient *client.Client, imageCfg *container.Config
 		logrus.Infof("[%s] Container %s is not running on host [%s]", plane, containerName, hostname)
 		return nil
 	}
+	logrus.Debugf("[%s] Pulling Image on host [%s]", plane, hostname)
```
galal-hussein: can this be info not debug

cmd/remove.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"github.com/rancher/rke/cluster"
+	"github.com/sirupsen/logrus"
+	"github.com/urfave/cli"
+)
+
+func RemoveCommand() cli.Command {
+	removeFlags := []cli.Flag{
+		cli.StringFlag{
+			Name:   "cluster-file",
```
alena1108: as we imply `cluster` in `rke up/remove`, should we rename this parameter to `config`?

alena1108: It would also align with `generate config` command better

cmd/up.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/rancher/rke/pki"
+	"github.com/sirupsen/logrus"
+	"github.com/urfave/cli"
+	"k8s.io/client-go/util/cert"
+)
+
+func UpCommand() cli.Command {
+	upFlags := []cli.Flag{
+		cli.StringFlag{
+			Name:   "cluster-file",
```
alena1108: Same here, we should consider renaming it.

cmd/version.go
```diff
@@ -0,0 +1,35 @@
...
+		cli.StringFlag{
+			Name:   "cluster-file",
+			Usage:  "Specify an alternate cluster YAML file",
+			Value:  cluster.DefaultClusterConfig,
+			EnvVar: "CLUSTER_FILE",
+		},
+	}
+	return cli.Command{
+		Name:   "version",
+		Usage:  "Show Cluster Kubernetes version",
```
alena1108: In other commands descriptions, we have `cluster` lowercased, should probably do the same here

cmd/version.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"fmt"
+
+	"github.com/rancher/rke/cluster"
+	"github.com/urfave/cli"
+)
+
+func VersionCommand() cli.Command {
+	versionFlags := []cli.Flag{
+		cli.StringFlag{
```
alena1108: Rename the param

PR #75, opened 2017-11-29
allow single node deployments

moelsayed:


services/workerplane.go
```diff
@@ -21,12 +21,20 @@ func RunWorkerPlane(controlHosts []hosts.Host, workerHosts []hosts.Host, workerS
 	}
 	for _, host := range workerHosts {
 		// run nginx proxy
-		err := runNginxProxy(host, controlHosts)
-		if err != nil {
-			return err
+		isControlPlaneHost := false
+		for _, role := range host.Role {
+			if role == ControlRole {
+				isControlPlaneHost = true
```
galal-hussein: can you add break here, to break after it finds that its a controlplane host

services/workerplane.go
```diff
@@ -21,12 +21,21 @@ func RunWorkerPlane(controlHosts []hosts.Host, workerHosts []hosts.Host, workerS
...
+				break
+			}
+		}
+		if !isControlPlaneHost {
+			err := runNginxProxy(host, controlHosts)
```
alena1108: Nit: use `if err := runNginxProxy(); err != nil {
return err
}`

services/workerplane.go
```diff
@@ -21,12 +21,21 @@ func RunWorkerPlane(controlHosts []hosts.Host, workerHosts []hosts.Host, workerS
...
+			if err != nil {
+				return err
+			}
 		}
 		// run kubelet
-		err = runKubelet(host, workerServices.Kubelet, false)
+		err := runKubelet(host, workerServices.Kubelet, false)
```
alena1108: same here

Issue #70, opened 2017-11-29
Attempting to delete a node that is not reported as k8S node should be ignored


sangeethah:
rke version v0.0.7-dev

In my case , because of this -  https://github.com/rancher/rke/issues/66 , it left my cluster.yml inconsistent with the actual K8s cluster. 
cluster.yml had an additional control plane entry that is not present in the k8s cluster.
So to make the  cluster.yml consistent with K8s cluster , I removed the control plane entry from cluster.yml. 

Attempted to do "up" .
This results in FATAL error since the attempt made to delete the control plane entry (that was removed from the cluster.yml) fails.

```
INFO[0099] [reconcile] Check Control plane hosts to be deleted 
INFO[0099] [hosts] Cordoning host [<ip>]          
DEBU[0099] Error getting node <ip>: nodes "<ip>" not found 
DEBU[0104] Error getting node <ip>: nodes "<ip>" not found 
DEBU[0109] Error getting node <ip>: nodes "<ip>" not found 
DEBU[0114] Error getting node <ip>: nodes "<ip>" not found 
DEBU[0119] Error getting node <ip>: nodes "<ip>" not found 
DEBU[0124] Error getting node <ip>: nodes "<ip>" not found 
FATA[0129] Failed to delete controlplane node <ip> from cluster 
```

Why would there be an attempt to even delete this host since cluster.yml is now in sync with K8s cluster?

sangeethah:
Tested with rke version v0.0.7-dev

Created a cluster with following configuration using rke up command:
Node1 - etcd
Node2, Node3 - master
Node4, Node5 - worker

Using kubectl command , removed Node3 and Node5
Removed Node3 and Node5 entries from cluster.yml.

Attempted rke up command . 
This succeeded and rolling update of nginx hosts with new list of control plane host also succeeded.

Attempting to delete a node that is not reported as k8S node is ignored.
```
INFO[0005] [reconcile] Reconciling cluster state        
INFO[0005] [hosts] Cordoning host [<ip>] 
WARN[0005] [hosts] Can't find node by name [<ip>] 
```

PR #78, opened 2017-11-30
Warn instead of fatal if deleted host is not found in k8s.

moelsayed:
https://github.com/rancher/rke/issues/70

### DIFF
--- a/hosts/hosts.go
+++ b/hosts/hosts.go
@@ -9,6 +9,7 @@ import (
 	"github.com/rancher/rke/k8s"
 	"github.com/rancher/types/apis/cluster.cattle.io/v1"
 	"github.com/sirupsen/logrus"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/client-go/kubernetes"
 )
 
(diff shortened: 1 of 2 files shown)


hosts/hosts.go
```diff
@@ -56,13 +57,19 @@ func (h *Host) CleanUp() error {
 
 func DeleteNode(toDeleteHost *Host, kubeClient *kubernetes.Clientset) error {
 	logrus.Infof("[hosts] Cordoning host [%s]", toDeleteHost.Address)
-	err := k8s.CordonUncordon(kubeClient, toDeleteHost.HostnameOverride, true)
-	if err != nil {
+	if _, err := k8s.GetNode(kubeClient, toDeleteHost.HostnameOverride); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return err
+		} else {
+			logrus.Warnf("[hosts] Host [%s] is not a Kubernetes node")
```
galal-hussein: can you give the warning some context, like `can't cordon the host because ...`

hosts/hosts.go
```diff
@@ -56,13 +57,19 @@ func (h *Host) CleanUp() error {
 
 func DeleteNode(toDeleteHost *Host, kubeClient *kubernetes.Clientset) error {
 	logrus.Infof("[hosts] Cordoning host [%s]", toDeleteHost.Address)
-	err := k8s.CordonUncordon(kubeClient, toDeleteHost.HostnameOverride, true)
-	if err != nil {
+	if _, err := k8s.GetNode(kubeClient, toDeleteHost.HostnameOverride); err != nil {
+		if !apierrors.IsNotFound(err) {
```
alena1108: Start with `if apierrors.isNotFound`, and if met, log a messgae `Can't find k8s node by name []`

hosts/hosts.go
```diff
@@ -56,13 +57,19 @@ func (h *Host) CleanUp() error {
...
 func DeleteNode(toDeleteHost *Host, kubeClient *kubernetes.Clientset) error {
 	logrus.Infof("[hosts] Cordoning host [%s]", toDeleteHost.Address)
-	err := k8s.CordonUncordon(kubeClient, toDeleteHost.HostnameOverride, true)
-	if err != nil {
+	if _, err := k8s.GetNode(kubeClient, toDeleteHost.HostnameOverride); err != nil {
+		if apierrors.IsNotFound(err) {
+			logrus.Warnf("[hosts] Can't find node by name [%s]")
+			return nil
+		} else {
+			return err
```
alena1108: I don't think there is a need for `else`. Just return err

hosts/hosts.go
galal-hussein: specify the host name for %s

PR #81, opened 2017-11-30
support weave-net network plugin

kyamazawa:
We wanted to integrate weave so added some code. Tested that it works on hosts of same subnets. This could possibly be some start point? Thanks in advance!


galal-hussein:
@kyamazawa Thanks a lot of your contribution, i will review and test the new network plugin and let you know

galal-hussein:
@kyamazawa I think you are missing configuring the cluster cidr, by default weave uses `IPALLOC_RANGE` variable to configure the cluster cidr which defaults to `10.32.0.0/12`, you should pass this value to the yaml file like we do in Flannel and Calico

kyamazawa:
@galal-hussein Thank you for your feedback! Now, weave network plugin supports these options.
- `cluster_cidr`
- `weave_image`
- `weave_cni_image`

Issue #72, opened 2017-11-29
Attempting to remove controlplane role from a node that had etcd and controlplane node results in broken cluster.


sangeethah:
rke version v0.0.7-dev
Steps to reproduce the problem:
Create a kubetnetes setup with following configuration:
node1 - etcd , controlplane
node 2 - worker
node 3 - controlplane.

Modify node1's role to be only "etcd"  (by removing controlplane node).
This results in k8s cluster broken.

Controlplane tear down results in  etcd data dir being removed as well resulting in broken cluster.

sangeethah:
rke version v0.0.7-dev
Steps to reproduce the problem:

Create a kubernetes setup with following configuration:
node1 - etcd , controlplane
node2 ,node3 - worker
node4, node5 - controlplane.

Update node1's role to be only "etcd".

node1 gets removed from the k8s cluster as master node.
etcd container on this node continues to be present.
K8s cluster is healthy . Existing pods continue to work and we are able to create new pods.

Issue #66, opened 2017-11-29
When existing worker node is made a control plane node , the node gets removed from the cluster.


sangeethah:
rke version v0.0.7-dev

Steps to reproduce the problem:
Created a k8s with 1 controlplane node , 1 etcd  and 3 worker nodes.
Updated one of the node in worker plane to become a controlplane node.

When controlplane nodes are getting evaluated:
The updated host, becomes a controlplane node (I do see the system containers relating to control plane being deployed in this host).
Also rolling updates for the nginx-proxy container for all existing worker nodes gets updated with this new host in controlplane node as expetced.

When worker nodes are getting evaluated:
This host gets removed from the k8s cluster.

Expected behavior:
The host should continue to be part of K8s cluster but should be a cotrolplane node.

sangeethah:
Tested with rke version v0.0.8-dev

Created a k8s with 2 controlplane node , 1 etcd and 3 worker nodes.
Updated one of the nodes in worker plane to become a control plane node.

Node that was updated to become a control plane , continues to be in part of K8s cluster and becomes a master node as expected. It has all the control plane system containers running it and no nginx-proxy container running in it.

```
kube-proxy
kubelet
scheduler
kube-controller
kube-api
```

Also nginx-proxy containers in all existing worker nodes now have this new node programmed.

PR #85, opened 2017-12-01
Refactor reconciliation

galal-hussein:
Regactoring Reconciliation
Add role based add/delete for services

#72 #66

### DIFF
--- a/cluster/reconcile.go
+++ b/cluster/reconcile.go
@@ -1,0 +1,129 @@
+package cluster
+
+import (
+	"fmt"
+
+	"github.com/rancher/rke/hosts"
+	"github.com/rancher/rke/k8s"
+	"github.com/rancher/rke/services"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/crypto/ssh"
+	"k8s.io/client-go/kubernetes"
+)
+
+func ReconcileCluster(kubeCluster, currentCluster *Cluster) error {
+	logrus.Infof("[reconcile] Reconciling cluster state")
+	if currentCluster == nil {
+		logrus.Infof("[reconcile] This is newly generated cluster")
+
+		return nil
+	}
+	// to handle if current local admin is down and we need to use new cp from the list
+	if !isLocalConfigWorking(kubeCluster.LocalKubeConfigPath) {
+		if err := rebuildLocalAdminConfig(kubeCluster); err != nil {
+			return err
+		}
+	}
+
+	kubeClient, err := k8s.NewClient(kubeCluster.LocalKubeConfigPath)
+	if err != nil {
+		return fmt.Errorf("Failed to initialize new kubernetes client: %v", err)
+	}
+
+	key, _ := checkEncryptedKey(kubeCluster.SSHKeyPath)
+
+	if err := reconcileWorker(currentCluster, kubeCluster, key, kubeClient); err != nil {
+		return err
+	}
+
+	if err := reconcileControl(currentCluster, kubeCluster, key, kubeClient); err != nil {
+		return err
+	}
+	logrus.Infof("[reconcile] Reconciled cluster state successfully")
+	return nil
+}
+
+func reconcileWorker(currentCluster, kubeCluster *Cluster, key ssh.Signer, kubeClient *kubernetes.Clientset) error {
+	// worker deleted first to avoid issues when worker+controller on same host
+	logrus.Debugf("[reconcile] Check worker hosts to be deleted")
+	wpToDelete := hosts.GetToDeleteHosts(currentCluster.WorkerHosts, kubeCluster.WorkerHosts)
+	for _, toDeleteHost := range wpToDelete {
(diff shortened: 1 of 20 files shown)


cluster/reconcile.go
```diff
@@ -0,0 +1,134 @@
...
+	if err := reconcileControl(currentCluster, kubeCluster, key, kubeClient); err != nil {
+		return err
+	}
+	logrus.Infof("[reconcile] Reconciled cluster state successfully")
+	return nil
+}
+
+func reconcileWorker(currentCluster, kubeCluster *Cluster, key ssh.Signer, kubeClient *kubernetes.Clientset) error {
+	// worker deleted first to avoid issues when worker+controller on same host
+	logrus.Infof("[reconcile] Check worker hosts to be deleted")
```
alena1108: I don't think we should log the `check` statements in INFO, may be the debug would be more appropriate

cluster/reconcile.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	kubeClient, err := k8s.NewClient(kubeCluster.LocalKubeConfigPath)
+	if err != nil {
+		return fmt.Errorf("Failed to initialize new kubernetes client: %v", err)
+	}
+
+	key, _ := checkEncryptedKey(kubeCluster.SSHKeyPath)
+
+	if err := reconcileWorker(currentCluster, kubeCluster, key, kubeClient); err != nil {
```
alena1108: Any particular reason workers are being reconciled first? The official upgrade doc advises to upgrade etcd/master first: https://kubernetes.io/docs/getting-started-guides/ubuntu/upgrades/

cluster/reconcile.go
```diff
@@ -0,0 +1,134 @@
...
+		if err := hosts.DeleteNode(&toDeleteHost, kubeClient, services.ControlRole); err != nil {
+			return fmt.Errorf("Failed to delete worker node %s from cluster", toDeleteHost.Address)
+		}
+		// attempting to clean services/files on the host
+		if err := reconcileHost(toDeleteHost, key, true); err != nil {
+			logrus.Warnf("[reconcile] Couldn't clean up worker node [%s]: %v", toDeleteHost.Address, err)
+			continue
+		}
+	}
+	if err := currentCluster.InvertIndexHosts(); err != nil {
```
alena1108: I feel like this check belongs to the caller method. Also it is a bit concerning we are calling this check method after delete is already performed. I'd do it as a very first step.

docker/docker.go
```diff
@@ -217,10 +218,10 @@ func IsContainerUpgradable(dClient *client.Client, imageCfg *container.Config, c
 	if err != nil {
 		return false, err
 	}
-	if containerInspect.Config.Image == imageCfg.Image {
-		logrus.Debugf("[%s] Container %s is not eligible for updgrade on host [%s]", plane, containerName, hostname)
-		return false, nil
+	if containerInspect.Config.Image != imageCfg.Image || !reflect.DeepEqual(containerInspect.Config.Cmd, imageCfg.Cmd) {
+		logrus.Debugf("[%s] Container %s is eligible for updgrade on host [%s]", plane, containerName, hostname)
```
alena1108: Nit: should %s be surrounded with [] for container? to kinda follow up the format you are using

hosts/hosts.go
```diff
@@ -28,17 +28,54 @@ const (
...
 		ToCleanCNIBin,
 		ToCleanCalicoRun,
 	}
+	return h.CleanUp(toCleanPaths)
+}
+
+func (h *Host) CleanUpWorkerHost(controlRole string) error {
+	for _, role := range h.Role {
+		if role == controlRole {
+			logrus.Infof("[hosts] Host [%s] is already a controlplane host, skipping cleanup.", h.Address)
```
alena1108: I'd rephrase this message to something like: `Worker host is a control place host..`

services/controlplane.go
```diff
@@ -9,6 +9,15 @@ import (
 func RunControlPlane(controlHosts []hosts.Host, etcdHosts []hosts.Host, controlServices v1.RKEConfigServices) error {
 	logrus.Infof("[%s] Building up Controller Plane..", ControlRole)
 	for _, host := range controlHosts {
+
+		for _, role := range host.Role {
```
alena1108: Looks like we iterate over host roles a lot in the code to figure out if the host is worker vs control. Would it be helpful to have 2 booleans in the host struct - isControl,isWorker - and refer to them instead?

cluster/reconcile.go
galal-hussein: This is for the cases where a host has both controller and worker role, and this same host is the only controller host, when the user run ./rke up it should first remove worker components including kubelet and kubeproxy, if we remove the controller components first it won't be able to remove the worker

galal-hussein: Also upgrade is not a part of reconcile, its a part of adding components

alena1108:
LGTM

Issue #84, opened 2017-11-30
Allow user to provide internal-use images for air-gapped setups 


moelsayed:
We need to allow users to provide their own cert downloader and nginx proxy images to allow running using private registries and air-gapped deployments.

galal-hussein:
all images used and statically hardcoded:

- alpine:latest
- rancher/rke-nginx-proxy:0.1.0
- rancher/rke-cert-deployer:0.1.0
- quay.io/calico/kube-controllers:v1.0.0
- quay.io/calico/node:v2.6.2
- quay.io/calico/cni:v1.11.0
- quay.io/coreos/flannel:v0.9.1
- quay.io/coreos/flannel-cni:v0.2.0

soumyalj:
Tested with rke version v0.0.9-dev. Pushed the images above to a private repository and used them in the cluster.yml file. 
When the images are private in the repository, the pull is not successful. 
```
INFO[0002] [reconcile] Reconciling cluster state        
INFO[0002] [reconcile] This is newly generated cluster  
INFO[0002] [certificates] Deploying kubernetes certificates to Cluster nodes 
INFO[0002] [certificates] Checking image [quay.io/soumyalj/rke-cert-deployer-private:latest] on host [159.89.176.30] 
DEBU[0002] Checking if image [quay.io/soumyalj/rke-cert-deployer-private:latest] exists on host [159.89.176.30] 
DEBU[0002] Image [quay.io/soumyalj/rke-cert-deployer-private:latest] does not exist on host [159.89.176.30]: Error: No such image: quay.io/soumyalj/rke-cert-deployer-private:latest 
INFO[0002] [certificates] Pulling image [quay.io/soumyalj/rke-cert-deployer-private:latest] on host [159.89.176.30] 
{"status":"Pulling repository quay.io/soumyalj/rke-cert-deployer-private"}
{"errorDetail":{"code":403,"message":"Error: Status 403 trying to pull repository soumyalj/rke-cert-deployer-private: \"{\\\"error\\\": \\\"Permission Denied\\\"}\""},"error":"Error: Status 403 trying to pull repository soumyalj/rke-cert-deployer-private: \"{\\\"error\\\": \\\"Permission Denied\\\"}\""}
INFO[0003] [certificates] Successfully pulled image [quay.io/soumyalj/rke-cert-deployer-private:latest] on host [159.89.176.30] 
[...]

PR #91, opened 2017-12-02
Configurable images

moelsayed:
https://github.com/rancher/rke/issues/84

### DIFF
--- a/addons/kubedns.go
+++ b/addons/kubedns.go
@@ -1,6 +1,15 @@
 package addons
 
-func GetKubeDNSManifest(clusterDNSServer, clusterDomain string) string {
+const (
+	KubeDNSImage           = "kubeDNSImage"
+	DNSMasqImage           = "DNSMasqImage"
+	KubeDNSSidecarImage    = "kubednsSidecarImage"
+	KubeDNSAutoScalerImage = "kubeDNSAutoScalerImage"
+	KubeDNSServer          = "clusterDNSServer"
+	KubeDNSClusterDomain   = "clusterDomain"
+)
+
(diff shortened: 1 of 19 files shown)


cluster.yml
```diff
@@ -49,6 +51,16 @@ services:
   kubeproxy:
     image: rancher/k8s:v1.8.3-rancher2
 
+
+rke_images:
```
alena1108: lets call it `system_images`, rke is a bit redundant

addons/kubedns.go
```diff
@@ -17,7 +17,7 @@ spec:
     spec:
       containers:
       - name: autoscaler
-        image: gcr.io/google_containers/cluster-proportional-autoscaler-amd64:1.0.0
+        image: ` + kubeDNSConfig["autoscaler_image"] + `
```
alena1108: We should define images names as constants. To avoid repeating them in the code

cluster/cluster.go
```diff
@@ -119,30 +111,30 @@ func (c *Cluster) setClusterDefaults() {
...
+	c.setClusterNetworkDefaults()
+	c.setClusterImageDefaults()
+}
+
+func (c *Cluster) setClusterImageDefaults() {
+	if c.RKEImages == nil {
+		// don't break if the user didn't define rke_images
+		c.RKEImages = make(map[string]string)
+	}
+	setDefaultIfEmptyMapValue(c.RKEImages, "alpine", DefaultAplineImage)
```
alena1108: Consider doing it in some kind of a loop, there is a lot of repeatable stuff. When you read the config, you can put all images in a map, then iterate over this map

PR #92, opened 2017-12-02
SSH key and path per host

galal-hussein:
https://github.com/rancher/rke/issues/97


cluster/cluster.go
```diff
@@ -118,6 +122,9 @@ func (c *Cluster) setClusterDefaults() {
 			// This is a temporary modification
 			c.Nodes[i].HostnameOverride = c.Nodes[i].Address
 		}
+		if len(host.SSHKeyPath) == 0 {
```
alena1108: should we default to cluster only when both sshkeypath and sshkey are empty? As user can pass either of those

galal-hussein: ssh key will have the priority over either of options, so i doesn't matter if the user only passed ssh key, because it will be used instead of any path

cluster/hosts.go
```diff
@@ -2,42 +2,28 @@ package cluster
...
 )
 
 func (c *Cluster) TunnelHosts() error {
-	key, err := checkEncryptedKey(c.SSHKeyPath)
-	if err != nil {
-		return fmt.Errorf("Failed to parse the private key: %v", err)
-	}
 	for i := range c.EtcdHosts {
-		err := c.EtcdHosts[i].TunnelUp(key)
+		err := c.EtcdHosts[i].TunnelUp()
```
alena1108: Nit: lets use if err := <>; err != nil { return fmt.errorf...}

cluster/hosts.go
```diff
@@ -2,42 +2,28 @@ package cluster
...
 	for i := range c.ControlPlaneHosts {
-		err := c.ControlPlaneHosts[i].TunnelUp(key)
+		err := c.ControlPlaneHosts[i].TunnelUp()
 		if err != nil {
 			return fmt.Errorf("Failed to set up SSH tunneling for Control hosts: %v", err)
 		}
 	}
 	for i := range c.WorkerHosts {
-		err := c.WorkerHosts[i].TunnelUp(key)
+		err := c.WorkerHosts[i].TunnelUp()
```
alena1108: same here, lets use one line check for err

hosts/dialer.go
```diff
@@ -86,3 +91,43 @@ func makeSSHConfig(user string, signer ssh.Signer) (*ssh.ClientConfig, error) {
...
+func checkEncryptedKey(sshKey, sshKeyPath string) (ssh.Signer, error) {
+	logrus.Debugf("[ssh] Checking private key")
+	var err error
+	var key ssh.Signer
+	if len(sshKey) > 0 {
+		key, err = parsePrivateKey(sshKey)
+	} else {
+		key, err = parsePrivateKey(privateKeyPath(sshKeyPath))
+	}
+	if err != nil {
```
alena1108: what are we trying to do here? Why do we do all this stuff when there is an error?

galal-hussein: this check is for encrypted keys, parsing the key with the normal function will generate an error, if this error is encrypted key error we prompt to the user a password to enter, if this generated an error to we return error to the user

alena1108: Then can we make it a bit pretty :) if err == nil - return immediately. Then parse the error

Issue #113, opened 2017-12-06
K8s cluster broken when trying to remove control plane node after it in not reachable.


sangeethah:
RKE version: v0.0.8-dev,

Create a cluster with following configuration:
Node1 -> control , worker , Node2-> control and Node3-> etcd
Power down Node1.
changed cluster.yml  with following configuration to remove Node1 from cluster:
Node2-> control , worker and Node3-> etcd

Up command does not succeed with following error:

```
FATA[0088] [certificates] Failed to Save Kubernetes certificates: Failed to save certificate [kube-admin] to kubernetes: [certificates] Timeout waiting for kubernetes to be ready 
```

Not able to connect with updated config using kubectl:

```
$ ./kubectl get nodes
Unable to connect to the server: x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "kube-ca")
```

soumyalj:
Tested with rke version v0.0.9-dev
Create a cluster with following configuration:
Node1 -> control , worker , Node2-> control and Node3-> etcd
Power down Node1.
Changed cluster.yml with following configuration to remove Node1 from cluster:
Node2-> control , worker and Node3-> etcd
./rke up command succeeded and the cluster was created successfully. Creation of pods was successful in the cluster

PR #114, opened 2017-12-06
Rebuild local admin if the current master host is down

galal-hussein:
#113 

If the current cp host is unreachable, rke will try each cp host in the list until it finds one that is working and use it to get cluster state.

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -171,26 +171,27 @@ func GetLocalKubeConfig(configPath string) string {
 }
 
 func rebuildLocalAdminConfig(kubeCluster *Cluster) error {
-	logrus.Infof("[reconcile] Rebuilding and update local kube config")
-	var workingConfig string
+	logrus.Infof("[reconcile] Rebuilding and updating local kube config")
+	var workingConfig, newConfig string
 	currentKubeConfig := kubeCluster.Certificates[pki.KubeAdminCommonName]
 	caCrt := kubeCluster.Certificates[pki.CACertName].Certificate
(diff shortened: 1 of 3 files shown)


cluster/cluster.go
```diff
@@ -171,26 +171,27 @@ func GetLocalKubeConfig(configPath string) string {
...
 		if err := pki.DeployAdminConfig(newConfig, kubeCluster.LocalKubeConfigPath); err != nil {
 			return fmt.Errorf("Failed to redeploy local admin config with new host")
 		}
 		workingConfig = newConfig
-		if _, err := GetK8sVersion(kubeCluster.LocalKubeConfigPath); err != nil {
-			logrus.Infof("[reconcile] host [%s] is not active master on the cluster", cpHost.Address)
-			continue
-		} else {
+		if _, err := GetK8sVersion(kubeCluster.LocalKubeConfigPath); err == nil {
+			logrus.Infof("[reconcile] host [%s] is active master on the cluster", cpHost.Address)
```
alena1108: shall we move `workingConfig = newConfig` inside the if condition? otherwise it would be set even when the getk8sversion call fails

galal-hussein: Actually that has been done on purpose, because there are cases where we need a new config with even a not working control plane host, consider this scenario:

- run rke with one control hosts
- remove these old control host and add new one in the cluster.yml
- run rke up again

rke in that case will pull the certificates from the old node and rebuild local admin with the new (and not working) host and use it after it deploys all components to save kubernetes certificates in the end

Issue #115, opened 2017-12-06
Calico inter-host connectivity problem on AWS.


moelsayed:
**RKE version:**
v0.0.8-dev
**Docker version: (`docker version`,`docker info` preferred)**
1.12
**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)**
ubuntu 16.04
**Type/provider of hosts: (VirtualBox/Bare-metal/AWS/GCE/DO)**
aws
**cluster.yml file:**

**Steps to Reproduce:**
Configure a cluster of multiple hosts with calico network plugin.
Pods on different hosts can't reach each other 
**Results:**

moelsayed:
Calico needs specific [configuration](https://docs.projectcalico.org/v2.6/reference/public-cloud/aws) to work on AWS, both on AWS side and on the cluster side. 

The fix implements the needed configuration on the cluster side. On AWS, you need to [disable src/dest checks](http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_NAT_Instance.html#EIP_Disable_SrcDestCheck) on the cluster nodes.

soumyalj:
Tested with rke version v0.0.9-dev. Disabled src/des checks on AWS hosts and used them to create a cluster successfully. The config file had the calico settings as below:
```
auth:
  strategy: x509
network:
  plugin: calico
  options:
    calico_cloudprovider : aws
```
Created user pods and verified inter-host communication by exec'ing to the pods. 
Ping between pods on different hosts was successful
```
root@k8testrc2-9xvt4:/# ping 10.233.127.67
PING 10.233.127.67 (10.233.127.67): 48 data bytes
56 bytes from 10.233.127.67: icmp_seq=0 ttl=62 time=0.535 ms
56 bytes from 10.233.127.67: icmp_seq=1 ttl=62 time=0.587 ms
56 bytes from 10.233.127.67: icmp_seq=2 ttl=62 time=0.524 ms
56 bytes from 10.233.127.67: icmp_seq=3 ttl=62 time=0.541 ms
56 bytes from 10.233.127.67: icmp_seq=4 ttl=62 time=0.475 ms
56 bytes from 10.233.127.67: icmp_seq=5 ttl=62 time=0.531 ms


```

PR #118, opened 2017-12-07
Enable AWS specific configuration for Calico

moelsayed:
https://github.com/rancher/rke/issues/115

### DIFF
--- a/cluster.yml
+++ b/cluster.yml
@@ -10,6 +10,9 @@ auth:
 # calico
 # canal
 # weave
+#
+# If you are using calico on AWS, use the network plugin config option:
+# 'cloud_provider: aws'
 network:
   plugin: flannel
   options:
--- a/network/calico.go
+++ b/network/calico.go
@@ -1,6 +1,54 @@
 package network
 
 func GetCalicoManifest(calicoConfig map[string]string) string {
+	awsIPPool := ""
+	if calicoConfig[CloudProvider] == AWSCloudProvider {
+		awsIPPool = `
+---
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: aws-ippool
+  namespace: kube-system
+data:
+  aws-ippool: |-
+    apiVersion: v1
+    kind: ipPool
+    metadata:
+      cidr: ` + calicoConfig[ClusterCIDR] + `
+    spec:
+      nat-outgoing: true
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: calicoctl
+  namespace: kube-system
+spec:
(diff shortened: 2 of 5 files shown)


cluster.yml
```diff
@@ -9,6 +9,8 @@ auth:
 # flannel
 # calico
 # canal
+#
+# if you are using calico on AWS, use the config option: calico_aws: true
```
alena1108: I feel aws should be a part of plugin options...not a top level. @galal-hussein @moelsayed what are your thoughts on that

alena1108: Oh, looked down the code, seems like it is a part of options. Can we rephrase the help message? Something like "for plugin calico, put aws=true to network/options field", otherwise it not clear what "config options" implies. Lets also rename calico_aws to aws as calico is redundant in this case.

cluster.yml
```diff
@@ -10,6 +10,9 @@ auth:
 # calico
 # canal
 # weave
+#
+# If you are using calico on AWS, use the network plugin config option:
+# 'aws: true'
```
alena1108: @moelsayed as a second thought, got concerned if there may be other cloud providers requiring special handling for calico plugins? Or is it just aws. If there is more than one cloud provider, then may be we should consider changing it to `cloudprovider: aws/gce/etc`. Let me know

network/network.go
```diff
@@ -11,6 +11,9 @@ const (
 	CNIImage         = "cniImage"
 	NodeImage        = "nodeImage"
 	ControllersImage = "controllersImage"
+	CalicoctlImage   = "calicoctlImage"
 	FlannelImage     = "flannelImage"
 	FlannelCNIImage  = "flannelCNIImage"
+	CloudProvider    = "cloudprovider"
```
alena1108: I believe CloudProvider constant is defined in cluster/network.go already? is there a reason why we are redefining it here?

network/calico.go
```diff
@@ -1,6 +1,54 @@
 package network
 
 func GetCalicoManifest(calicoConfig map[string]string) string {
+	awsIPPool := ""
+	if calicoConfig[CloudProvider] == AWSCloudProvider {
+		awsIPPool = `
+---
+kind: ConfigMap
```
alena1108: Is there way to not embed the yaml file contents in the calico.go? But rather store them in the separate yaml file, and just invoke it?

network/network.go
moelsayed: Yes, the same constant name is used, but they don't necessarily have to have the same value, this is simply to avoid sending constants across packages.

Issue #124, opened 2017-12-08
no idempotency for rke remove


galal-hussein:
**RKE version:**
v0.0.8-dev

running `rke remove` twice will result in the following error:
```
INFO[0012] Removing local admin Kubeconfig: ./.kube_config_cluster.yml 
FATA[0012] Failed to remove local admin Kubeconfig file: remove ./.kube_config_cluster.yml: no such file or directory 
```
we should just add warning instead of fatal

soumyalj:
Tested with rke version v0.0.9-dev. Created a cluster and performed a remove twice. 
The following warning message was displayed instead of a FATAL error
```
INFO[0016] Removing local admin Kubeconfig: ./.kube_config_cluster1.yml 
WARN[0016] Failed to remove local admin Kubeconfig file: remove ./.kube_config_cluster1.yml: no such file or directory 
INFO[0016] Cluster removed successfully                 
```

PR #126, opened 2017-12-08
Add replace fatal with warning when can't remove local kubeconfig

galal-hussein:
#124

### DIFF
--- a/pki/deploy.go
+++ b/pki/deploy.go
@@ -111,11 +111,11 @@ func DeployAdminConfig(kubeConfig, localConfigPath string) error {
 	return nil
 }
 
-func RemoveAdminConfig(localConfigPath string) error {
+func RemoveAdminConfig(localConfigPath string) {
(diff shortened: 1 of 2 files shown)


pki/deploy.go
```diff
@@ -111,11 +111,10 @@ func DeployAdminConfig(kubeConfig, localConfigPath string) error {
 	return nil
 }
 
-func RemoveAdminConfig(localConfigPath string) error {
+func RemoveAdminConfig(localConfigPath string) {
 	logrus.Infof("Removing local admin Kubeconfig: %s", localConfigPath)
 	if err := os.Remove(localConfigPath); err != nil {
-		return fmt.Errorf("Failed to remove local admin Kubeconfig file: %v", err)
+		logrus.Warningf("Failed to remove local admin Kubeconfig file: %v", err)
```
alena1108: Should return right after logging. Otherwise info message will be logged as well.

PR #132, opened 2017-12-11
Using a custom dialer for cluster hosts

galal-hussein:


cmd/remove.go
```diff
@@ -72,5 +73,5 @@ func clusterRemoveFromCli(ctx *cli.Context) error {
 	if err != nil {
 		return fmt.Errorf("Failed to resolve cluster file: %v", err)
 	}
-	return ClusterRemove(clusterFile)
+	return ClusterRemove(clusterFile, nil)
```
alena1108: Just to confirm that is when the code is called from cli context, the connection should be done using old way - ssh. And the customDialer will be used only when called from the kontainer-engine

cmd/up.go
```diff
@@ -27,10 +28,10 @@ func UpCommand() cli.Command {
 	}
 }
 
-func ClusterUp(clusterFile string) (string, string, string, string, error) {
+func ClusterUp(clusterFile string, customDialer hosts.CustomDialer) (string, string, string, string, error) {
```
alena1108: doesn't look like all the values this method return, being used? why not just return an error then?

galal-hussein: these are the values that kontainer-engine requested like `APIURL`,`cacrt`,`cakey`, etc

alena1108: Unless they are used when called from kontainer-engine?

Issue #134, opened 2017-12-12
Add support for --iface to be specified for flannel


chrisurwin:
**RKE version:**
0.8-dev

Flannel out of the box supports the specifying of --iface to allow you to specify the interface that flannel is to bind to. By default it uses the default gateway interface, this causes a problem for Virtualbox as this is the eth0 and this always gets 10.0.2.15 on all machines which causes flannel to fail.

https://coreos.com/flannel/docs/latest/running.html under the Running on Vagrant details this

soumyalj:
Tested with RKE version 0.0.9-dev. Created a cluster using AWS hosts by providing the flannel iface option.A new interface "eth1" was added and attached to the hosts. 
```
network:
  plugin: flannel
  options:
    flannel_iface: "eth1"
```
Cluster creation was successful . The flannel container logs were observed to ensure that the specified interface was used:
```
sudo docker logs -f 25c215720b46
I0108 21:54:46.970042       1 main.go:487] Using interface with name eth1 and address 172.31.19.24
I0108 21:54:46.970179       1 main.go:504] Defaulting external address to interface address (172.31.19.24)
I0108 21:54:46.985931       1 kube.go:130] Waiting 10m0s for node controller to sync
```
On a setup with GCE hosts, the interface "ens4" was provided in the cluster config file
The flannel logs showed the interface used correctly
```
sudo docker logs -f 0b6fc5cc8cc4
I0109 05:10:28.263802       1 main.go:487] Using interface with name ens4 and address 10.128.0.13
I0109 05:10:28.263983       1 main.go:504] Defaulting external address to interface address (10.128.0.13)
I0109 05:10:28.299511       1 kube.go:130] Waiting 10m0s for node controller to sync
I0109 05:10:28.299552       1 kube.go:283] Starting kube subnet manager
I0109 05:10:29.299705       1 kube.go:137] Node controller sync successful


```

PR #135, opened 2017-12-12
Add iface option to flannel

galal-hussein:
Added `flannel_iface` option to flannel network deployment:
```
network:
  plugin: flannel
  options:
    flannel_iface: "ens4"
```

#134


galal-hussein:
CC @chrisurwin

Issue #137, opened 2017-12-12
Request: option to not pull images if already present


bgehman:
**RKE version:**
`rke version v0.0.8-dev`

**Docker version: (`docker version`,`docker info` preferred)**
```
$ docker info

Containers: 0
 Running: 0
 Paused: 0
 Stopped: 0
Images: 12
Server Version: 1.12.6
Storage Driver: aufs
 Root Dir: /var/lib/docker/aufs
 Backing Filesystem: extfs
 Dirs: 67
 Dirperm1 Supported: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: host null bridge overlay
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Security Options: apparmor seccomp
Kernel Version: 4.4.0-103-generic
Operating System: Ubuntu 16.04.3 LTS
OSType: linux
Architecture: x86_64
CPUs: 2
Total Memory: 3.855 GiB
Name: ubuntu
ID: W4UI:5PVV:WIJ3:BUKF:SOKU:2F7F:UM7A:PXS6:GUOM:YCJP:2AQ6:L6VR
Docker Root Dir: /var/lib/docker
Debug Mode (client): false
Debug Mode (server): false
Registry: https://index.docker.io/v1/
WARNING: No swap limit support
Insecure Registries:
 127.0.0.0/8
```

**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)**
```
$ cat /etc/os-release

NAME="Ubuntu"
VERSION="16.04.3 LTS (Xenial Xerus)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 16.04.3 LTS"
VERSION_ID="16.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
VERSION_CODENAME=xenial
UBUNTU_CODENAME=xenial
```

```
$ uname -r
4.4.0-103-generic
```

**Type/provider of hosts: (VirtualBox/Bare-metal/AWS/GCE/DO)**
KVM VM on laptop -- simulating a network isolated "on-prem" installation.

**cluster.yml file:**
```
nodes:
- address: 192.168.123.27 
  role:
  - controlplane
  - worker
  - etcd
  user: ubuntu 
  docker_socket: /var/run/docker.sock

services:

[...]

PR #139, opened 2017-12-13
Check image locally before pull

superseb:
https://github.com/rancher/rke/issues/137

### DIFF
--- a/docker/docker.go
+++ b/docker/docker.go
@@ -36,12 +36,10 @@ func DoRunContainer(dClient *client.Client, imageCfg *container.Config, hostCfg 
 		return nil
 	}
 
-	logrus.Infof("[%s] Pulling Image on host [%s]", plane, hostname)
-	err = PullImage(dClient, hostname, imageCfg.Image)
+	err = UseLocalOrPull(dClient, hostname, imageCfg.Image, plane)
 	if err != nil {
 		return err
 	}
-	logrus.Infof("[%s] Successfully pulled [%s] image on host [%s]", plane, containerName, hostname)
 	resp, err := dClient.ContainerCreate(context.Background(), imageCfg, hostCfg, nil, containerName)
 	if err != nil {
 		return fmt.Errorf("Failed to create [%s] container on host [%s]: %v", containerName, hostname, err)
@@ -64,12 +62,10 @@ func DoRollingUpdateContainer(dClient *client.Client, imageCfg *container.Config
 		logrus.Infof("[%s] Container %s is not running on host [%s]", plane, containerName, hostname)
 		return nil
 	}
-	logrus.Infof("[%s] Pulling Image on host [%s]", plane, hostname)
-	err = PullImage(dClient, hostname, imageCfg.Image)
+	err = UseLocalOrPull(dClient, hostname, imageCfg.Image, plane)
 	if err != nil {
 		return err
 	}
-	logrus.Infof("[%s] Successfully pulled [%s] image on host [%s]", plane, containerName, hostname)
 	logrus.Debugf("[%s] Stopping old container", plane)
 	oldContainerName := "old-" + containerName
 	if err := StopRenameContainer(dClient, hostname, containerName, oldContainerName); err != nil {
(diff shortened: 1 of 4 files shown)


docker/docker.go
```diff
@@ -130,6 +130,21 @@ func IsContainerRunning(dClient *client.Client, hostname string, containerName s
 	return false, nil
 }
 
+func LocalImageExists(dClient *client.Client, hostname string, containerImage string) (bool, error) {
```
galal-hussein: i don't think this function should be exported

docker/docker.go
```diff
@@ -64,12 +64,12 @@ func DoRollingUpdateContainer(dClient *client.Client, imageCfg *container.Config
...
 	}
-	logrus.Infof("[%s] Pulling Image on host [%s]", plane, hostname)
-	err = PullImage(dClient, hostname, imageCfg.Image)
+	logrus.Infof("[%s] Checking image on host [%s]", plane, hostname)
+	err = UseLocalOrPull(dClient, hostname, imageCfg.Image)
 	if err != nil {
 		return err
 	}
-	logrus.Infof("[%s] Successfully pulled [%s] image on host [%s]", plane, containerName, hostname)
+	logrus.Infof("[%s] Image [%s] exists on host [%s]", plane, containerName, hostname)
```
galal-hussein: this info message will show that image exist even if it pulled the image, it should reflect what happened exactly (pulled or found)

galal-hussein: maybe add the message in UseLocalOrPull function, instead of using it for each occurrence

docker/docker.go
```diff
@@ -130,6 +130,21 @@ func IsContainerRunning(dClient *client.Client, hostname string, containerName s
...
 
+func LocalImageExists(dClient *client.Client, hostname string, containerImage string) (bool, error) {
+	logrus.Debugf("Checking if image [%s] exists on host [%s]", containerImage, hostname)
+	_, _, err := dClient.ImageInspectWithRaw(context.Background(), containerImage)
+	if err != nil {
+		if client.IsErrNotFound(err) {
+			logrus.Debugf("Image [%s] does not exist on host [%s]: %v", containerImage, hostname, err)
+			return false, nil
+		}
+		logrus.Debugf("Error checking if image [%s] exists on host [%s]: %v", containerImage, hostname, err)
```
galal-hussein: there is no need for extra debug message, its the same as the error returned

pki/deploy.go
```diff
@@ -62,8 +62,8 @@ func DeployCertificatesOnWorkers(workerHosts []*hosts.Host, crtMap map[string]Ce
 }
 
 func doRunDeployer(host *hosts.Host, containerEnv []string, certDownloaderImage string) error {
-	logrus.Debugf("[certificates] Pulling Certificate downloader Image on host [%s]", host.Address)
-	if err := docker.PullImage(host.DClient, host.Address, certDownloaderImage); err != nil {
+	logrus.Debugf("[certificates] Checking image [%s] on host [%s]", certDownloaderImage, host.Address)
```
galal-hussein: same as before, can you remove the `Checking image` logs from all calls and add it instead to the UseLocalOrPull

docker/docker.go
superseb: The reasoning here is that you would lose `plane` in the log message when you move it to `UseLocalOrPull`. Passing `plane` for just the log doesn't seem the right way, even as printing the message as return from the function.

docker/docker.go
```diff
@@ -145,6 +155,24 @@ func PullImage(dClient *client.Client, hostname string, containerImage string) e
 	return nil
 }
 
+func UseLocalOrPull(dClient *client.Client, hostname string, containerImage string, plane string) error {
+	logrus.Infof("[%s] Checking image [%s] on host [%s]", plane, containerImage, hostname)
+	imageExists, err := localImageExists(dClient, hostname, containerImage)
+	if err != nil {
+		return err
+	}
+	if !imageExists {
```
alena1108: Nit: better to use if imageExists { return nil} construct. Always better to check the positive

docker/docker.go
```diff
@@ -145,6 +155,24 @@ func PullImage(dClient *client.Client, hostname string, containerImage string) e
...
 }
 
+func UseLocalOrPull(dClient *client.Client, hostname string, containerImage string, plane string) error {
+	logrus.Infof("[%s] Checking image [%s] on host [%s]", plane, containerImage, hostname)
+	imageExists, err := localImageExists(dClient, hostname, containerImage)
+	if err != nil {
+		return err
+	}
+	if imageExists {
+		logrus.Infof("[%s] No pull necessary, image [%s] exists on host [%s]", plane, containerImage, hostname)
```
alena1108: we should log it in debug

Issue #131, opened 2017-12-11
Add RBAC and PodSecurity Policy Support


galal-hussein:
Add RBAC with option at the cluster level, for example: `rbac: true`

soumyalj:
Tested with rke version v0.0.9-dev. Created an RBAC setup. Created some user pods. Created role for reading pods and rolebinding to user1 for the role created. Ymls used below:
```
kind: Role
apiVersion: rbac.authorization.k8s.io/v1beta1
metadata:
  namespace: default
  name: pod-reader
rules:
- apiGroups: [""] # "" indicates the core API group
  resources: ["pods", "services"]
  verbs: ["get", "watch", "list"]
```
```
kind: RoleBinding
apiVersion: rbac.authorization.k8s.io/v1beta1
metadata:
  name: read-pods
  namespace: default
subjects:
- kind: User
  name: user1
  apiGroup: rbac.authorization.k8s.io
roleRef:
  kind: Role
  name: pod-reader
  apiGroup: rbac.authorization.k8s.io
```
User1 was able to only get the pods and creation of pods was not allowed for user1
```
rke soumya$ kubectl --kubeconfig .kube_config_clusterwithrbacdashboard.yml --as=user1 create -f pod5.yml 
Error from server (Forbidden): error when creating "pod5.yml": services is forbidden: User "user1" cannot create services in the namespace "default"
Error from server (Forbidden): error when creating "pod5.yml": replicationcontrollers is forbidden: User "user1" cannot create replicationcontrollers in the namespace "default"
```
If the role is edited to add create option, user1 can create pods.
Clusterrole and Clusterrolebinding are also functional

PR #147, opened 2017-12-14
Enable RBAC

moelsayed:
https://github.com/rancher/rke/issues/131

### DIFF
--- a/addons/addons.go
+++ b/addons/addons.go
@@ -11,6 +11,7 @@ spec:
       name: pi
     spec:
         hostNetwork: true
+        serviceAccountName: rke-job-deployer
         nodeName: ` + nodeName + `
         containers:
           - name: ` + addonName + `-pod
--- a/authz/authz.go
+++ b/authz/authz.go
@@ -1,0 +1,35 @@
+package authz
+
+import (
+	"github.com/rancher/rke/k8s"
+	"github.com/sirupsen/logrus"
+)
+
+func ApplyJobDeployerServiceAccount(kubeConfigPath string) error {
+	logrus.Infof("[authz] Creating rke-job-deployer ServiceAccount")
+	k8sClient, err := k8s.NewClient(kubeConfigPath)
+	if err != nil {
+		return err
+	}
+	if err := k8s.UpdateClusterRoleBindingFromYaml(k8sClient, jobDeployerClusterRoleBinding); err != nil {
+		return err
+	}
+	if err := k8s.UpdateServiceAccountFromYaml(k8sClient, jobDeployerServiceAccount); err != nil {
+		return err
+	}
+	logrus.Infof("[authz] rke-job-deployer ServiceAccount created successfully")
+	return nil
+}
+
+func ApplySystemNodeClusterRoleBinding(kubeConfigPath string) error {
+	logrus.Infof("[authz] Creating system:node ClusterRoleBinding")
+	k8sClient, err := k8s.NewClient(kubeConfigPath)
+	if err != nil {
+		return err
+	}
+	if err := k8s.UpdateClusterRoleBindingFromYaml(k8sClient, systemNodeClusterRoleBinding); err != nil {
+		return err
+	}
+	logrus.Infof("[authz] system:node ClusterRoleBinding created successfully")
+	return nil
+}
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/rancher/rke/authz"
 	"github.com/rancher/rke/hosts"
 	"github.com/rancher/rke/pki"
 	"github.com/rancher/rke/services"
@@ -58,9 +59,14 @@ func (c *Cluster) DeployClusterPlanes() error {
 	err = services.RunControlPlane(c.ControlPlaneHosts,
 		c.EtcdHosts,
 		c.Services,
-		c.SystemImages[ServiceSidekickImage])
+		c.SystemImages[ServiceSidekickImage],
+		c.Authorization.Mode)
 	if err != nil {
 		return fmt.Errorf("[controlPlane] Failed to bring up Control Plane: %v", err)
+	}
+	err = c.ApplyRBACResources()
+	if err != nil {
+		return fmt.Errorf("[auths] Failed to apply RBAC resources: %v", err)
 	}
 	err = services.RunWorkerPlane(c.ControlPlaneHosts,
 		c.WorkerHosts,
@@ -238,4 +244,16 @@ func getLocalAdminConfigWithNewAddress(localConfigPath, cpAddress string) string
 		string(config.CAData),
 		string(config.CertData),
 		string(config.KeyData))
+}
+
+func (c *Cluster) ApplyRBACResources() error {
+	if err := authz.ApplyJobDeployerServiceAccount(c.LocalKubeConfigPath); err != nil {
+		return fmt.Errorf("Failed to apply the ServiceAccount needed for job execution: %v", err)
+	}
+	if c.Authorization.Mode == services.RBACAuthorizationMode {
+		if err := authz.ApplySystemNodeClusterRoleBinding(c.LocalKubeConfigPath); err != nil {
+			return fmt.Errorf("Failed to apply the ClusterRoleBinding needed for node authorization: %v", err)
+		}
+	}
+	return nil
 }
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -128,11 +128,18 @@ func clusterConfig(ctx *cli.Context) error {
 	cluster.Network = *networkConfig
 
 	// Get Authentication Config
-	authConfig, err := getAuthConfig(reader)
+	authnConfig, err := getAuthnConfig(reader)
 	if err != nil {
 		return err
 	}
-	cluster.Authentication = *authConfig
+	cluster.Authentication = *authnConfig
+
+	// Get Authorization config
+	authzConfig, err := getAuthzConfig(reader)
+	if err != nil {
+		return err
+	}
+	cluster.Authorization = *authzConfig
 
 	// Get Services Config
 	serviceConfig, err := getServiceConfig(reader)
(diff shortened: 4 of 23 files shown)


cmd/config.go
```diff
@@ -273,8 +273,8 @@ func getServiceConfig(reader *bufio.Reader) (*v3.RKEConfigServices, error) {
 	return &servicesConfig, nil
 }
 
-func getAuthConfig(reader *bufio.Reader) (*v3.AuthConfig, error) {
-	authConfig := v3.AuthConfig{}
+func getAuthConfig(reader *bufio.Reader) (*v3.AuthnConfig, error) {
```
galal-hussein: Can this also be called getAuthnConfig also we need to create getAuthzConfig to ask user about authorization mode

k8s/clusterrole.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"bytes"
+
+	rbacv1 "k8s.io/api/rbac/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	yamlutil "k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/client-go/kubernetes"
+)
+
+func ApplySystemNodeClusterRoleBinding(kubeConfigPath string) error {
```
galal-hussein: We shouldn't have this in k8s package, k8s package should only have wrapper functions to create functions, this should be in a separate package like `authorization` to include the apply and static manifests as constants

k8s/job.go
```diff
@@ -14,6 +14,40 @@ import (
 	yamlutil "k8s.io/apimachinery/pkg/util/yaml"
 )
 
+func ApplyJobExecuterServiceAccount(kubeConfigPath string) error {
```
galal-hussein: same as above, we can include this in a separate authorization package, better to have also a separate package for rbac

k8s/job.go
```diff
@@ -14,6 +14,40 @@ import (
...
+	jobExecuterServiceAccount := `
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: rke-job-executer
+  namespace: kube-system
+`
+
+	jobExecuterClusterRoleBinding := `
```
galal-hussein: this should be included as cosntants since they are static and will not be changed

galal-hussein: also i think this is a duplicate

k8s/manifests.go
```diff
@@ -0,0 +1,40 @@
+package k8s
+
+var systemNodeClusterRoleBinding = `
```
galal-hussein: should be constants instead of variables

k8s/serviceaccount.go
```diff
@@ -0,0 +1,31 @@
...
+	"bytes"
+
+	"k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	yamlutil "k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/client-go/kubernetes"
+)
+
+func UpdateServiceAccount(k8sClient *kubernetes.Clientset, serviceAccountYaml string) error {
```
galal-hussein: can we include the fact that we are decoding yaml files in the function name something like `UpdateServiceAccountFromYaml`

cmd/config.go
```diff
@@ -273,15 +280,25 @@ func getServiceConfig(reader *bufio.Reader) (*v3.RKEConfigServices, error) {
...
+		return nil, err
+	}
+	authnConfig.Strategy = authnType
+	return &authnConfig, nil
+}
 
-	authType, err := getConfig(reader, "Authentication Strategy", "x509")
+func getAuthzConfig(reader *bufio.Reader) (*v3.AuthzConfig, error) {
+	authzConfig := v3.AuthzConfig{}
+	authzMode, err := getConfig(reader, "Authorization Mode", "")
```
galal-hussein: can we add `rbac` here

cluster/cluster.go
```diff
@@ -70,6 +72,14 @@ func (c *Cluster) DeployClusterPlanes() error {
 	if err != nil {
 		return fmt.Errorf("[workerPlane] Failed to bring up Worker Plane: %v", err)
 	}
+	if err = authz.ApplyJobExecuterServiceAccount(c.LocalKubeConfigPath); err != nil {
```
galal-hussein: I think this should happen after deploying control plane directly and not in after worker plane to avoid kubelet to generate error logs

k8s/clusterrole.go
```diff
@@ -0,0 +1,28 @@
...
+func UpdateClusterRoleBindingFromYaml(k8sClient *kubernetes.Clientset, clusterRoleBindingYaml string) error {
+	clusterRoleBinding := rbacv1.ClusterRoleBinding{}
+	decoder := yamlutil.NewYAMLToJSONDecoder(bytes.NewReader([]byte(clusterRoleBindingYaml)))
+	if err := decoder.Decode(&clusterRoleBinding); err != nil {
+		return err
+	}
+	if _, err := k8sClient.RbacV1().ClusterRoleBindings().Create(&clusterRoleBinding); err != nil {
+		if apierrors.IsAlreadyExists(err) {
+			if _, err := k8sClient.RbacV1().ClusterRoleBindings().Update(&clusterRoleBinding); err != nil {
+				return err
```
galal-hussein: We should add retry for N seconds to make sure kubernetes is up and running, this causes errors in newly created clusters 
```
FATA[0013] Failed to apply the ServiceAccount needed for job execution: Post https://35.226.36.97:6443/apis/rbac.authorization.k8s.io/v1/clusterrolebindings: dial tcp 35.226.36.97:6443: getsockopt: connection refused  
```

k8s/serviceaccount.go
```diff
@@ -0,0 +1,31 @@
...
+
+	decoder := yamlutil.NewYAMLToJSONDecoder(bytes.NewReader([]byte(serviceAccountYaml)))
+	if err := decoder.Decode(&serviceAccount); err != nil {
+		return err
+	}
+
+	if _, err := k8sClient.CoreV1().ServiceAccounts(metav1.NamespaceSystem).Create(&serviceAccount); err != nil {
+		if apierrors.IsAlreadyExists(err) {
+			if _, err := k8sClient.CoreV1().ServiceAccounts(metav1.NamespaceSystem).Update(&serviceAccount); err != nil {
+				return err
```
galal-hussein: same as above

authz/authz.go
```diff
@@ -0,0 +1,25 @@
+package authz
+
+import "github.com/rancher/rke/k8s"
+
+func ApplyJobExecuterServiceAccount(kubeConfigPath string) error {
+
```
galal-hussein: Nit: can we add some logging information to indicate that we are applying service account

authz/authz.go
```diff
@@ -0,0 +1,25 @@
...
+		return err
+	}
+	if err := k8s.UpdateClusterRoleBindingFromYaml(k8sClient, jobExecuterClusterRoleBinding); err != nil {
+		return err
+	}
+	return k8s.UpdateServiceAccountFromYaml(k8sClient, jobExecuterServiceAccount)
+
+}
+
+func ApplySystemNodeClusterRoleBinding(kubeConfigPath string) error {
```
galal-hussein: Nit: can we add some logging information to indicate that we are applying a role binding

addons/addons.go
```diff
@@ -11,6 +11,7 @@ spec:
       name: pi
     spec:
         hostNetwork: true
+        serviceAccountName: rke-job-executer
```
alena1108: should it be executOr?

cluster/cluster.go
```diff
@@ -58,10 +59,19 @@ func (c *Cluster) DeployClusterPlanes() error {
 	err = services.RunControlPlane(c.ControlPlaneHosts,
 		c.EtcdHosts,
 		c.Services,
-		c.SystemImages[ServiceSidekickImage])
+		c.SystemImages[ServiceSidekickImage],
+		c.Authorization.Mode)
 	if err != nil {
 		return fmt.Errorf("[controlPlane] Failed to bring up Control Plane: %v", err)
 	}
+	if err = authz.ApplyJobExecuterServiceAccount(c.LocalKubeConfigPath); err != nil {
```
alena1108: can we extract RBAC config call into a separate method? Would be easier to read.

k8s/serviceaccount.go
```diff
@@ -0,0 +1,38 @@
...
+	if err != nil {
+		return err
+	}
+	for retries := 0; retries <= 5; retries++ {
+		_, err = k8sClient.CoreV1().ServiceAccounts(metav1.NamespaceSystem).Create(&serviceAccount)
+		if err != nil {
+			if apierrors.IsAlreadyExists(err) {
+				if _, err = k8sClient.CoreV1().ServiceAccounts(metav1.NamespaceSystem).Update(&serviceAccount); err != nil {
+					logrus.Debugf("Failed to update ServiceAccount, retrying..")
+					time.Sleep(time.Second * 5)
```
alena1108: Can we just break from the loop when there is no error? This way you don't have to repeat the sleep in both conditions, and do it only when the execution was not successful

cmd/config.go
```diff
@@ -273,15 +280,25 @@ func getServiceConfig(reader *bufio.Reader) (*v3.RKEConfigServices, error) {
 	return &servicesConfig, nil
 }
 
-func getAuthConfig(reader *bufio.Reader) (*v3.AuthConfig, error) {
-	authConfig := v3.AuthConfig{}
+func getAuthnConfig(reader *bufio.Reader) (*v3.AuthnConfig, error) {
+	authnConfig := v3.AuthnConfig{}
+
+	authnType, err := getConfig(reader, "Authentication Strategy", "x509")
```
alena1108: I believe x509 is defined in defaults. We should avoid hardcoding.

k8s/clusterrole.go
```diff
@@ -0,0 +1,36 @@
...
+func UpdateClusterRoleBindingFromYaml(k8sClient *kubernetes.Clientset, clusterRoleBindingYaml string) error {
+	clusterRoleBinding := rbacv1.ClusterRoleBinding{}
+	decoder := yamlutil.NewYAMLToJSONDecoder(bytes.NewReader([]byte(clusterRoleBindingYaml)))
+	err := decoder.Decode(&clusterRoleBinding)
+	if err != nil {
+		return err
+	}
+	for retries := 0; retries <= 5; retries++ {
+		_, err = k8sClient.RbacV1().ClusterRoleBindings().Create(&clusterRoleBinding)
+		if err != nil {
```
alena1108: I'm sure I've seen this complicated loop in some other place in this PR :) I'd suggest you simplify it

PR #154, opened 2017-12-16
Add Dialer Factory and receive rkeConfig instead of cluster yaml

galal-hussein:


hosts/dialer.go
```diff
@@ -51,3 +48,23 @@ func (d *sshDialer) Dial(network, addr string) (net.Conn, error) {
 	}
 	return remote, err
 }
+
+func (h *Host) newHTTPClient(dialerFactory DialerFactory) (*http.Client, error) {
+	var defaultFactory DialerFactory
```
alena1108: Nit: should we drop `default` from the name? As it is not really the default, you are assigning the value to it

alena1108:
LGTM

PR #155, opened 2017-12-16
Make RBAC default authz mode

moelsayed:


cluster.yml
```diff
@@ -22,7 +22,9 @@ network:
 ssh_key_path: ~/.ssh/test
 ignore_docker_version: false
 
-# Kubernetes authorization mode, currently only `rbac` is supported
+# Kubernetes authorization mode, currently only `rbac` is supported and enabled by default.
+# Use `none` to disable authorization
```
alena1108: none can be treated as absent here. Can we say Use mode: none to disable authorization?

cluster/cluster.go
```diff
@@ -246,10 +250,13 @@ func getLocalAdminConfigWithNewAddress(localConfigPath, cpAddress string) string
 		string(config.KeyData))
 }
 
-func (c *Cluster) ApplyRBACResources() error {
+func (c *Cluster) ApplyAuthzResources() error {
 	if err := authz.ApplyJobDeployerServiceAccount(c.LocalKubeConfigPath); err != nil {
 		return fmt.Errorf("Failed to apply the ServiceAccount needed for job execution: %v", err)
 	}
+	if c.Authorization.Mode == NoneAuthorizationMode {
```
alena1108: shouldn't we check that before sending ApplyJobDeployerServiceAccount?

cluster.yml
```diff
@@ -22,7 +22,9 @@ network:
 ssh_key_path: ~/.ssh/test
 ignore_docker_version: false
 
-# Kubernetes authorization mode, currently only `rbac` is supported
+# Kubernetes authorization mode, currently only `rbac` is supported and enabled by default.
```
alena1108: Use ';' instead of ','

cluster/cluster.go
moelsayed: No we need that anyway because it's added to the job manifest. If it's removed, the job will break.

PR #165, opened 2017-12-18
Update README.md (cluster.yaml -> cluster.yml)

xaka:
The default file name is cluster.yml, but in documentation examples it's cluster.yaml, which makes overall setup process to be not that smooth as it could be.


galal-hussein:
@xaka thanks for your contribution, but this PR is duplicate of this PR https://github.com/rancher/rke/pull/158

xaka:
D'oh! Many apologizes for the noise...

galal-hussein:
@xaka no problem at all :)

PR #166, opened 2017-12-19
Enable PodSecurityPolicy

moelsayed:


authz/manifests.go
```diff
@@ -39,4 +39,70 @@ subjects:
 - kind: ServiceAccount
   namespace: kube-system
   name: rke-job-deployer`
+
+	// DefaultPodSecurityPolicy is very permissive, look into reducing this.
```
alena1108: does this comment still apply? given that the policy is limited to kube-system namespace, and we are not gonna reduce the permissions for those

authz/psp.go
```diff
@@ -0,0 +1,35 @@
...
+func ApplyDefaultPodSecurityPolicy(kubeConfigPath string) error {
+	logrus.Infof("[authz] Applying default PodSecurityPolicy")
+	k8sClient, err := k8s.NewClient(kubeConfigPath)
+	if err != nil {
+		return err
+	}
+	if err := k8s.UpdatePodSecurityPolicyFromYaml(k8sClient, DefaultPodSecurityPolicy); err != nil {
+		return err
+	}
+	logrus.Infof("[authz] Default PodSecurityPolicy created successfully")
```
alena1108: Nit: applied instead of created

k8s/role.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+func UpdateRoleBindingFromYaml(k8sClient *kubernetes.Clientset, roleBindingYaml string) error {
+	roleBinding := rbacv1.RoleBinding{}
+	err := decodeYamlResource(&roleBinding, roleBindingYaml)
+	if err != nil {
+		return err
+	}
+
+	for retries := 0; retries <= 5; retries++ {
```
alena1108: This retry block is being repeated in several places. Consider moving it to its own separate method, and passing func() as an argument

Issue #152, opened 2017-12-15
Add kubenretes components validation


galal-hussein:
RKE code should make sure that the deployed component is up and running before deploying the next component, rke should check /healthz endpoints and should check if etcd cluster healthy

sheng-liang:
I think RKE code needs to go one step further. Requirements like required Docker version or open ports should be checked at a separate initial phase prior to deploying Kubernetes components. For example, perhaps RKE can deploy test containers on each node and run test scripts in these containers to make sure ports are open.

soumyalj:
Tested with rke version v0.0.9-dev. 

1. Health check is being performed after components kube-controller, kube-api, kubelet, scheduler, kube-proxy are deployed. The log messages are displayed as shown in the snippet below:
```
INFO[0116] [healthcheck] Start Healthcheck on service [kubelet] on host [xyz] 
INFO[0120] [worker] Successfully pulled image [rancher/k8s:v1.8.3-rancher2] on host [abc] 
INFO[0120] [worker] Successfully started [kubelet] container on host [abc] 
INFO[0120] [healthcheck] Start Healthcheck on service [kubelet] on host [abc] 
INFO[0123] [healthcheck] service [kubelet] on host [abc] is healthy 
INFO[0123] [worker] Checking image [rancher/k8s:v1.8.3-rancher2] on host [abc 
INFO[0123] [worker] No pull necessary, image [rancher/k8s:v1.8.3-rancher2] exists on host [abc] 
INFO[0124] [worker] Successfully started [kube-proxy] container on host [abc] 
INFO[0124] [healthcheck] Start Healthcheck on service [kube-proxy] on host [abc] 
INFO[0125] [healthcheck] service [kube-proxy] on host [abc] is healthy 
INFO[0127] [healthcheck] service [kubelet] on host [abc] is healthy 
INFO[0127] [worker] Checking image [rancher/k8s:v1.8.3-rancher2] on host [xyz] 
INFO[0127] [worker] No pull necessary, image [rancher/k8s:v1.8.3-rancher2] exists on host [abc] 
INFO[0128] [worker] Successfully started [kube-proxy] container on host [xyz] 
INFO[0128] [healthcheck] Start Healthcheck on service [kube-proxy] on host [xyz] 
INFO[0129] [healthcheck] service [kube-proxy] on host [xyz] is healthy 
INFO[0129] [worker] Successfully started Worker Plane.. 
INFO[0129] [network] Setting up network plugin: flannel 
[...]

PR #170, opened 2017-12-19
Add healthcheck for services components

galal-hussein:
https://github.com/rancher/rke/issues/152

### DIFF
--- a/cmd/up.go
+++ b/cmd/up.go
@@ -31,10 +31,10 @@ func UpCommand() cli.Command {
 	}
 }
 
-func ClusterUp(rkeConfig *v3.RancherKubernetesEngineConfig, dialerFactory hosts.DialerFactory) (string, string, string, string, error) {
+func ClusterUp(rkeConfig *v3.RancherKubernetesEngineConfig, dockerDialerFactory, healthcheckDialerFactory hosts.DialerFactory) (string, string, string, string, error) {
 	logrus.Infof("Building Kubernetes cluster")
 	var APIURL, caCrt, clientCert, clientKey string
-	kubeCluster, err := cluster.ParseCluster(rkeConfig, clusterFilePath, dialerFactory)
+	kubeCluster, err := cluster.ParseCluster(rkeConfig, clusterFilePath, dockerDialerFactory, healthcheckDialerFactory)
 	if err != nil {
 		return APIURL, caCrt, clientCert, clientKey, err
 	}
@@ -113,6 +113,6 @@ func clusterUpFromCli(ctx *cli.Context) error {
 	if err != nil {
 		return fmt.Errorf("Failed to parse cluster file: %v", err)
 	}
-	_, _, _, _, err = ClusterUp(rkeConfig, nil)
+	_, _, _, _, err = ClusterUp(rkeConfig, nil, nil)
 	return err
 }
--- a/hosts/dialer.go
+++ b/hosts/dialer.go
@@ -24,10 +24,22 @@ func SSHFactory(h *Host) (func(network, address string) (net.Conn, error), error
 		host:   h,
 		signer: key,
 	}
-	return dialer.Dial, nil
+	return dialer.DialDocker, nil
 }
 
-func (d *dialer) Dial(network, addr string) (net.Conn, error) {
+func HealthcheckFactory(h *Host) (func(network, address string) (net.Conn, error), error) {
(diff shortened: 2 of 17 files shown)


cmd/up.go
```diff
@@ -31,7 +31,7 @@ func UpCommand() cli.Command {
 	}
 }
 
-func ClusterUp(rkeConfig *v3.RancherKubernetesEngineConfig, dialerFactory hosts.DialerFactory) (string, string, string, string, error) {
+func ClusterUp(rkeConfig *v3.RancherKubernetesEngineConfig, dialerFactory, healthcheckFactory hosts.DialerFactory) (string, string, string, string, error) {
 	logrus.Infof("Building Kubernetes cluster")
 	var APIURL, caCrt, clientCert, clientKey string
 	kubeCluster, err := cluster.ParseCluster(rkeConfig, clusterFilePath, dialerFactory)
```
alena1108: may be it would make sense to set the healthcheckFactory on per cluster level instead of passing it as an argument to DeployClusterPlanes

hosts/dialer.go
```diff
@@ -49,6 +61,25 @@ func (d *dialer) Dial(network, addr string) (net.Conn, error) {
...
+	cfg, err := makeSSHConfig(d.host.User, d.signer)
+	if err != nil {
+		return nil, fmt.Errorf("Error configuring SSH: %v", err)
+	}
+	// Establish connection with SSH server
+	conn, err := ssh.Dial("tcp", sshAddr, cfg)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to dial ssh using address [%s]: %v", sshAddr, err)
+	}
+	remote, err := conn.Dial("tcp", fmt.Sprintf("localhost:%d", d.host.HealthcheckPort))
```
alena1108: why localhost?

services/healthcheck.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/rancher/rke/hosts"
+	"github.com/sirupsen/logrus"
+)
+
+func runHealthcheck(host *hosts.Host, port int, useTLS bool, serviceName string, healthcheckDialerFactory hosts.DialerFactory) error {
+	logrus.Infof("[healthcheck] Start Healthcheck on service [%s] on host [%s]", serviceName, host.Address)
+	var logErr error
+	proto := "http"
+	for retries := 0; retries <= 5; retries++ {
+		client, err := getHealthCheckHTTPClient(host, port, healthcheckDialerFactory)
```
alena1108: Do you want to sleep between the retries? Also can we extract the code called from the for loop?

alena1108: Ok, I see that we do sleep in one of the errors. I think we should sleep in any error case. Also is there a reason the sleep is 5 seconds? sleeping for 25 seconds on per service basis seems a bit extreme to me.

services/controlplane.go
```diff
@@ -20,17 +20,17 @@ func RunControlPlane(controlHosts, etcdHosts []*hosts.Host, controlServices v3.R
 			return err
 		}
 		// run kubeapi
-		err := runKubeAPI(host, etcdHosts, controlServices.KubeAPI, authorizationMode)
+		err := runKubeAPI(host, etcdHosts, controlServices.KubeAPI, authorizationMode, df)
```
alena1108: I don't want to sound as a syntax nazi, but can we use more descriptive name instead of df? May be dFactory?

services/healthcheck.go
galal-hussein: got it, i will make this change

services/healthcheck.go
```diff
@@ -0,0 +1,73 @@
...
+func getHealthz(client *http.Client, useTLS bool, serviceName, hostAddress string) error {
+	proto := HTTPProtoPrefix
+	if useTLS {
+		proto = HTTPSProtoPrefix
+	}
+	resp, err := client.Get(fmt.Sprintf("%s%s%s", proto, HealthzAddress, HealthzEndpoint))
+	if err != nil {
+		return fmt.Errorf("Failed to check %s for service [%s] on host [%s]: %v", HealthzEndpoint, serviceName, hostAddress, err)
+	}
+	statusBody, _ := ioutil.ReadAll(resp.Body)
```
alena1108: we don't have to read the body. Or you can do it only when the error happens (response is not ok)

Issue #148, opened 2017-12-14
Does etcd not use https?


DiamondYuan:
**RKE version:**
v0.8

```
eac3e4f2005a        quay.io/coreos/etcd:latest                                                                                                                   "/usr/local/bin/et..."   2 days ago          Up 52 minutes       0.0.0.0:2379-2380->2379-2380/tcp   etcd
```

```
root@docker-host:~# docker exec eac3 ps -fe
PID   USER     TIME   COMMAND
    1 root       1:02 /usr/local/bin/etcd --name=etcd-192.168.10.67 --data-dir=/etcd-data --advertise-client-urls=http://192.168.10.67:2379,http://192.168.10.67:4001 --listen-client-urls=http://0.0.0.0:2379 --initial-advertise-peer-urls=http://192.168.10.67:2380 --listen-peer-urls=http://0.0.0.0:2380 --initial-cluster-token=etcd-cluster-1 --initial-cluster=etcd-192.168.10.67=http://192.168.10.67:2380,etcd-192.168.10.68=http://192.168.10.68:2380,etcd-192.168.10.69=http://192.168.10.69:2380 --initial-cluster-state=new
   18 root       0:00 ps -fe
```

DiamondYuan:
if one etcd nodes is delete.how can i recovery?

galal-hussein:
@DiamondYuan Currently we don't use tls connection for etcd.

For the second question, you can recover etcd manually, but rke doesn't currently do any add/remove operations regarding etcd nodes.

deniseschannon:
Available on master of RKE

moelsayed:
Tested using master:
- I was able to confirm etcd is working over tls. 
- I was able to add and remove 2 etcd nodes to a 3 etcd-nodes cluster with no problem.

PR #174, opened 2017-12-21
Add TLS to etcd

galal-hussein:
https://github.com/rancher/rke/issues/148


network/calico.go
```diff
@@ -73,9 +77,9 @@ data:
 
   # If you're using TLS enabled etcd uncomment the following.
```
alena1108: Is this comment still relevant? Looks like it should be the other way around

pki/constants.go
```diff
@@ -58,4 +58,8 @@ const (
 	KubeAdminOrganizationName = "system:masters"
 	KubeAdminConfigPrefix     = ".kube_config_"
 	KubeAdminConfigENVName    = "KUBECFG_ADMIN"
+
+	EtcdName              = "etcd"
```
alena1108: We should perhaps name it similar to KubeNodeCertENVName, to reflect the purpose

pki/deploy.go
```diff
@@ -61,6 +61,29 @@ func DeployCertificatesOnWorkers(workerHosts []*hosts.Host, crtMap map[string]Ce
...
+	// list of certificates that should be deployed on the etcd
+	crtList := []string{
+		CACertName,
+	}
+	for i := range etcdHosts {
+		crtList = append(crtList, fmt.Sprintf("%s-%d", EtcdName, i))
+	}
+	env := []string{}
+	for _, crtName := range crtList {
+		c := crtMap[crtName]
```
alena1108: here we expect the key in crdMap be the same as crt name we generate on the line 70. Consider moving the generation code to its separate method so you don't have to maintain it in 2 places

alena1108: Also is it possible that the cert `crtName` does not exist in map? If yes, we should handle it

pki/util.go
```diff
@@ -0,0 +1,159 @@
...
+		return nil, nil, fmt.Errorf("Failed to generate private key for kube-apiserver certificate: %v", err)
+	}
+	caConfig := cert.Config{
+		CommonName: EtcdName,
+		Usages:     []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth},
+		AltNames:   *altNames,
+	}
+	etcdCert, err := cert.NewSignedCert(caConfig, rootKey, caCrt, caKey)
+	if err != nil {
+		return nil, nil, fmt.Errorf("Failed to generate kube-apiserver certificate: %v", err)
```
alena1108: message seems to be for kube-apisrever certificate. Also shouldn't certificate generation should be somewhat generic? And called for all the services. Only Cname would be different, but you can pass it in as an argument.

services/etcd.go
```diff
@@ -7,15 +7,17 @@ import (
...
 	"github.com/sirupsen/logrus"
 )
 
 func RunEtcdPlane(etcdHosts []*hosts.Host, etcdService v3.ETCDService) error {
 	logrus.Infof("[%s] Building up Etcd Plane..", ETCDRole)
 	initCluster := getEtcdInitialCluster(etcdHosts)
-	for _, host := range etcdHosts {
-		imageCfg, hostCfg := buildEtcdConfig(host, etcdService, initCluster)
+	for i, host := range etcdHosts {
+		nodeName := fmt.Sprintf("%s-%d", pki.EtcdKeyCertPathPrefix, i)
```
alena1108: yea, this part, I've seen it in the code somewhere already :) https://github.com/rancher/rke/pull/174/files#diff-fc62d6e548ca9429f12a9edf0f3a8415R50 Consider extracting

services/etcd.go
```diff
@@ -81,7 +94,7 @@ func buildEtcdConfig(host *hosts.Host, etcdService v3.ETCDService, initCluster s
 func GetEtcdConnString(hosts []*hosts.Host) string {
 	connString := ""
 	for i, host := range hosts {
-		connString += "http://" + host.InternalAddress + ":2379"
+		connString += "https://" + host.InternalAddress + ":2379"
```
alena1108: better use fmt.sprintf construct for concatenation

pki/util.go
```diff
@@ -0,0 +1,195 @@
...
+
+func getKeyEnvFromEnv(env string) string {
+	return fmt.Sprintf("%s_KEY", env)
+}
+
+func getConfigEnvFromEnv(env string) string {
+	return fmt.Sprintf("KUBECFG_%s", env)
+}
+
+func GetEtcdCrtName(index int) string {
```
alena1108: What would be the index here? is it the order in which etcd hosts are defined in cluster spec? What would happen if the order get changed? If that is the point of concern, should we rely on something else but the index? Also think that in the future we would need a cert per worker node, so see if there could be any generic approach to naming them.

pki/deploy.go
```diff
@@ -157,34 +182,36 @@ func DeployCertificatesOnHost(ctx context.Context, host *hosts.Host, crtMap map[
 	return doRunDeployer(ctx, host, env, certDownloaderImage)
 }
 
-func FetchCertificatesFromHost(ctx context.Context, host *hosts.Host, image, localConfigPath string) (map[string]CertificatePKI, error) {
+func FetchCertificatesFromHost(ctx context.Context, etcdHosts []*hosts.Host, host *hosts.Host, image, localConfigPath string) (map[string]CertificatePKI, error) {
```
alena1108: if this method is generic, should we consider renaming the param etcdHosts?

pki/pki.go
```diff
@@ -29,287 +28,117 @@ type CertificatePKI struct {
...
-		KeyPath:       KubeSchedulerKeyPath,
-		ConfigEnvName: KubeSchedulerConfigENVName,
-		ConfigPath:    KubeSchedulerConfigPath,
-	}
+	certs[KubeSchedulerCertName] = ToCertObject(KubeSchedulerCertName, "", "", kubeSchedulerCrt, kubeSchedulerKey)
 
 	// generate Kube Proxy certificate and key
 	log.Infof(ctx, "[certificates] Generating Kube Proxy certificates")
-	kubeProxyCrt, kubeProxyKey, err := generateClientCertAndKey(caCrt, caKey, KubeProxyCommonName, []string{})
+	kubeProxyCrt, kubeProxyKey, err := GenerateSignedCertAndKey(caCrt, caKey, nil, nil, getDefaultCN(KubeProxyCertName), nil, false)
```
alena1108: Nit: when generate method signature, try to put  params that are most likely will not be nil, first.

pki/util.go
```diff
@@ -0,0 +1,208 @@
...
+
+func GetKeyTempPath(name string) string {
+	return fmt.Sprintf("%s%s-key.pem", TempCertPath, name)
+}
+
+func GetConfigTempPath(name string) string {
+	return fmt.Sprintf("%skubecfg-%s.yaml", TempCertPath, name)
+}
+
+func ToCertObject(componentName, commonName, ouName string, cert *x509.Certificate, key *rsa.PrivateKey) CertificatePKI {
```
alena1108: is this function ever called from other packages? if not, consider reducing the visibility.

galal-hussein: Yeah unfortunately I call ToCertObject from cluster package when i am regenerating kubeAPI certificate, we do that because when adding new kubeapi node, we need to create a new certificate for it with its IP (internal and public) in the ALTNames of the certificate

cluster/hosts.go
```diff
@@ -80,13 +80,13 @@ func (c *Cluster) SetUpHosts(ctx context.Context) error {
 		if err != nil {
 			return err
 		}
-		// Deploying worker certs on etcd hosts as well
-		err = pki.DeployCertificatesOnWorkers(ctx, c.EtcdHosts, c.Certificates, c.SystemImages[CertDownloaderImage])
+		// Deploying etcd certificates
+		err = pki.DeployCertificatesOnEtcd(ctx, c.EtcdHosts, c.Certificates, c.SystemImages[CertDownloaderImage])
```
alena1108: nit: may be we can write it in one line `if err := blah ; err !=nil

cluster/hosts.go
```diff
@@ -80,13 +80,13 @@ func (c *Cluster) SetUpHosts(ctx context.Context) error {
...
 		if err != nil {
 			return err
 		}
 
-		err = pki.DeployAdminConfig(ctx, c.Certificates[pki.KubeAdminCommonName].Config, c.LocalKubeConfigPath)
+		err = pki.DeployAdminConfig(ctx, c.Certificates[pki.KubeAdminCertName].Config, c.LocalKubeConfigPath)
```
alena1108: same here

PR #177, opened 2017-12-21
Minor fixes

### LINKED ISSUE #175: RKE config: minor improvements are needed
1) Remove redundant `config` from help:

```
COMMANDS:
     up              Bring the cluster up
     remove          Teardown the cluster and clean cluster nodes
     version         Show cluster Kubernetes version
     config, config  Setup cluster configuration
     help, h         Shows a list of commands or help for one command
```
2) The default number of hosts should be 1
```
Number of Hosts [3]
```
3) SSH Private Key Path of host should be defaulted to Cluster level ssh private key path.
4) SSH Private Key of host should not be needed when `SSH Private Key Path of host` is defined.
5) We should display the default + all available options for networking plugin:

`Network Plugin Type [none]`
6) Authorization mode is on by default, and config doesn't display this fact:
`Authorization Mode [none]`

galal-hussein:
#175

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -23,10 +23,9 @@ const (
 
 func ConfigCommand() cli.Command {
 	return cli.Command{
-		Name:      "config",
-		ShortName: "config",
-		Usage:     "Setup cluster configuration",
-		Action:    clusterConfig,
+		Name:   "config",
+		Usage:  "Setup cluster configuration",
+		Action: clusterConfig,
 		Flags: []cli.Flag{
 			cli.StringFlag{
 				Name:  "name,n",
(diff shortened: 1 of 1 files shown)


cmd/config.go
```diff
@@ -164,13 +163,21 @@ func getHostConfig(reader *bufio.Reader, index int) (*v3.RKEConfigNode, error) {
 	if err != nil {
 		return nil, err
 	}
-	host.SSHKeyPath = sshKeyPath
-
-	sshKey, err := getConfig(reader, fmt.Sprintf("SSH Private Key of host (%s)", address), "")
-	if err != nil {
-		return nil, err
+	if len(sshKeyPath) == 0 {
+		fmt.Printf("[-] You have entered empty SSH key path\n")
```
alena1108: we should perhaps identify our next action on empty ssh key path. Similar the way you've done on line 173

PR #179, opened 2017-12-22
Create a general retry function for k8s requests

moelsayed:


k8s/clusterrole.go
```diff
@@ -1,36 +1,31 @@
...
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	yamlutil "k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/client-go/kubernetes"
 )
 
 func UpdateClusterRoleBindingFromYaml(k8sClient *kubernetes.Clientset, clusterRoleBindingYaml string) error {
 	clusterRoleBinding := rbacv1.ClusterRoleBinding{}
-	decoder := yamlutil.NewYAMLToJSONDecoder(bytes.NewReader([]byte(clusterRoleBindingYaml)))
-	err := decoder.Decode(&clusterRoleBinding)
+	err := decodeYamlResource(&clusterRoleBinding, clusterRoleBindingYaml)
```
galal-hussein: can you make this if `err:= decode...; err !=nil {` instead

k8s/k8s.go
```diff
@@ -2,12 +2,20 @@ package k8s
...
 	"bytes"
+	"time"
 
 	yamlutil "k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
 )
 
+const (
+	DefaultRetries      = 5
```
alena1108: local constants should start with lowercase

Issue #153, opened 2017-12-15
Addons: consider switching from using manifests 


alena1108:
We should come up with a better way of managing addons files. Templates perhaps? And see if we can keep them in the same folder under the root, and import it as a package to the place where it needs to be modified.

galal-hussein:
@soumyalj to verify this issue, just make sure that system and user addons are deployed normally as well as the network plugin pods.

soumyalj:
Tested with rke version v0.0.9-dev. Dashboard addon could be deployed on a cluster with rbac disabled.
```
kubectl --kubeconfig .kube_config_clusterdigworking.yml get pod --namespace kube-system
NAME                                   READY     STATUS    RESTARTS   AGE
kube-dns-778977457c-x2dd9              3/3       Running   0          29s
kube-dns-autoscaler-5cb4f55974-5v67k   1/1       Running   0          30s
kube-flannel-7cmkj                     2/2       Running   0          34s
kube-flannel-b4m5c                     2/2       Running   0          34s
kube-flannel-skc4r                     2/2       Running   0          34s
kubernetes-dashboard-747c4f7cf-5hfnr   1/1       Running   0          18s
```
Dashboard addon could be deployed on a cluster with rbac enabled.
```
$ kubectl --kubeconfig .kube_config_clusterwithrbacdashboard.yml get pod --namespace kube-system
NAME                                   READY     STATUS    RESTARTS   AGE
kube-dns-778977457c-sxlmh              3/3       Running   0          2m
kube-dns-autoscaler-5cb4f55974-l8xh5   1/1       Running   0          2m
kube-flannel-gj6k7                     2/2       Running   0          2m
kube-flannel-lnbsn                     2/2       Running   0          2m
kube-flannel-vnh4b                     2/2       Running   0          2m
kubernetes-dashboard-747c4f7cf-kztrc   1/1       Running   0          2m

```

PR #181, opened 2017-12-22
Use Go templates for addons, network plugins and other manifests

moelsayed:
https://github.com/rancher/rke/issues/153


cluster/network.go
```diff
@@ -32,6 +32,26 @@ const (
 	WeaveNetworkPlugin = "weave"
 	WeaveImage         = "weave_node_image"
 	WeaveCNIImage      = "weave_cni_image"
+
+	EtcdEndpoints    = "EtcdEndpoints"
```
galal-hussein: Nit: Can you add proper comments for these constants

templates/weave.go
```diff
@@ -1,13 +1,7 @@
...
 
-import "github.com/rancher/rke/services"
-
-func GetWeaveManifest(weaveConfig map[string]string) string {
-	rbacConfig := ""
-	if weaveConfig[RBACConfig] == services.RBACAuthorizationMode {
-		rbacConfig = getWeaveRBACManifest()
-	}
-	return `
+const WeaveTemplate = `
```
galal-hussein: Weave template doesn't contain if condition for RBAC

moelsayed: Fixed.

Issue #182, opened 2017-12-22
Save certificate/state right after deploying control plane


galal-hussein:
There is no point of waiting for worker plane to be deployed to save the kubernetes certificates and state to k8s, we should be able to include this stage right after deploying the control plane

soumyalj:
Tested with rke version v0.0.9-dev. Deployed a cluster with one [controlplane, etcd] node and 2 worker nodes. It was observed that certificates and cluster state were saved right after control plane deployment. Snippet from the cluster deployment
```
INFO[0207] [authz] Creating rke-job-deployer ServiceAccount 
INFO[0207] [authz] rke-job-deployer ServiceAccount created successfully 
INFO[0207] [authz] Creating system:node ClusterRoleBinding 
INFO[0207] [authz] system:node ClusterRoleBinding created successfully 
INFO[0207] [certificates] Save kubernetes certificates as secrets 
INFO[0225] [certificates] Successfully saved certificates as kubernetes secret [k8s-certs] 
INFO[0225] [state] Saving cluster state to Kubernetes   
INFO[0226] [state] Successfully Saved cluster state to Kubernetes ConfigMap: cluster-state 
INFO[0226] [worker] Building up Worker Plane..    
INFO[0226] [sidekick] Sidekick container already created on host [xyz] 
INFO[0226] [worker] Checking image [rancher/k8s:v1.8.3-rancher2] on host [xyz] 
INFO[0226] [worker] No pull necessary, image [rancher/k8s:v1.8.3-rancher2] exists on host [xyz] 
```

PR #185, opened 2017-12-26
Save certificates/state after deploying control plane

galal-hussein:
https://github.com/rancher/rke/issues/182

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -52,32 +52,35 @@ const (
 	NoneAuthorizationMode      = "none"
 )
 
-func (c *Cluster) DeployClusterPlanes() error {
-	// Deploy Kubernetes Planes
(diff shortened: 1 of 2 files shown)


cluster/cluster.go
```diff
@@ -52,32 +52,35 @@ const (
...
 		c.SystemImages[ServiceSidekickImage],
 		c.Authorization.Mode,
-		c.HealthcheckDialerFactory)
-	if err != nil {
+		c.HealthcheckDialerFactory); err != nil {
 		return fmt.Errorf("[controlPlane] Failed to bring up Control Plane: %v", err)
 	}
-	err = c.ApplyAuthzResources()
-	if err != nil {
+	// Apply RBAC resources after deploying controlplane
```
moelsayed: Nit: not just RBAC, authz configuration in general, even if RBAC is disabled, or another authz mode is enabled.

PR #199, opened 2018-01-03
Support for Calico on GCE

superseb:
https://github.com/rancher/rke/issues/200

### DIFF
--- a/templates/calico.go
+++ b/templates/calico.go
@@ -469,13 +469,12 @@ metadata:
   namespace: kube-system
 
 
-{{if eq .CloudProvider "aws"}}
-## aws stuff here
+{{if ne .CloudProvider ""}}
 ---
 kind: ConfigMap
 apiVersion: v1
 metadata:
-  name: aws-ippool
+  name: {{.CloudProvider}}-ippool
   namespace: kube-system
 data:
   aws-ippool: |-
@@ -484,6 +483,8 @@ data:
     metadata:
       cidr: {{.ClusterCIDR}}
     spec:
+      ipip:
+        enabled: true
       nat-outgoing: true
 ---
 apiVersion: v1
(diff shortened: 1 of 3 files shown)


templates/calico.go
```diff
@@ -515,4 +515,52 @@ spec:
...
+{{if eq .CloudProvider "gce"}}
+## gce stuff here
+---
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: gce-ippool
+  namespace: kube-system
+data:
+  gce-ippool: |-
```
galal-hussein: Since the ippool is very similar to aws pool, you can do something like {{ .CloudProvider }}-ippool and use one manifest for config map and one calicoctl pod

templates/calico.go
```diff
@@ -515,4 +515,52 @@ spec:
...
+        configMapKeyRef:
+          name: calico-config
+          key: etcd_endpoints
+    volumeMounts:
+    - name: ippool-config
+      mountPath: /root/
+  volumes:
+  - name: ippool-config
+    configMap:
+      name: gce-ippool
```
galal-hussein: same thing here, you can just change gce with the variable

templates/calico.go
```diff
@@ -515,4 +515,52 @@ spec:
...
+  name: gce-ippool
+  namespace: kube-system
+data:
+  gce-ippool: |-
+    apiVersion: v1
+    kind: ipPool
+    metadata:
+      cidr: {{.ClusterCIDR}}
+    spec:
+      ipip:
```
superseb: @galal-hussein The difference is here, I can make this part conditional on cloud provider but thought it would be more clear if we kept it separate. Let me know what you want to do.

superseb:
@galal-hussein This has been rewritten and tested on AWS and GCE

PR #204, opened 2018-01-05
Automate publishing

cloudnautique:
The PR will automate the following when a tag is placed on Master.

This will create the GH release if it doesn't exist yet. 
This will publish binaries from the ./bin/* directory to the GH release.
This will push a rancher/rke-certs-deployer:${tag} image to docker hub.


.drone.yml
```diff
@@ -7,3 +7,44 @@ pipeline:
...
+    - dapper build
+    when:
+      branch: master
+      event: tag
+   
+
+  stage-binaries:
+    image: rancher/dapper:1.11.2
+    commands:
+    - cp -r ./bin/* ./package/
```
galal-hussein: With CROSS=1 make will create the binaries in build/bin

.drone.yml
```diff
@@ -7,3 +7,44 @@ pipeline:
...
+
+  github_binary_release:
+    image: plugins/github-release
+    files: bin/*
+    secrets: [github_token]
+    when:
+      branch: master
+      event: tag
+   
+  publish-rke-certs-deployer:
```
galal-hussein: Regarding the other docker images, should they also be included as a part of publishing images to docker hub

galal-hussein: nginx proxy name: rancher/rke-nginx-proxy:0.1.0
rancher cert deployer: rancher/rke-cert-deployer:0.1.0
service side kick: rancher/rke-service-sidekick:0.1.0

.drone.yml
cloudnautique: In the staging step, this just moves the binary for the Docker image. The `scripts/build` still build bin/* binaries, so I think this part is OK.

galal-hussein: okay, thanks

Issue #192, opened 2018-01-02
Regression with using passphrased SSH keys


galal-hussein:
**RKE version:**
v0.0.9-dev

Using ssh keys protected by passphrase causes two issues:

1- unexpected error:
```
INFO[0000] [dialer] Setup tunnel for host [x.x.x.x]
Passphrase for Private SSH Key:
FATA[0002] Failed to set up SSH tunneling for Etcd host [x.x.x.x]: Can't establish dialer connection: Failed to parse the private key: ssh: cannot decode encrypted private keys
```
this caused by wrong return of err instead of nil

2- After the first issue is resolved, user will have difficulty of running rke because rke will keep asking for passphrase due to the healthcheck addition

galal-hussein:
CC @vincent99

galal-hussein:
Can be tested with v0.0.10-dev

soumyalj:
Reproduced the issue using encrypted ssh keys for hosts with rke version 0.0.9-dev
```
$ ./rke_darwin-amd64-0.0.9-dev up --config cluster.yml 
INFO[0000] Building Kubernetes cluster                  
INFO[0000] [dialer] Setup tunnel for host [x.y.z.a] 
Passphrase for Private SSH Key: 
FATA[0005] Failed to set up SSH tunneling for Etcd host [x.y.z.a]: Can't establish dialer connection: Failed to parse the private key: ssh: cannot decode encrypted private keys
```
Tested with rke version v0.0.12-dev
1. Generated encrypted ssh keys and specified the ssh_key_path in the cluster.yml file. 
2. Created cluster with two nodes using ./rke up command
3. Cluster creation was successful

marcusianlevine:
This seems to have regressed again. 

```
$ rke --version
rke version v0.1.7
```

Using an SSH key with passphrase I get:

```
INFO[0000] Building Kubernetes cluster
INFO[0000] [dialer] Setup tunnel for host [docker-1.***]
WARN[0000] Failed to set up SSH tunneling for host [docker-1.***]: Can't retrieve Docker Info: error during connect: Get http://%2Fvar%2Frun%2Fdocker.sock/v1.24/info: Failed to dial ssh using address [docker-1.***:22]: Error configuring SSH: ssh: cannot decode encrypted private keys
```

_(hostnames redacted with ***)_

bmenant:
I confirm the regression with `rke_linux-amd64` version **v0.1.7**.
[...]

PR #210, opened 2018-01-09
Fix regression with passphrased keys

galal-hussein:
#192

### DIFF
--- a/hosts/tunnel.go
+++ b/hosts/tunnel.go
@@ -77,14 +77,14 @@ func makeSSHConfig(user string, signer ssh.Signer) (*ssh.ClientConfig, error) {
 	return &config, nil
 }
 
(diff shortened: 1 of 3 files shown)


hosts/tunnel.go
```diff
@@ -77,31 +77,38 @@ func makeSSHConfig(user string, signer ssh.Signer) (*ssh.ClientConfig, error) {
...
 	// parse encrypted key
 	if strings.Contains(err.Error(), "decode encrypted private keys") {
-		fmt.Printf("Passphrase for Private SSH Key: ")
-		passphrase, err := terminal.ReadPassword(int(syscall.Stdin))
-		fmt.Printf("\n")
-		if err != nil {
-			return nil, err
+		var passphrase []byte
+		if len(h.SavedKeyPhrase) == 0 {
+			fmt.Printf("Passphrase for Private SSH Key: ")
```
alena1108: do we want to print it?

PR #220, opened 2018-01-12
Add config dir parameter to cluster up/remove

galal-hussein:


galal-hussein:
will merge this PR with #180

PR #232, opened 2018-01-15
Start containers if stopped and use net=host for etcd

galal-hussein:


docker/docker.go
```diff
@@ -22,11 +22,26 @@ var K8sDockerVersions = map[string][]string{
 }
 
 func DoRunContainer(ctx context.Context, dClient *client.Client, imageCfg *container.Config, hostCfg *container.HostConfig, containerName string, hostname string, plane string) error {
-	isRunning, err := IsContainerRunning(ctx, dClient, hostname, containerName, false)
+	container, err := dClient.ContainerInspect(ctx, containerName)
 	if err != nil {
+		if client.IsErrNotFound(err) {
```
alena1108: can we return right away when error is other than not found? then all the `if not found` code can be moved outside of nested if condition

galal-hussein: okay will fix

alena1108: Also may be extract this logic to a separate method called `createContainer` or something

PR #238, opened 2018-01-16
Add open port checks

moelsayed:


cluster/network.go
```diff
@@ -209,3 +231,236 @@ func (c *Cluster) getNetworkPluginManifest(pluginConfig map[string]string) (stri
...
+func (c *Cluster) checkKubeAPIPort(ctx context.Context) error {
+	log.Infof(ctx, "[network] Checking KubeAPI port Control Plane hosts")
+	for _, host := range c.ControlPlaneHosts {
+		logrus.Debugf("[network] Checking KubeAPI port [%s] on host: %s", KubeAPIPort, host.Address)
+		address := fmt.Sprintf("%s:%s", host.Address, KubeAPIPort)
+		conn, err := net.Dial("tcp", address)
+		if err != nil {
+			return fmt.Errorf("[network] Can't access KubeAPI port [%s] on Control Plane host: %s", KubeAPIPort, host.Address)
+		}
+		conn.Close()
```
alena1108: should we defer conn.Close()? in case error happens, is the conn gonna be nil? if not, we should defer closing it.

cluster/network.go
```diff
@@ -209,3 +231,236 @@ func (c *Cluster) getNetworkPluginManifest(pluginConfig map[string]string) (stri
...
+		PortBindings: nat.PortMap{
+			"1337/tcp": portBindingList,
+		},
+	}
+	hostList := []*hosts.Host{}
+	hostList = append(hostList, c.EtcdHosts...)
+	hostList = append(hostList, c.ControlPlaneHosts...)
+	hostList = append(hostList, c.WorkerHosts...)
+	// little trick to get a unique host list
+	uniqHosts := make(map[*hosts.Host]bool)
```
alena1108: why pointer as a key?

cluster/network.go
```diff
@@ -209,3 +231,236 @@ func (c *Cluster) getNetworkPluginManifest(pluginConfig map[string]string) (stri
...
+	if len(c.EtcdHosts) > 1 {
+		log.Infof(ctx, "[network] Running etcd <-> etcd port checks")
+		for _, host := range c.EtcdHosts {
+			runHost := host
+			errgrp.Go(func() error {
+				return checkPlaneTCPPortsFromHost(ctx, runHost, etcdPortList, c.EtcdHosts, c.SystemImages[AplineImage])
+			})
+		}
+	}
+	if err := errgrp.Wait(); err != nil {
```
alena1108: nit: this code would be hit only when len(c.EtcdHosts) > 1, so may be put it in inside the condition?

cluster/network.go
```diff
@@ -209,3 +231,236 @@ func (c *Cluster) getNetworkPluginManifest(pluginConfig map[string]string) (stri
...
+		return fmt.Errorf("[netwok] Port check for ports: [%s] failed on host: [%s]", strings.Join(portCheckLogs, ", "), host.Address)
+
+	}
+	return nil
+}
+
+func getPortCheckLogs(reader io.ReadCloser) []string {
+	logLines := bufio.NewScanner(reader)
+	hostPortLines := []string{}
+	for logLines.Scan() {
```
alena1108: shall we care about scanner returning an error?

moelsayed: I am not 100% sure, but I added the error check just to be safe

cluster/network.go
moelsayed: Because host lists are list of host pointers, not host objects.

cluster/network.go
moelsayed: If there is an error, conn is nil and the defer panics, so this should be fine.

PR #245, opened 2018-01-18
Fix port-check update issue

moelsayed:
https://github.com/rancher/rancher/issues/10937

### DIFF
--- a/cluster/network.go
+++ b/cluster/network.go
@@ -23,8 +23,10 @@ import (
 const (
 	NetworkPluginResourceName = "rke-network-plugin"
 
-	PortCheckContainer  = "rke-port-checker"
-	PortListenContainer = "rke-port-listener"
+	PortCheckContainer        = "rke-port-checker"
+	EtcdPortListenContainer   = "rke-etcd-port-listener"
+	CPPortListenContainer     = "rke-cp-port-listener"
+	WorkerPortListenContainer = "rke-worker-port-listener"
 
 	KubeAPIPort         = "6443"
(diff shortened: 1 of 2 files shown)


cluster/network.go
```diff
@@ -259,19 +269,71 @@ func (c *Cluster) checkKubeAPIPort(ctx context.Context) error {
...
+	}
+	if err := errgrp.Wait(); err != nil {
+		return err
+	}
+
+	// deploy controlplane listeners
+	controlPlanePortList := []string{
+		KubeAPIPort,
+	}
+	for _, host := range cpHosts {
```
alena1108: the code blocks for deploying portlisten containers look pretty similar. Can we extract it to its own function, and just pass list of hosts/container to deploy as an argument?

cluster/network.go
```diff
@@ -293,37 +355,45 @@ func (c *Cluster) deployTCPPortListeners(ctx context.Context) error {
...
 	if err := errgrp.Wait(); err != nil {
 		return err
 	}
-	log.Infof(ctx, "[network] Port listener containers deployed successfully")
-	return nil
-}
-func (c *Cluster) removeTCPPortListeners(ctx context.Context) error {
-	log.Infof(ctx, "[network] Removing port listener containers")
-	var errgrp errgroup.Group
+	for _, host := range c.ControlPlaneHosts {
```
alena1108: Same for remove. Lets think about extracting similar code, and passing hosts/container to remove as an argument.

PR #246, opened 2018-01-19
Add unschedulable kubelet on etcd nodes

galal-hussein:
https://github.com/rancher/rancher/issues/10894

### DIFF
--- a/cluster/reconcile.go
+++ b/cluster/reconcile.go
@@ -12,6 +12,10 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
+const (
+	taintKey = "node-role.kubernetes.io/etcd"
+)
+
 func ReconcileCluster(ctx context.Context, kubeCluster, currentCluster *Cluster) error {
 	log.Infof(ctx, "[reconcile] Reconciling cluster state")
 	if currentCluster == nil {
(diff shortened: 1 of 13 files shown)


cluster/reconcile.go
```diff
@@ -55,6 +59,15 @@ func reconcileWorker(ctx context.Context, currentCluster, kubeCluster *Cluster,
 			continue
 		}
 	}
+	// attempt to remove unschedulable taint
+	toAddHosts := hosts.GetToAddHosts(currentCluster.WorkerHosts, kubeCluster.WorkerHosts)
+	for _, host := range toAddHosts {
+		if host.IsEtcd {
+			if err := hosts.RemoveTaintFromHost(ctx, host, taintKey, kubeClient); err != nil {
+				log.Warnf(ctx, "[reconcile] Failed to remove unschedulable taint from node [%s]", host.Address)
```
alena1108: we don't fail, just log the failure?

k8s/node.go
```diff
@@ -58,3 +59,41 @@ func IsNodeReady(node v1.Node) bool {
...
+				logrus.Debugf("[hosts] Can't find node by name [%s]", nodeName)
+				return nil
+			}
+			return err
+		}
+		foundTaint := false
+		for i, taint := range node.Spec.Taints {
+			if taint.Key == taintKey {
+				foundTaint = true
+				node.Spec.Taints = append(node.Spec.Taints[:i], node.Spec.Taints[i+1:]...)
```
alena1108: shall we break once taint is found?

PR #276, opened 2018-01-26
Private registry authentication

moelsayed:
https://github.com/rancher/rke/issues/84

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/rancher/rke/authz"
+	"github.com/rancher/rke/docker"
 	"github.com/rancher/rke/hosts"
 	"github.com/rancher/rke/log"
 	"github.com/rancher/rke/pki"
@@ -35,6 +36,7 @@ type Cluster struct {
 	ClusterDNSServer                 string
 	DockerDialerFactory              hosts.DialerFactory
 	LocalConnDialerFactory           hosts.DialerFactory
+	PrivateRegistriesMap             map[string]v3.PrivateRegistry
 }
 
 const (
@@ -51,7 +53,7 @@ const (
 
 func (c *Cluster) DeployControlPlane(ctx context.Context) error {
 	// Deploy Etcd Plane
-	if err := services.RunEtcdPlane(ctx, c.EtcdHosts, c.Services.Etcd, c.LocalConnDialerFactory); err != nil {
+	if err := services.RunEtcdPlane(ctx, c.EtcdHosts, c.Services.Etcd, c.LocalConnDialerFactory, c.PrivateRegistriesMap); err != nil {
 		return fmt.Errorf("[etcd] Failed to bring up Etcd Plane: %v", err)
 	}
 	// Deploy Control plane
@@ -60,7 +62,8 @@ func (c *Cluster) DeployControlPlane(ctx context.Context) error {
 		c.Services,
 		c.SystemImages.KubernetesServicesSidecar,
 		c.Authorization.Mode,
-		c.LocalConnDialerFactory); err != nil {
+		c.LocalConnDialerFactory,
+		c.PrivateRegistriesMap); err != nil {
 		return fmt.Errorf("[controlPlane] Failed to bring up Control Plane: %v", err)
 	}
 	// Apply Authz configuration after deploying controlplane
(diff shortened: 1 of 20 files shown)


docker/docker.go
```diff
@@ -150,8 +154,41 @@ func localImageExists(ctx context.Context, dClient *client.Client, hostname stri
...
 }
 
-func pullImage(ctx context.Context, dClient *client.Client, hostname string, containerImage string) error {
-	out, err := dClient.ImagePull(ctx, containerImage, types.ImagePullOptions{})
+func pullImage(ctx context.Context, dClient *client.Client, hostname string, containerImage string, prs []v3.PrivateRegistry) error {
+
+	pullOptions := types.ImagePullOptions{}
+	prsMap := make(map[string]v3.PrivateRegistry)
+	// build the Private Registry map
+	if len(prs) > 0 {
```
alena1108: @moelsayed do you think it would make sense to pass the map all way through? Right from the very top. So this way you won't have to rebuild it from list every time pullImage is called

docker/docker.go
```diff
@@ -150,8 +154,41 @@ func localImageExists(ctx context.Context, dClient *client.Client, hostname stri
...
+		for _, pr := range prs {
+			if pr.URL == "" {
+				pr.URL = "docker.io"
```
alena1108: docker.io should be defaulted in constant somewhere

docker/docker.go
```diff
@@ -302,3 +339,25 @@ func ReadContainerLogs(ctx context.Context, dClient *client.Client, containerNam
...
 }
+
+func tryRegistryAuth(pr v3.PrivateRegistry) types.RequestPrivilegeFunc {
+	return func() (string, error) {
+		return getRegistryAuth(pr)
+	}
+}
+
+func hasURL(imageName string) bool {
+	return strings.Contains(strings.Split(imageName, "/")[0], ".")
```
alena1108: there should be some standard library for parsing image, in docker, or k8s. Lets try to use it. Relying on prefix might not be a good idea.

docker/docker.go
```diff
@@ -150,8 +154,41 @@ func localImageExists(ctx context.Context, dClient *client.Client, hostname stri
...
+			}
+			pullOptions.RegistryAuth = regAuth
+		}
+	}
+	// We have a registry, but it's not docker.io
+	if hasURL(containerImage) &&
+		!strings.HasPrefix(containerImage, "docker.io") {
+		// this could be public or private, ImagePull() can handle it
+		// if we provide a PrivilegeFunc
+		regURL := strings.Split(containerImage, "/")[0]
```
alena1108: see comment below...there might be some standard library for parsing the docker image, lets find and use it

docker/docker.go
moelsayed: There are a log of places where this call would be triggered. I would  need to rebuild it at every to level call or add it to the cluster option to use it later. I will look into the second option.

docker/docker.go
```diff
@@ -150,8 +158,32 @@ func localImageExists(ctx context.Context, dClient *client.Client, hostname stri
...
-	out, err := dClient.ImagePull(ctx, containerImage, types.ImagePullOptions{})
+func pullImage(ctx context.Context, dClient *client.Client, hostname string, containerImage string, prsMap map[string]v3.PrivateRegistry) error {
+
+	pullOptions := types.ImagePullOptions{}
+	containerNamed, err := ref.ParseNormalizedNamed(containerImage)
+	if err != nil {
+		return err
+	}
+	// We do this if we have some docker.io login information
+	if ref.Domain(containerNamed) == DockerRegistryURL {
```
alena1108: @moelsayed is there a reason why we distinguish between ref.Domain(containerNamed) being == vs != DockerRegistryURL? in both cases can we use `getRegistryAuth(pr)`? Or if there is a big diff between `pullOptions.RegistryAuth` vs `pullOptions.PrivilegeFunc`, can we refactor the code as follows, so it's a bit easier to read:

```
if pr, ok := prsMap[regURL]; ok {
  if regURL {
   // getRegistryAuth
  } else {
   //  tryRegistryAuth
   }
}
```

moelsayed: Internally, `pullOptions.PrivilegeFunc` will generate and set `pullOptions.RegistryAuth` if the initial unauthenticated pull request failed with an Unauthorized error and try again.

However, docker registry API returns a not found error if the image is private, which doesn't trigger the `pullOptions.PrivilegeFunc` call. I am not sure why they do this. Security maybe ?

So, if it's a Docker registry, it could be a private or a public image. I check if I have auth credentials for docker. If I found it, I explicitly generate  `pullOptions.RegistryAuth` and use it directly.

cluster/cluster.go
```diff
@@ -128,6 +133,16 @@ func ParseCluster(
 		c.ConfigPath = DefaultClusterConfig
 	}
 	c.LocalKubeConfigPath = GetLocalKubeConfig(c.ConfigPath, configDir)
+
+	if len(c.PrivateRegistries) > 0 {
```
alena1108: There is no need for this check. If len is 0, the loop below will handle it

PR #279, opened 2018-01-27
Resolve ip to non-loopback interface if --local is used

GregoireW:
If we use "rke up --local" then process use 'localhost' (127.0.0.1)  as a node. 

Issue is some processes launch containers and check in the containers are made to access the node ip (127.0.0.1). This will fail as 127.0.0.1 will resolve to the container itself.

This PR try to fix that by detecting which is the local host name and associated local ip.  If not resolvable it fallback to 'localhost' '127.0.0.1'. 

Note that this takes the 1st resolved ipV4 from hostname.  I just hope that the ip list always begin with ips from hardware interface and not alias one (like docker one...)

### DIFF
--- a/cluster/local.go
+++ b/cluster/local.go
@@ -3,6 +3,9 @@ package cluster
 import (
 	"github.com/rancher/rke/services"
 	"github.com/rancher/types/apis/management.cattle.io/v3"
+	"github.com/sirupsen/logrus"
+	"net"
+	"os"
 )
 
 func GetLocalRKEConfig() *v3.RancherKubernetesEngineConfig {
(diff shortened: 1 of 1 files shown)


rickard-von-essen:
I tried this on a fresh RancherOS VM, but unfortunately this wasn't enough to get `--local` to work.

Running the same build against the same VM from another host worked fine so there is still some issue with `--local` to resolve.

GregoireW:
Not sure it will be resolved here or in a new issue, but can you copy/paste the logs you get with '--local'  if possible in the rancher slack ( \#rke ) or here

rickard-von-essen:
I need to reinitialize the VM and reproduce it again, maybe I'll have some time this weekend.

galal-hussein:
@GregoireW Thanks very much for your PR, sorry i didn't realize that there is an attempt to solve this before putting this PR https://github.com/rancher/rke/pull/283, but i think adding net=host for port checker is enough for local deployment to pass

galal-hussein:
Closing this PR in favor of #283 thanks @GregoireW

Issue #272, opened 2018-01-26
Add ingress controller option to cluster


alena1108:
RKE should start ingress controller (provider is configurable) as a part of cluster.up

galal-hussein:
@moelsayed can you make sure that ingress resources are being created and can be accessed on k8s nodes on ports 80/443

moelsayed:
Tested on v0.1.1-rc1. Created one ingress, and two backend services. I was able to reach the pods using the configured hostname and urls.

PR #290, opened 2018-01-31
Ingress controller

galal-hussein:
depends on https://github.com/rancher/types/pull/159
#272


cluster/cluster.go
```diff
@@ -252,3 +253,32 @@ func (c *Cluster) getUniqueHostList() []*hosts.Host {
 	}
 	return uniqHostList
 }
+
+func (c *Cluster) DeployAddons(ctx context.Context) error {
+	if err := c.DeployK8sAddOns(ctx); err != nil {
+		return err
+	}
+	if err := c.DeployK8sIngress(ctx); err != nil {
```
alena1108: Is there a reason why ingress is treated specially here, and not a part of DeployK8sAddOns (like dns)? it is really similar, and probably should be a part of k8s addons

cluster/ingress.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+type ingressOptions struct {
+	RBACConfig   string
+	Options      map[string]string
+	NodeSelector map[string]string
+}
+
+func (c *Cluster) DeployK8sIngress(ctx context.Context) error {
+	log.Infof(ctx, "[ingress] Setting up %s ingress controller..", c.Ingress.Type)
```
alena1108: no need for `..` :)

k8s/node.go
```diff
@@ -98,3 +100,106 @@ func RemoveTaintFromNodeByKey(k8sClient *kubernetes.Clientset, nodeName, taintKe
...
+		}
+		// Remove Taints from node
+		for i, taintStr := range toDelTaints {
+			if isTaintExist(toTaint(taintStr), node.Spec.Taints) {
+				node.Spec.Taints = append(node.Spec.Taints[:i], node.Spec.Taints[i+1:]...)
+			}
+		}
+
+		//node.Spec.Taints
+		_, err = k8sClient.CoreV1().Nodes().Update(node)
```
alena1108: performance nit: we should only call update when something got changed.

k8s/node.go
```diff
@@ -98,3 +100,106 @@ func RemoveTaintFromNodeByKey(k8sClient *kubernetes.Clientset, nodeName, taintKe
...
+	}
+	return nil
+}
+
+func SyncTaints(k8sClient *kubernetes.Clientset, nodeName string, toAddTaints, toDelTaints []string) error {
+	updated := false
+	var err error
+	var node *v1.Node
+	for retries := 0; retries <= 5; retries++ {
+		node, err = GetNode(k8sClient, nodeName)
```
alena1108: can we extract the code from inside the for loop, to a separate method? so it is easier to read

PR #291, opened 2018-01-31
Add kubernetes version and pin etcd version in cluster.yml

moelsayed:


cluster.yml
```diff
@@ -76,7 +79,7 @@ services:
 
 
 system_images:
-  etcd: quay.io/coreos/etcd:latest
+  etcd: quay.io/coreos/etcd:v3.3
```
superseb: Supported version of etcd is 3.0.17 for 1.8 and 3.1.10 for 1.9. (https://github.com/kubernetes/kubernetes/pull/49393)

Issue #261, opened 2018-01-24
Performance issues


alena1108:
RKE is great, but some things need improvements in terms of speed.

1. Image pull should be done in parallel on all the nodes; perhaps during control plane coming up. So by the time the workers are ready to deploy, we have to make sure that the images are pulled. It means in that in the code we should separate image pull from component deployment.

2. Certificate deployment is being done on per role (component). If node has more than one role, the deployer will run 3 times. We have to run it only once, per node. 

3. Port checks in parallel

4. Saving certificates and the states in parallel.

moelsayed:
Note: Port checking is already running in parallel.

galal-hussein:
version v0.1.1-rc1

I can verify all 4 points with rke:

- k8s image is being pulled on all nodes before deployment in parallel:
```
INFO[0031] Pre-pulling kubernetes images                
DEBU[0031] [pre-deploy] Checking image [docker.io/rancher/k8s:v1.9.2-rancher1-1] on host [x.x.x.x] 
DEBU[0031] [pre-deploy] Checking image [docker.io/rancher/k8s:v1.9.2-rancher1-1] on host [y.y.y.y] 
DEBU[0031] Checking if image [docker.io/rancher/k8s:v1.9.2-rancher1-1] exists on host [z.z.z.z] 
```
- Certificate deployment has been refactored to only run cert deployer once per node
- Port checks is being executed in parallel
- Saving certificate has been verified #293

PR #298, opened 2018-02-01
Refactor certificates deployment

moelsayed:
No. 2 in https://github.com/rancher/rke/issues/261

### DIFF
--- a/cluster/hosts.go
+++ b/cluster/hosts.go
@@ -10,6 +10,7 @@ import (
 	"github.com/rancher/rke/pki"
 	"github.com/rancher/rke/services"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sync/errgroup"
 )
 
 func (c *Cluster) TunnelHosts(ctx context.Context, local bool) error {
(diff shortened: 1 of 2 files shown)


cluster/hosts.go
```diff
@@ -72,15 +73,17 @@ func (c *Cluster) InvertIndexHosts() error {
...
+			runHost := host
+			errgrp.Go(func() error {
+				return pki.DeployCertificatesOnPlaneHost(ctx, runHost, c.EtcdHosts, c.Certificates, c.SystemImages.CertDownloader, c.PrivateRegistriesMap)
+			})
 		}
-		// Deploying etcd certificates
-		if err := pki.DeployCertificatesOnEtcd(ctx, c.EtcdHosts, c.Certificates, c.SystemImages.CertDownloader, c.PrivateRegistriesMap); err != nil {
-			return err
+		if err := errgrp.Wait(); err != nil {
+			return nil
```
galal-hussein: Why are we returning null here?

moelsayed: Wishful thinking!

PR #321, opened 2018-02-06
[RKE Refactor] Generate cluster and node certs

galal-hussein:


pki/util.go
```diff
@@ -207,3 +210,61 @@ func ToCertObject(componentName, commonName, ouName string, cert *x509.Certifica
 func getDefaultCN(name string) string {
 	return fmt.Sprintf("system:%s", name)
 }
+
+func getControlCertKeys(ctx context.Context) []string {
+	return []string{
```
alena1108: why do we need ctx in the signature? can it be removed?

pki/pki.go
```diff
@@ -26,17 +28,13 @@ type CertificatePKI struct {
...
-	return certs, nil
-}
+const (
+	etcdRole    = "etcd"
+	controlRole = "controlplane"
+	workerRole  = "worker"
+)
 
-func generateCerts(ctx context.Context, cpHosts, etcdHosts []*hosts.Host, clusterDomain, localConfigPath string, KubernetesServiceIP net.IP) (map[string]CertificatePKI, error) {
+func GenerateRKECerts(ctx context.Context, rkeConfig v3.RancherKubernetesEngineConfig, configPath, configDir string) (map[string]CertificatePKI, error) {
```
alena1108: lets remove all unused parameters from API signature, like ctx in this case

PR #346, opened 2018-02-13
External etcd

galal-hussein:
RFR


cluster/certificates.go
```diff
@@ -123,11 +123,14 @@ func saveCertToKubernetes(kubeClient *kubernetes.Clientset, crtName string, crt
...
-		"Certificate": cert.EncodeCertPEM(crt.Certificate),
-		"Key":         cert.EncodePrivateKeyPEM(crt.Key),
-		"EnvName":     []byte(crt.EnvName),
-		"KeyEnvName":  []byte(crt.KeyEnvName),
+	secretData := make(map[string][]byte)
+	if crt.Certificate != nil {
+		secretData["Certificate"] = cert.EncodeCertPEM(crt.Certificate)
+		secretData["EnvName"] = []byte(crt.EnvName)
+	}
+	if crt.Key != nil {
```
alena1108: what if the key or cert is nil? Would it be considered as error ,and caught later?

alena1108:
Would be nice to have an example of specifying external etcd

cluster/certificates.go
galal-hussein: no this is not considered an error, its optional to have env to a cert object, the env will cause the cert to be deployed, we may have in the future any case that we need to generate a cert and not deploy it.

PR #365, opened 2018-02-20
Add support for custom WrapTransport for Kubernetes Client

moelsayed:


k8s/k8s.go
```diff
@@ -16,13 +17,24 @@ const (
...
 type k8sCall func(*kubernetes.Clientset, interface{}) error
 
-func NewClient(kubeConfigPath string) (*kubernetes.Clientset, error) {
+type WrapTransport func(rt http.RoundTripper) http.RoundTripper
+
+var K8sClientSet *kubernetes.Clientset
+
+func NewClient(kubeConfigPath string, k8sWrapTransport WrapTransport) (*kubernetes.Clientset, error) {
+	// Do we already have a clientset
+	if K8sClientSet != nil {
```
galal-hussein: This approach may cause problems as sometimes in the code we need to reinitialize the k8s client, for example when deleting cp host (check reconcile code)

PR #370, opened 2018-02-21
Add/Remove etcd idempotency

galal-hussein:


services/etcd.go
```diff
@@ -59,28 +61,35 @@ func RemoveEtcdPlane(ctx context.Context, etcdHosts []*hosts.Host, force bool) e
...
+		}
 		etcdClient, err := getEtcdClient(ctx, host, localConnDialerFactory, cert, key)
 		if err != nil {
 			logrus.Debugf("Failed to create etcd client for host [%s]: %v", host.Address, err)
 			continue
 		}
 		memAPI := etcdclient.NewMembersAPI(etcdClient)
 		if _, err := memAPI.Add(ctx, peerURL); err != nil {
-			logrus.Debugf("Failed to list etcd members from host [%s]: %v", host.Address, err)
+			if strings.Contains(err.Error(), AlreadyExistString) {
```
alena1108: is this line does the same thing? https://github.com/rancher/rke/pull/370/files#diff-cb303a10043d7bf111f0dd3a923627f2R197 // checking if already exists

services/etcd.go
```diff
@@ -140,3 +149,36 @@ func ReloadEtcdCluster(ctx context.Context, readyEtcdHosts []*hosts.Host, localC
...
+			continue
+		}
+		for _, member := range members {
+			if strings.Contains(member.PeerURLs[0], peerURL) {
+				logrus.Infof("[etcd] member [%s] is already part of the etcd cluster", etcdHost.Address)
+				return true, nil
+			}
+		}
+	}
+	if listErr != nil {
```
alena1108: should we return the contents of listError, so we can get the log for the reason in the caller stack? You can use `errors.Wrapf(err, "Failed to check if host [%s] is already part of the etcd cluster", etcdHost.Address)` construct.

alena1108: there is also a typo `par tof`

services/etcd.go
galal-hussein: yes i guess its the same thing, i will remove this part

PR #373, opened 2018-02-21
Ignore inactive hosts

galal-hussein:
https://github.com/rancher/rancher/issues/11529

### DIFF
--- a/cluster/hosts.go
+++ b/cluster/hosts.go
@@ -9,6 +9,7 @@ import (
 	"github.com/rancher/rke/log"
 	"github.com/rancher/rke/pki"
 	"github.com/rancher/rke/services"
+	"github.com/rancher/types/apis/management.cattle.io/v3"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sync/errgroup"
 )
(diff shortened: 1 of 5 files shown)


hosts/hosts.go
```diff
@@ -169,7 +169,15 @@ func GetToDeleteHosts(currentHosts, configHosts []*Host) []*Host {
 			}
 		}
 		if !found {
-			toDeleteHosts = append(toDeleteHosts, currentHost)
+			inactive := false
+			for _, inactiveHost := range inactiveHosts {
+				if inactiveHost.Address == currentHost.Address {
+					inactive = true
```
alena1108: should break here

hosts/hosts.go
```diff
@@ -169,7 +169,15 @@ func GetToDeleteHosts(currentHosts, configHosts []*Host) []*Host {
 			}
 		}
 		if !found {
```
alena1108: Nit: neater to use `if found {return }` construct

cluster/hosts.go
```diff
@@ -26,21 +27,19 @@ func (c *Cluster) TunnelHosts(ctx context.Context, local bool) error {
...
-		err := c.ControlPlaneHosts[i].TunnelUp(ctx, c.DockerDialerFactory)
-		if err != nil {
-			return fmt.Errorf("Failed to set up SSH tunneling for Control host [%s]: %v", c.ControlPlaneHosts[i].Address, err)
-		}
-	}
-	for i := range c.WorkerHosts {
-		if err := c.WorkerHosts[i].TunnelUp(ctx, c.DockerDialerFactory); err != nil {
-			return fmt.Errorf("Failed to set up SSH tunneling for Worker host [%s]: %v", c.WorkerHosts[i].Address, err)
-		}
+	for _, host := range c.InactiveHosts {
```
alena1108: Make sure that validation for etcd/control/worker hosts happen after invalid hosts are removed from the lists

cluster/hosts.go
```diff
@@ -26,21 +27,19 @@ func (c *Cluster) TunnelHosts(ctx context.Context, local bool) error {
...
+		c.EtcdHosts = removeFromHosts(host, c.EtcdHosts)
```
alena1108: may be we should add the warning about host being skipped from the service node list, here as well.

alena1108:
LGTM

Issue #218, opened 2018-01-11
Support the SSH agent


zommarin:
**RKE version:**
v0.0.9-dev

When running RKE it would be nice if it could use the SSH key credentials from the SSH agent just as SSH itself does. This way we do not have to specify the private key password when running RKE (which SSH agent users in general prefers not to do).

superseb:
Can you check if this is a duplicate of https://github.com/rancher/rke/issues/192?

zommarin:
#192 is related, but not the same in my opinion. This issue is however one possible solution for issue #192. Another possible solution for #192 would be to cache the password and not ask for it several times.

zommarin:
I would prefer that RKE never (not even once) asks for the password, instead it gets the credentials from the SSH Agent, if available. That way I would feel more secure as less secrets are exposed to RKE.

ibuildthecloud:
I'd love to see support for this too.  We use golang.org/x/crypto/ssh for SSH.  So the question is whether that library support SSH agent.

ibuildthecloud:
Sure seems like there is some support, but not really sure how to use it.

https://godoc.org/golang.org/x/crypto/ssh/agent

rickard-von-essen:
[Here is an example](https://github.com/hashicorp/packer/blob/master/builder/amazon/common/ssh.go#L95-L111) of how we use the SSH agent in Packer.

rickard-von-essen:
I've done a ugly fix for this that works for my personal use, but I'd like to do a proper PR but I'l like some comments before completing it.

I would like to remove password prompting and instead fail with a proper error message if the private key is encrypted.
And add a new configuration option `ssh_agent` (bool) which defaults to false, if true enable using the SSH agent.

moelsayed:
Tested and confirmed working with latest master.

rickard-von-essen:
Fixed in #387

zommarin:
Thanks Guys! Great job!

rossedman:
@moelsayed @superseb does this mean that if we have a custom ssh_config with bastion proxies it will work?

PR #387, opened 2018-02-26
Use SSH Agent

cloudnautique:
This PR makes it so that:

1. The SSH Agent specified by SSH_AUTH_SOCK is always used. 
2. If run from an environment without the SSH Agent running it will fall back to using a non-passphrase protected private key.

- This does remove private keys with passphrases and no agent support. 

Additionally, the code was re-factored to de-duplicate some code paths with the Dial methods. 
The host object was also removed from the dialer type.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.idea
 /.dapper
 /bin
 /dist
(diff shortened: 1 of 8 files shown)


cloudnautique:
Addresses #218

PR #393, opened 2018-03-02
Check if labels maps is nil

galal-hussein:


k8s/node.go
```diff
@@ -139,17 +139,22 @@ func SyncTaints(k8sClient *kubernetes.Clientset, nodeName string, toAddTaints, t
...
-		oldLabels[k] = v
-	}
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			logrus.Debugf("[hosts] Can't find node by name [%s]", nodeName)
 			return nil
 		}
 		return err
 	}
+	oldLabels := make(map[string]string)
```
alena1108: use oldLabels := map[string]string{} instead

PR #397, opened 2018-03-03
Update kube-api runtime config

moelsayed:


cluster/plan.go
```diff
@@ -86,8 +86,7 @@ func (c *Cluster) BuildKubeAPIProcess() v3.Process {
 		"--kubelet-preferred-address-types=InternalIP,ExternalIP,Hostname",
 		"--service-cluster-ip-range=" + c.Services.KubeAPI.ServiceClusterIPRange,
 		"--admission-control=ServiceAccount,NamespaceLifecycle,LimitRanger,PersistentVolumeLabel,DefaultStorageClass,ResourceQuota,DefaultTolerationSeconds",
-		"--runtime-config=batch/v2alpha1",
-		"--runtime-config=authentication.k8s.io/v1beta1=true",
+		"--runtime-config=batch/v1beta1",
```
alena1108: @moelsayed do we have to explicitly enable it because of "beta", or because of "batch"? Could you make sure that you can jobs are enabled (batch/v1) in addition to cronjobs (batch/v1beta1)

ibuildthecloud: batch/v1beta1 should be enabled by default without the need for arguments

galal-hussein:
closed in favor of #399

PR #402, opened 2018-03-05
Fix double admission-controller parameter when enabling pod security policies

Overv:
Right now enabling pod security policies will cause the `--admission-control=PodSecurityPolicy` flag to be added to the kube API server, but it is overridden by the existing `--admission-control` parameter. This commit fixes it by replacing the whole parameter.


galal-hussein:
Closing this PR as PR #414 solves the override admission controller for all services, thanks a lot

HighwayofLife:
@Overv Thanks for your contribution, I applied your principal to PR #414

PR #403, opened 2018-03-06
Leverage global SSHAgentAuth setting

cloudnautique:
This addresses users issues in being unable to use RKE command line
using SSH_AUTH_SOCK. On OSX the socket env var is set, but nothing
is listening. Also, Linux users have reported issues. To address this
the default mode is to not use SSH Agent Auth. A user must set it
explicitly in either the config file or on the CLI. The only way
to use a passphrase protected key file is with a properly configured
SSH Agent and using SSH Agent Auth.


cloudnautique:
Dependent on rancher/types#283

cmd/common.go
```diff
@@ -6,9 +6,17 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/rancher/types/apis/management.cattle.io/v3"
 	"github.com/urfave/cli"
 )
 
+var sshCliOptions = []cli.Flag{
```
galal-hussein: is the cli flag needed since you already add the option to cluster.yml

cloudnautique: It is a convenience thing. If I download the yaml, I can just run `rke up --ssh-agent-auth` instead of having to edit the file. If the work flow for RKE is simliar to other devops tools different team members might have different setups. You and I apparently would have to edit the file and each time we ran a common yaml as we found out with this issue.

galal-hussein:
#398

PR #401, opened 2018-03-05
Add rprivate parameters to kubelet mounts. Fixed #285 #263.

Jason-ZW:
Such as Docker 1.13, bind mounts propagation not default to `rprivate` which cause problems when kubelet unmount cgroup.
For rke, this will result in the failure of the cluster to be initialized with the docker specified version like 1.13.1.

Issue #404, opened 2018-03-06
System break when restart node


Jason-ZW:
**RKE version:**
master branch - last release
**Docker version: (`docker version`,`docker info` preferred)**
1.13.1
```
Containers: 1
 Running: 0
 Paused: 0
 Stopped: 1
Images: 1
Server Version: 17.03.2-ce
Storage Driver: aufs
 Root Dir: /var/lib/docker/aufs
 Backing Filesystem: extfs
 Dirs: 10
 Dirperm1 Supported: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins: 
 Volume: local
 Network: bridge host macvlan null overlay
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: 4ab9917febca54791c5f071a9d1f404867857fcc
runc version: 54296cf40ad8143b62dbcaa1d90e520a2136ddfe
init version: 949e6fa
Security Options:
 apparmor
 seccomp
  Profile: default
Kernel Version: 4.4.0-1052-aws
Operating System: Ubuntu 16.04.3 LTS
OSType: linux
Architecture: x86_64
CPUs: 1
Total Memory: 990.8 MiB
Name: ip-172-31-35-214
ID: BHCW:MLYL:7QNW:TMOF:IUZC:54IF:YHJI:LDBJ:IU5N:SXPU:OXGF:GMJY
Docker Root Dir: /var/lib/docker
Debug Mode (client): false
Debug Mode (server): false
Registry: https://index.docker.io/v1/
Experimental: false
Insecure Registries:
 127.0.0.0/8
Live Restore Enabled: false
```
**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)**
```
NAME="Ubuntu"
VERSION="16.04.3 LTS (Xenial Xerus)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 16.04.3 LTS"
VERSION_ID="16.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
VERSION_CODENAME=xenial
UBUNTU_CODENAME=xenial
```
```
4.4.0-1052-aws
```

**Type/provider of hosts: (VirtualBox/Bare-metal/AWS/GCE/DO)**
AWS
**cluster.yml file:**
```
rke up --local
```
[...]

PR #405, opened 2018-03-06
Add rprivate to fix system break after reboot

Jason-ZW:
System break when restart node, update for PR #401


Jason-ZW:
relation: #404

Jason-ZW:
@moelsayed 
Could you please take a look at this PR? This pr is update for pr #401 and relate to Issue #404.
I found a bug: default bind mounts propagation cause system break after reboot system when rke using docker 1.13.1.
This PR LGTM to fixed the issue.
Thanks a lot.

Issue #190, opened 2018-01-02
Add support for user addons as URLs to yaml manifests.


moelsayed:
We need to allow uses to define addons as URLs in addition to inline manifests, for example:
```
addons_include:
- https://example.com/my.yaml
- https://example.com/another.yaml
```

rossedman:
@moelsayed Wanted to open another issue for local files as well. Could this be a similar feature?

```
addons_include:
- ./manifests/something.yaml
- ./manifests/another.yaml
```

dahendel:
I am interested in working on this if work has not been done already but not sure the best way to proceed as we would need a new field in [rke_types](https://github.com/rancher/types/blob/master/apis/management.cattle.io/v3/rke_types.go#L13)  which appears to be generated.

dahendel:
Here is my code to handle [urls and paths](https://github.com/dahendel/rke/blob/addons_include/cluster/addons.go#L57). I added AddonsInclude to rancher/types that will be another pull request. Built it locally by adding AddonsInclude []string to the RancherKubernetesEngineConfig in the vendor directory

moelsayed:
@dahendel Thank you for the PR, can you also send a rancher/rke PR ? We will review and merge as soon as possible.

rossedman:
@dahendel should json so be included in the types? Do we see a situation where someone is piping in machine generated manifests? Awesome start!

dahendel:
I will get json added in the rke types in the morning.

moelsayed:
Works as expected with a minor issue if the any of the included manifests after the first one doesn't start with `---`. 

#440 should fix this.

PR #421, opened 2018-03-19
Add support for URLS and File Paths for addons

dahendel:
PR for #190 goes along with [Types PR](https://github.com/rancher/types/pull/289).


cluster/addons.go
```diff
@@ -39,14 +45,113 @@ func (c *Cluster) deployUserAddOns(ctx context.Context) error {
 	log.Infof(ctx, "[addons] Setting up user addons..")
 	if c.Addons == "" {
 		log.Infof(ctx, "[addons] No user addons configured..")
+		return c.deployAddonsInclude(ctx)
```
galal-hussein: the log message here might be confusing since you are still running user addon includes

cluster/addons.go
```diff
@@ -39,14 +45,113 @@ func (c *Cluster) deployUserAddOns(ctx context.Context) error {
...
+	logrus.Debugf("[addons] Compiled addons yaml: %s", userAddons)
+	if err := c.doAddonDeploy(ctx, userAddons, UserAddonsIncludeResourceName); err != nil {
 		return err
 	}
-	log.Infof(ctx, "[addons] User addon deployed successfully..")
+
 	return nil
+}
+
+func (c *Cluster) isFilePath(addonPath string) bool {
```
galal-hussein: i am not sure if this function should be a method struct since its not related to the cluster object at all

cluster/addons.go
```diff
@@ -39,14 +45,113 @@ func (c *Cluster) deployUserAddOns(ctx context.Context) error {
...
+	if _, err := os.Stat(addonPath); os.IsNotExist(err) {
+		return false
+	} else {
+		return true
+	}
+}
+
+func (c *Cluster) getAddonFromUrl(yamlURL string) (string, error) {
```
galal-hussein: same as above ^

dahendel:
@galal-hussein  Done.

cluster/addons.go
```diff
@@ -38,15 +44,113 @@ func (c *Cluster) deployK8sAddOns(ctx context.Context) error {
...
+}
+
+// deployAddonsInclude takes any passed addons from the addons key and appends yaml contents from urls/paths
+// then deploys all addons at once. Not doing this in this manner will fail to deploy all of the plugins.
+// Only the addons from the addons: key would be deployed.
+func (c *Cluster) deployAddonsInclude(ctx context.Context) error {
+	var userAddons = ""
+	log.Infof(ctx, "[addons] Checking for included user addons")
+
+	yamlDelimiter := regexp.MustCompile("(\n---$|\n---\n$)")
```
galal-hussein: is it possible to skip the regex check and check using the yaml library in golang

cluster/addons.go
```diff
@@ -38,15 +44,113 @@ func (c *Cluster) deployK8sAddOns(ctx context.Context) error {
...
+					return err
+				}
+
+				log.Infof(ctx, "[addons] Adding addon from %s", addon)
+
+				// Debug statements to troubleshoot addons not deploying correctly
+				logrus.Debugf("FilePath Yaml: %s", yamlContents)
+
+				// Some manifests contain --- at the end of the file, if it does we do not need to append it
+				if yamlDelimiter.MatchString(string(yamlContents)){
```
galal-hussein: same as above

cluster/addons.go
```diff
@@ -38,15 +44,113 @@ func (c *Cluster) deployK8sAddOns(ctx context.Context) error {
...
-		log.Infof(ctx, "[addons] No user addons configured..")
-		return nil
+		return c.deployAddonsInclude(ctx)
 	}
 
 	if err := c.doAddonDeploy(ctx, c.Addons, UserAddonResourceName); err != nil {
 		return err
 	}
 	log.Infof(ctx, "[addons] User addon deployed successfully..")
+	return c.deployAddonsInclude(ctx)
```
galal-hussein: instead of returning user include here, can you add it with `if err :=` and if err is nil then we can print the success message

cluster/addons.go
dahendel: May I ask what the reasoning for this is? I don't understand the approach you are wanting to take, and I don't see anything in the go yaml lib to do something like this.

galal-hussein:
Can you also add example for this in the `cluster.yml` and in the documentation

cluster/addons.go
```diff
@@ -38,15 +44,113 @@ func (c *Cluster) deployK8sAddOns(ctx context.Context) error {
...
+// deployAddonsInclude takes any passed addons from the addons key and appends yaml contents from urls/paths
+// then deploys all addons at once. Not doing this in this manner will fail to deploy all of the plugins.
+// Only the addons from the addons: key would be deployed.
+func (c *Cluster) deployAddonsInclude(ctx context.Context) error {
+	var userAddons = ""
+	log.Infof(ctx, "[addons] Checking for included user addons")
+
+	yamlDelimiter := regexp.MustCompile("(\n---$|\n---\n$)")
+	// If there are contents in the array continue
+	if len(c.AddonsInclude) != 0 {
```
moelsayed: can we revers this ? as in:
```
if len(c.AddonsInclude) == 0 {
    log.Infof(ctx, "[addons] No included addon paths or urls..")
    return nil
}
```
Makes the code a bit simpler.

dahendel: Sure thing

galal-hussein:
@dahendel The types PR has been merged, can you update the vendors and squash the commits to two commits, one for vendor update and one for code change.
Also i am not sure i understand the use of regex in the code to identify the yaml.

dahendel:
@galal-hussein it isn't Identifying the yaml, it is just ensuring there is the `---` seperator between the different files. Because I am just concatenating the strings together as I read the manifests, some of the manifests do not have the `---` to distinguish between the two. 
Lets say we have the following to manifests:
```
apiVersion: rook.io/v1alpha1
kind: Pool
metadata:
  name: replicapool
  namespace: rook
spec:
  replicated:
    size: 3
```

```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
   name: rook-block
provisioner: rook.io/block
parameters:
  pool: replicapool
```

When I concatenate these they would look like:
```
apiVersion: rook.io/v1alpha1
kind: Pool
metadata:
  name: replicapool
  namespace: rook
spec:
  replicated:
    size: 3
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
   name: rook-block
provisioner: rook.io/block
parameters:
  pool: replicapool
```
Which would not deploy correctly, the regex would catch that the `---` was missing and add it
```
apiVersion: rook.io/v1alpha1
kind: Pool
metadata:
  name: replicapool
  namespace: rook
spec:
  replicated:
    size: 3
---
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
   name: rook-block
provisioner: rook.io/block
parameters:
  pool: replicapool
```
I used regex so I didn't have to marshal and unmarshal all the manifests then add them all together. I tried a couple ways using strings.HasSuffix("---") and a few variations of that but it didn't work well at all. 
I am definitely open to feedback on the regex piece.

I will flatten get these flattened today.

deniseschannon:
@dahendel can you rebase?

galal-hussein:
@dahendel i understand the point but i don't think we should patch the user's yaml, i think the right way is to validate the yaml using the yaml go library and if its not correct then display an error to the user

dahendel:
@galal-hussein sounds good. I will work on that. 

I do have one other question, should we keep all addons in one big config map or deploy them separately?
I can see reasons for both, but personally I like having them stored in separate config maps.

PR #424, opened 2018-03-21
add node ip option

StrongMonkey:


StrongMonkey:
there is another option which might be safer: add a field call overrideIPAddress and if this field is set, set node-ip. I don't know if setting node-ip to host.address by default will cause any problem.

StrongMonkey:
https://github.com/StrongMonkey/rke.git

galal-hussein:
@StrongMonkey can you resolve the conflicts in plan.go, thanks!

StrongMonkey:
@galal-hussein done

alena1108:
@StrongMonkey `there is another option which might be safer: add a field call overrideIPAddress` - after analyzing rke/k8s stuff, looks like that would be the best option. `address` and `internalAddress` from nodeConfig were originally meant for rke configuration purpose only. `address` - for rke to communicate with k8s components. `internalAddress` - in *some* cases only, it is passed to k8s component so it can be used for intern communication (worker to api server ,api server to etcd). The problem is: `internalAddress` gets defaulted to `address` if not specified. And we can't pass `intenalAddress` flag directly as a node-ip on kubelet start as it can hold external ip value.

As introducing a new flag on api would require changing components like kontainer-engine, custom host registration, I'd suggest we hold on making this fix.

deniseschannon:
Fixed in https://github.com/rancher/rke/pull/444

Issue #412, opened 2018-03-13
RKE Kuberenetes component logs


cloudnautique:
The container logs for the RKE Kubernetes components are difficult to capture without getting duplicate data. In a typical k8s logging setup `/var/log/containers` is being monitored by fluentd for Pod logs. These links are currently, for Docker envs, symlinks back into to `/var/lib/docker` which is where the RKE Kubernetes components log to. 

It would be ideal, if we could follow a similar model and create symlinks in `/var/lib/rancher/rke/logs` back to `/var/lib/docker/containers`. We could also launch with docker args for log_opts and log_driver.

This would allow users to capture the logs and also allow Rancher to send system logs to a unique index separate from the K8s application logs. Since the log types likely have different audiences.

aiwantaozi:
Hi @cloudnautique , RKE component are not in the cluster that rke deploy. So the log for this component are lack of info like namespace and so on. 

Kubenetes create the symlink /var/log/container -> /var/log/pods/ -> /var/lib/docker/containers. In 2.0 which part create the symlink? 

How to know whether the cluster is deploy by RKE?

moelsayed:
The docker log drivers don't allow selecting log files location. So, AFIAK that's a dead end. There is a fluentd driver but it would be really complicated to implement. 

I am trying to set `--log-dir` on the components to write logs to a specific location, but it's not working. Still looking into it.

So far, I think our best bet is use fluentd to filter logs based on tags maybe to avoid duplication and route it to it's different index. The place to look is `/var/lib/docker/containers/` and that's already mounted in the fluentd pod. So it's basically a matter of generating the configuration. 

@cloudnautique @aiwantaozi what do you think ?

cloudnautique:
@moelsayed, I was thinking that we would do exactly what K8s does and create symlinks. To implement it, just have another container that creates symlinks and points to the logs in `/var/lib/docker`. If you think we can watch all of /var/lib/docker/containers and just filter on labels... I suppose that would work. But I would hate to be tailing the files twice as it is already kind of expensive.

aiwantaozi:
I agree with Bill, k8s create the symlink for the container it deployed and put the info(like namespace) to the symlink path, it would be better if rke or another component create the symlink with the info we need. If we watch the path /var/lib/docker/containers, we have to parse all the log, talk to docker, filter the rke related log.

moelsayed:
[...]

PR #425, opened 2018-03-21
Log RKE components to a specific location

moelsayed:
https://github.com/rancher/rke/issues/412

### DIFF
--- a/cluster/logs.go
+++ b/cluster/logs.go
@@ -1,0 +1,58 @@
+package cluster
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/docker/docker/api/types/container"
+	"github.com/rancher/rke/docker"
+	"github.com/rancher/rke/hosts"
+	"github.com/rancher/rke/services"
+	"github.com/rancher/types/apis/management.cattle.io/v3"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sync/errgroup"
+)
+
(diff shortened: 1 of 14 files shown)


services/services.go
```diff
@@ -88,3 +94,41 @@ func GetHealthCheckURL(useTLS bool, port int) string {
...
+		Cmd: []string{
+			"sh",
+			"-c",
+			fmt.Sprintf("mkdir -p %s ; ln -s %s %s", RKELogsPath, containerLogPath, containerLogLink),
+		},
+	}
+	hostCfg := &container.HostConfig{
+		Binds: []string{
+			"/var/lib:/var/lib",
+			fmt.Sprintf("%s:%s", path.Dir(containerLogPath), path.Dir(containerLogPath)),
```
galal-hussein: i don't think this is needed since you already bind mount `/var/lib`

cluster/logs.go
```diff
@@ -0,0 +1,59 @@
...
+		Cmd: []string{
+			"sh",
+			"-c",
+			fmt.Sprintf("find %s -type l ! -exec test -e {} \\; -print -delete", services.RKELogsPath),
+		},
+	}
+	hostCfg := &container.HostConfig{
+		Binds: []string{
+			"/var/lib:/var/lib",
+			fmt.Sprintf("%s:%s", services.RKELogsPath, services.RKELogsPath),
```
galal-hussein: same here

PR #414, opened 2018-03-14
Support component command argument overrides

HighwayofLife:
This change allows component attribute `extra_args` to override existing arg=values if they exist.

This PR solves #275

PR #431, opened 2018-03-23
Add AWS cloud provider support

galal-hussein:
https://github.com/rancher/rke/issues/434 
will update vendors when types [PR](https://github.com/rancher/types/pull/311) is merged

### DIFF
--- a/cluster/plan.go
+++ b/cluster/plan.go
@@ -84,7 +84,7 @@ func (c *Cluster) BuildKubeAPIProcess() v3.Process {
 		"bind-address":                    "0.0.0.0",
 		"insecure-port":                   "0",
 		"secure-port":                     "6443",
-		"cloud-provider":                  "",
+		"cloud-provider":                  c.CloudProvider.Name,
 		"allow_privileged":                "true",
 		"kubelet-preferred-address-types": "InternalIP,ExternalIP,Hostname",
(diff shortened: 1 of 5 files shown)


cluster/plan.go
```diff
@@ -81,7 +81,7 @@ func (c *Cluster) BuildKubeAPIProcess() v3.Process {
 		"--bind-address=0.0.0.0",
 		"--insecure-port=0",
 		"--secure-port=6443",
-		"--cloud-provider=",
```
HighwayofLife: With PR #414, it would allow the following configuration in the cluster.yml file:
```
  extra_args:
    cloud-provider: aws
    configure-cloud-routes: false
```

galal-hussein:
@HighwayofLife thanks! i rebased it to work with the changes you made for plan.go

k8s/node.go
```diff
@@ -22,7 +27,16 @@ func GetNodeList(k8sClient *kubernetes.Clientset) (*v1.NodeList, error) {
 }
 
 func GetNode(k8sClient *kubernetes.Clientset, nodeName string) (*v1.Node, error) {
-	return k8sClient.CoreV1().Nodes().Get(nodeName, metav1.GetOptions{})
```
oldthreefeng, line 25: use Get(nodeName, metav1.GetOptions{}) method is More efficient， why use for loop to get this?  @galal-hussein

PR #442, opened 2018-03-26
Support for node config from docker-machine

dahendel:
This would enable users to import docker-machine node configs during the `rke config` by adding the `--from-machine` or `-m`. In case a user has changed the docker-machine store-path (/home/$USER/.docker/machine/machines) it also prompts for the location. The next step is to provide a comma separated list of the machines to use for the nodes. 
```
# rke config -m
[+] Cluster Level SSH Private Key Path [~/.ssh/id_rsa]: 
[+] Docker Machine storage path [/home/dhendel/.docker/machine/machines]: 
[+] Which nodes would you like to use [rk8s01,rk8s02,rk8s03,rk8s04]: rk8s01,rk8s02,rk8s03,rk8s04
```

It reads the config.json for the node data using the following struct:
```go
type NodeMachineConfig struct {
	ConfigVersion int `json:"ConfigVersion,omitempty"`
	Driver        struct {
		IPAddress  string `json:"IPAddress,omitempty"`
		SSHKeyPair string `json:"SSHKeyPair,omitempty"`
		SSHKeyPath string `json:"SSHKeyPath,omitempty"`
		SSHPort    int    `json:"SSHPort,omitempty"`
		SSHUser    string `json:"SSHUser,omitempty"`
	} `json:"Driver,omitempty"`
	HostOptions struct {
		EngineOptions struct {
			Labels []string `json:"Labels,omitempty"`
		} `json:"EngineOptions,omitempty"`
	}
	Name string `json:"Name,omitempty"`
}
```
For now I have included the type in the config.go. Wasn't sure if there was anything in ranhcer/types that I could use to read the config data. 

This option also assumes that labels were passed to the engine and reads them to configure the node:
   - worker=[true|false]
   - controlplane=[true|false]
   - etcd=[true|false]

If no labels are passed it assumes the node will have all roles, this of course could be changed once the config is generated.

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -5,9 +5,12 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"os/user"
 	"strconv"
 	"strings"
 
+	"encoding/json"
+	"github.com/pkg/errors"
 	"github.com/rancher/rke/cluster"
 	"github.com/rancher/rke/pki"
 	"github.com/rancher/rke/services"
@@ -40,6 +43,10 @@ func ConfigCommand() cli.Command {
 			cli.BoolFlag{
 				Name:  "print,p",
 				Usage: "Print configuration",
+			},
+			cli.BoolFlag{
+				Name:  "from-machine,m",
+				Usage: "Get node configuration from docker-machine",
 			},
 		},
 	}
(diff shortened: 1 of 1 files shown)


ibuildthecloud:
I like this idea.

moelsayed:
@dahendel Can you please add some documentation for this in the README ? It should clarify that they user used to connect needs to be added to the docker group on the machine.

dahendel:
@moelsayed will do.

dahendel:
Just let me know which path you want me to go. @moelsayed @superseb

moelsayed:
@dahendel I think @superseb has a good point. We might want to add more sources in the future. Can you please make it so the option takes an argument specifying the node source ?

dahendel:
@moelsayed will do. I will move the code into a separate file as well.

dahendel:
I will clean up the commits, but I have made a new package called providers. I can refactor it to a better name if you want it called something else. It provides a very simple NodeProvider interface that can be used to add different providers.

dahendel:
@moelsayed @superseb I am going to close this PR and open another one since this is going to support other node configuration providers

rbucker:
I think this change has been sitting around long enough. The longer you wait on a config feature that is transient anyway.... this committers hard work will be lost.

PR #443, opened 2018-03-26
Move pull message to info

ibuildthecloud:


ibuildthecloud:
This just helps the use know what is going on.  Since pulls take the most amount of time.

galal-hussein:
we moved this message to debug because it created too much log lines which made very hard to debug bugs

Issue #363, opened 2018-02-19
Ability to override node-ip on kubelet


superseb:
Systems with multiple interfaces get the wrong Node IP autodetected when that IP is not the IP to the default gateway.

kubelet option:
```     
 --node-ip string
IP address of the node. If set, kubelet will use this IP address for the node
```

Example here: https://github.com/rancher/rancher/issues/11365

moelsayed:
Tested using latest master:

```
   - hostname_override: worker4
     address: x.x.x.x
     internal_address: 172.31.19.171
     user: ubuntu
     role: [worker]
     35.163.129.209
```
Results:
```
$ kubectl describe node worker4
Name:               worker4
Roles:              worker
Labels:             beta.kubernetes.io/arch=amd64
                    beta.kubernetes.io/os=linux
                    kubernetes.io/hostname=worker4
                    node-role.kubernetes.io/worker=true
Annotations:        alpha.kubernetes.io/provided-node-ip=172.31.19.171
                    flannel.alpha.coreos.com/backend-data={"VtepMAC":"12:96:d0:3d:35:47"}
                    flannel.alpha.coreos.com/backend-type=vxlan
                    flannel.alpha.coreos.com/kube-subnet-manager=true
                    flannel.alpha.coreos.com/public-ip=172.31.26.110
                    node.alpha.kubernetes.io/ttl=0
                    rke.io/external-ip=x.x.x.x
                    rke.io/internal-ip=172.31.19.171
                    volumes.kubernetes.io/controller-managed-attach-detach=true
Taints:             <none>
CreationTimestamp:  Thu, 29 Mar 2018 15:46:06 +0200
Conditions:
  Type             Status  LastHeartbeatTime                 LastTransitionTime                Reason                       Message
  ----             ------  -----------------                 ------------------                ------                       -------
  OutOfDisk        False   Thu, 29 Mar 2018 17:22:43 +0200   Thu, 29 Mar 2018 15:46:06 +0200   KubeletHasSufficientDisk     kubelet has sufficient disk space available
  MemoryPressure   False   Thu, 29 Mar 2018 17:22:43 +0200   Thu, 29 Mar 2018 15:46:06 +0200   KubeletHasSufficientMemory   kubelet has sufficient memory available
  DiskPressure     False   Thu, 29 Mar 2018 17:22:43 +0200   Thu, 29 Mar 2018 15:46:06 +0200   KubeletHasNoDiskPressure     kubelet has no disk pressure
[...]

PR #444, opened 2018-03-26
Set node-ip and addresses annotations

galal-hussein:


k8s/node.go
```diff
@@ -244,3 +247,30 @@ func toTaint(taintStr string) v1.Taint {
 		Effect: effect,
 	}
 }
+
+func SetAddressesAnnotations(k8sClient *kubernetes.Clientset, nodeName, internalAddress, externalAddress string) error {
```
alena1108: can we run this update along with other node update? where we set labels

deniseschannon:
rancher/rancher#11365
#363

Issue #447, opened 2018-03-27
Lint fails on cluster/addons - redundant if


HighwayofLife:
Attempting to build binaries results in the following build error from golint:

```
Running: golint
cluster/addons.go:115:2: redundant if ...; err != nil check, just return error instead.
FATA[0131] exit status 1 
```

Relevant commit ID: 7875c0996645b1b0f3d5a4d55960068b238e8779

dahendel:
That should be fixed in [master](https://github.com/rancher/rke/blob/master/cluster/addons.go#L110) already

HighwayofLife:
@dahendel The problem exists in master.
I built off of master ( 19eb46332ef0f960a78a9fec42924b992284f3f9 )

PR #448, opened 2018-03-27
Fix for #447

HighwayofLife:


deniseschannon:
https://github.com/rancher/rke/issues/447

PR #449, opened 2018-03-28
Add azure cloud provider

galal-hussein:
depends on https://github.com/rancher/rke-service-sidekick/pull/2 and https://github.com/rancher/types/pull/320


README.md
```diff
@@ -295,6 +295,50 @@ services:
...
+AWS cloud provider has to be enabled on ec2 instances with the right IAM role.
+
+### Azure Cloud provider
+
+Azure cloud provider can be enabled by passing `azure` as the cloud provider name and set of options to the configuration file:
+```
+cloud_provider:
+  name: azure
+  cloud_config:
+    aadClientId: xxxxxxxxxxxx
```
alena1108: why camelcase in yml?

cluster/cloud-provider.go
```diff
@@ -0,0 +1,97 @@
...
+			continue
+		}
+		tmpFloat, err := strconv.ParseFloat(value, 64)
+		if err == nil {
+			tmpMap[key] = tmpFloat
+			continue
+		}
+		tmpMap[key] = value
+	}
+	if len(cloudProvider.CloudConfig) == 0 {
```
alena1108: we can make this check before iterating over map; this way it would be more clear

README.md
galal-hussein: they used the same naming convention in azure cloud provider implementation in kubernetes https://github.com/kubernetes/kubernetes/blob/v1.9.5/pkg/cloudprovider/providers/azure/azure.go#L101

cluster/hosts.go
```diff
@@ -120,6 +120,12 @@ func (c *Cluster) SetUpHosts(ctx context.Context) error {
 			return err
 		}
 		log.Infof(ctx, "[certificates] Successfully deployed kubernetes certificates to Cluster nodes")
+		if len(c.CloudProvider.Name) > 0 {
```
alena1108: why len? :) why can't just check if `c.CloudProvider.Name != ""`

PR #452, opened 2018-03-29
Network and cloud provider types for rke

galal-hussein:


cluster/hosts.go
```diff
@@ -121,7 +121,7 @@ func (c *Cluster) SetUpHosts(ctx context.Context) error {
 		}
 		log.Infof(ctx, "[certificates] Successfully deployed kubernetes certificates to Cluster nodes")
 		if c.CloudProvider.Name != "" {
-			if err := deployCloudProviderConfig(ctx, hosts, c.CloudProvider, c.SystemImages.Alpine, c.PrivateRegistriesMap); err != nil {
+			if err := deployCloudProviderConfig(ctx, hosts, c.CloudProvider, c.SystemImages.Alpine, c.PrivateRegistriesMap, c.AzureCloudProvider); err != nil {
```
alena1108: we should build cloud provider similar to host. Basically do it in the beginning.

PR #466, opened 2018-04-01
Add external Cloud Provider support

yamamoto-febc:
Support the ability to use external cloud-provider.

To use the external cloud provider, we need to do as follows.

- `kube-apiserver` and `kube-controller-manager` MUST NOT specify the --cloud-provider flag
- `kubelet` must run with --cloud-provider=external

> ref: [Kubernetes Document: Running cloud-controller-manager](https://kubernetes.io/docs/tasks/administer-cluster/running-cloud-controller/#running-cloud-controller-manager)

so this PR will adds processing for these.

Also, kubelets specifying --cloud-provider=external will add a taint `node.cloudprovider.kubernetes.io/uninitialized` with an effect NoSchedule during initialization.

`kube-dns` and network plugins can not be started due to this effect, so this PR adds `tolerations` to each templates.
Only the canal template already had `tolerations`, so I did not modify it.

> ref: https://github.com/rancher/rke/blob/master/templates/canal.go#L210-L214)

### DIFF
--- a/cluster/plan.go
+++ b/cluster/plan.go
@@ -91,7 +91,6 @@ func (c *Cluster) BuildKubeAPIProcess() v3.Process {
 		"bind-address":                    "0.0.0.0",
 		"insecure-port":                   "0",
 		"secure-port":                     "6443",
-		"cloud-provider":                  c.CloudProvider.Name,
 		"allow-privileged":                "true",
 		"kubelet-preferred-address-types": "InternalIP,ExternalIP,Hostname",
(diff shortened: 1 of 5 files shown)


yamamoto-febc:
I found one missing fix, so I pushed again.

galal-hussein:
@yamamoto-febc Thanks for the contribution, however this PR doesn't actually add any new cloud provider, the documentation you provided is for running **cloud-controller-manager** which is actually different component that is responsible for running the cloud provider, we don't actually support running cloud-controller-manager at the moment as it will require different service to run

PR #472, opened 2018-04-02
Add get node for delete and cordon

galal-hussein:


k8s/node.go
```diff
@@ -22,7 +22,11 @@ const (
 )
 
 func DeleteNode(k8sClient *kubernetes.Clientset, nodeName string) error {
```
alena1108: can we at least add a comment stating that the nodeName passed to the call, is really something else

PR #474, opened 2018-04-03
Add bind mount prompts for kubelet container

dahendel:
This PR prompts the user during configuration to do 2 things:
1. Prompt for the FlexVolume Driver mount and prompt for rke or gke deployment to determine appropriate bind mount path (based on the longhorn deployment)
2. Prompt for additional bind mounts for the kubelet service. This takes a comma separated list of bind mounts and writes them to the cluster.yml

```
[+] Enable the FlexVolume driver mount [no]: yes
[+] Deploying to GKE or RKE environment [rke]: rke
[+] Add additional bind mounts for kubelet service [no]: yes
[+] Additional bind mounts (separated by ",") [/dev:/dev]: /dev:/dev,/home:/tmp/home,testme:/test2
```

```
kubelet:
    image: rancher/k8s:v1.8.10-rancher1-1
    extra_args: {}
    extra_binds:
    - /var/lib/kubelet/volumeplugins:/var/lib/kubelet/volumeplugins
    - /dev:/dev
    - /home:/tmp/home
    - testme:/test2
```

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -303,6 +303,47 @@ func getServiceConfig(reader *bufio.Reader) (*v3.RKEConfigServices, error) {
 		return nil, err
 	}
 	servicesConfig.Kubelet.InfraContainerImage = infraPodImage
+
+	// Add the flexvolume mount to kublet ExtraBinds
+	flexVolMount, err := getConfig(reader, "Enable the FlexVolume driver mount", "no")
+	if err != nil {
+		return nil, err
+	}
+
+	if strings.ContainsAny(flexVolMount, "Yes y Y yes") {
+		gkeBind := "/home/kubernetes/flexvolume:/home/kubernetes/flexvolume"
+		rkeBind := "/var/lib/kubelet/volumeplugins:/var/lib/kubelet/volumeplugins"
+
+		getCE, err := getConfig(reader, "Deploying to GKE or RKE environment", "rke")
+
+		if err != nil {
+			return nil, err
+		}
+
+		if strings.ToLower(getCE) == "gke" {
+			servicesConfig.Kubelet.ExtraBinds = append(servicesConfig.Kubelet.ExtraBinds, gkeBind)
+		} else {
+			servicesConfig.Kubelet.ExtraBinds = append(servicesConfig.Kubelet.ExtraBinds, rkeBind)
+		}
+
+	}
+
+	// Add additional ExtraBinds
+	additionalBinds, err := getConfig(reader, "Add additional bind mounts for kubelet service", "no")
+	if err != nil {
+		return nil, err
+	}
+
+	if strings.ContainsAny(additionalBinds, "Yes y Y yes") {
+		bindMounts, err := getConfig(reader, "Additional bind mounts (separated by \",\")", "/dev:/dev")
+
+		if err != nil {
+			return nil, err
+		}
+
+		servicesConfig.Kubelet.ExtraBinds = append(servicesConfig.Kubelet.ExtraBinds, strings.Split(bindMounts, ",")...)
+	}
+
 	return &servicesConfig, nil
 }
 


cmd/config.go
```diff
@@ -303,6 +303,47 @@ func getServiceConfig(reader *bufio.Reader) (*v3.RKEConfigServices, error) {
...
+	}
+
+	// Add additional ExtraBinds
+	additionalBinds, err := getConfig(reader, "Add additional bind mounts for kubelet service", "no")
+	if err != nil {
+		return nil, err
+	}
+
+	if strings.ContainsAny(additionalBinds, "Yes y Y yes") {
+		bindMounts, err := getConfig(reader, "Additional bind mounts (separated by \",\")", "/dev:/dev")
```
moelsayed: Can you please remove this default ?

dahendel: Sure can.

cmd/config.go
```diff
@@ -303,6 +303,47 @@ func getServiceConfig(reader *bufio.Reader) (*v3.RKEConfigServices, error) {
...
 	}
 	servicesConfig.Kubelet.InfraContainerImage = infraPodImage
+
+	// Add the flexvolume mount to kublet ExtraBinds
+	flexVolMount, err := getConfig(reader, "Enable the FlexVolume driver mount", "no")
+	if err != nil {
+		return nil, err
+	}
+
+	if strings.ContainsAny(flexVolMount, "Yes y Y yes") {
```
alena1108: @dahendel seems like we can create a constant for `Yes y Y yes` as its use repeated

cmd/config.go
```diff
@@ -303,6 +303,47 @@ func getServiceConfig(reader *bufio.Reader) (*v3.RKEConfigServices, error) {
...
+		gkeBind := "/home/kubernetes/flexvolume:/home/kubernetes/flexvolume"
+		rkeBind := "/var/lib/kubelet/volumeplugins:/var/lib/kubelet/volumeplugins"
```
alena1108: is this parameter something that can be configured on RKE? if yes, can we reference instead of hardcoding here?

alena1108: Found the reference here: https://github.com/rancher/rke/pull/474/files#diff-31f9f805fb3e45c114bdaa74f9447759R315. Perhaps we can create a global variable, and use it in all places

cmd/config.go
```diff
@@ -303,6 +303,47 @@ func getServiceConfig(reader *bufio.Reader) (*v3.RKEConfigServices, error) {
...
+
+		getCE, err := getConfig(reader, "Deploying to GKE or RKE environment", "rke")
```
alena1108: "RKE environment" sounds a bit confusing - as it doesn't reflect the infrastructure provider (aws, baremetal, etc) it gets deployed to - especially when we oppose it to "GKE environment". Instead you can do something like:

a) `Is GKE cloud?` (false by default)
b) `Host type` (Node by default, GKE as an option). I'm trying to avoid use of `cloudProvider` construct as it's already being used for diff purpose.

cmd/config.go
dahendel: I pulled these paths from the longhorn README. Looking back they definitely should not be hard coded. Reading the [Flexvolume Docs](https://github.com/kubernetes/community/blob/master/contributors/devel/flexvolume.md) should I add `--volume-plugin-dir` to the extra args for the kubelet as well when the user wants to enable Flexvolume mount. I know I have seen some issues in slack asking how to set these values with rke as well. Thoughts? @alena1108

cmd/config.go
dahendel: Would `"Is GKE your cloud provider?" [no]` be more appropriate? 

With the changes I mentioned above we could just do away with this.

alena1108: Thats better. Do you happen to know if other cloud providers might need the same logic as well? If no, `is GKE` (bool) should be fine. If yes, "enum" choice is more appropriate

dahendel: I was thinking of letting the user just input the appropriate path. One reason being is that I have seen users in slack wanting/needing to change the flexvolume path for the kubelet.

dahendel:
```go
const 	defaultFlexVolumeBindPath = "/var/lib/kubelet/volumeplugins:/var/lib/kubelet/volumeplugins"
....
     if strings.ContainsAny(flexVolMount, yesAnswer) {
		// Allow the user to set custom flex volume mount points
		FlexVolumeBind, err = getConfig(reader, "Flex volume bind mount path for kubelet service", defaultFlexVolumeBindPath)

		if err != nil {
			return nil, err
		}
		//If a custom volume bind is passed, use it and set the --volume-plugin-dir extra_args for kubelet service
		if FlexVolumeBind != defaultFlexVolumeBindPath {
			volPluginDir := strings.Split(FlexVolumeBind, ":")
			
			servicesConfig.Kubelet.ExtraArgs["--volume-plugin-dir"] = volPluginDir[1]
		}
		
		servicesConfig.Kubelet.ExtraBinds = append(servicesConfig.Kubelet.ExtraBinds, FlexVolumeBind)

	}

```

cmd/config.go
alena1108: @dahendel `should I add --volume-plugin-dir to the extra args for the kubelet as well when the user wants to enable Flexvolume mount` - yes. This parameter as an extra arg would override https://github.com/rancher/rke/blob/9e32c532e8c426dbb8281da8389ca12a9ceb1e0b/cluster/plan.go#L308

alena1108:
@dahendel I feel like it would be more flexible to just implement `Prompt for additional bind mounts for the kubelet service. This takes a comma separated list of bind mounts and writes them to the cluster.yml`, and document the extra binds that *might* be needed depending on the provider. This way it would be more flexible. Thoughts? @galal-hussein @moelsayed

dahendel:
@alena1108 I agree. I will add it to my list and try to get it done soon.

PR #479, opened 2018-04-04
Accept different patches of supported docker versions

galal-hussein:


docker/docker.go
```diff
@@ -26,7 +27,8 @@ const (
 )
 
 var K8sDockerVersions = map[string][]string{
-	"1.8": {"1.12.6", "1.13.1", "17.03.2"},
+	"1.8": {"1.11.x", "1.12.x", "1.13.x", "17.03.x"},
+	"1.9": {"1.11.x", "1.12.x", "1.13.x", "17.03.x"},
 }
```
alena1108: we support k8s 1.10, so might need to add one more var here

docker/docker.go
```diff
@@ -359,3 +367,9 @@ func GetImageRegistryConfig(image string, prsMap map[string]v3.PrivateRegistry)
 	}
 	return "", "", nil
 }
+
+func convertToSemver(version string) (*semver.Version, error) {
+	compVersion := strings.SplitN(version, ".", 3)
+	compVersion[2] = "0"
```
alena1108: Lets add a check for compVersion len. In case someone unaware adds incorrect value to the map

PR #480, opened 2018-04-05
Docker versions and canal default network plugin

superseb:


ibuildthecloud:
LGTM, should be merged with https://github.com/rancher/rke/pull/481

PR #501, opened 2018-04-13
Fix: deploy rancher v2.0 using rke

rawmind0:
This PR contains following changes:
- Added extra_args to nginx-ingress-controller. 
- Added `rancher-minimal-ssl.yml` and `rancher-minimal-passthrough.yml` to deploy rancher v2.0 using rke. 
- Updated `README.md`


templates/nginx-ingress.go
```diff
@@ -203,6 +203,7 @@ spec:
             - --tcp-services-configmap=$(POD_NAMESPACE)/tcp-services
             - --udp-services-configmap=$(POD_NAMESPACE)/udp-services
             - --annotations-prefix=nginx.ingress.kubernetes.io
+            - --enable-ssl-passthrough
```
cloudnautique: Since the default is to disable it, should we just disable the ingress controller in the yaml and deploy it as an addon in the pass through?

rawmind0: Could be, but not sure, due to ssl-passthrough is not mandatory. It just give the option to use annotation `nginx.ingress.kubernetes.io/ssl-passthrough` at ingress definition, `false` by default.

cloudnautique:
@galal-hussein @moelsayed should we add support for extra args to the nginx ingress controller from the cluster.yml? I know some of our other customers would like that as well.

templates/nginx-ingress.go
ibuildthecloud: @rawmind0 I think adding arguments is going to a be common problem with the nginx controller.  Today we have options as below to control nginx options, but it does not cover arguments.  How about we add `ExtraArguments` where you can add in `--enable-ssl-passthrough`

```yaml
ingress:
  provider: nginx
  options:
  - foo: bar
  extra_arguments:
  - --enable-ssl-passthrough
```

rawmind0: @ibuildthecloud , `ExtraArguments` better option.
I'll update this PR once https://github.com/rancher/types/pull/366 is merged.

ibuildthecloud:
@rawmind0 types has been merged

rawmind0:
Thanks @ibuildthecloud 
Updated PR, builded, tested and working.

ibuildthecloud:
@rawmind0 argh... @HighwayofLife has a good point, we should be consistent.  Unfortunately that's another types change....

rawmind0:
Agreed, i was thinking same...
I already submitted another types PR rancher/types#368 . I'll update this PR once it is merged.

HighwayofLife:
The base services `extra_args` format is:

```
  extra_args:
    key: value
```
Whereas this would be:

```
  extra_args:
    - --key
    - --key=value
```

rawmind0:
@HighwayofLife , yes, in that case, `ExtraArgs      []string`
Ex.
```
ingress:
  provider: nginx
  extra_args:
  - --enable-ssl-passthrough
```

ibuildthecloud:
@rawmind0 It really should be the same, so change to a map.

rawmind0:
Updated.... `extra_args` is now a map. Tested and working.

Issue #512, opened 2018-04-16
Start using hyperkube images


moelsayed:

heww:
please using hyperkube images from dockerhub (not from gcr.io) for the default, thanks!

deniseschannon:
@heww They are in our rancher repo in dockerhub.

bmdepesa:
Validated with rke version v0.1.6-rc1. 

Showing `rancher/hyperkube:v1.8.10`.

Issue #509, opened 2018-04-16
Consolidate system images into a single tools image


moelsayed:
We need to consolidate cert-deployer, nginx, sidekick, and alpine into a single image and call functions using docker cmd instead of using entrypoints for each one.

bmdepesa:
Validated against rke version v0.1.6-rc1. 

Showing `rancher/rke-tools:v0.1.1`

PR #524, opened 2018-04-19
WIP Use new rke-tools image

ibuildthecloud:
Instead of having rke-cert-deployer, rke-nginx-proxy, and alpine as supporting images, we just consolidate to one (rke-tools) as ever host is going to end up with all three images regardless.

This is missing the corresponding types change.


ibuildthecloud:
Needs https://github.com/rancher/types/pull/380

deniseschannon:
rancher/rke#509

deniseschannon:
https://github.com/rancher/rke/issues/512

PR #526, opened 2018-04-20
Backup certs

moelsayed:
- Deploy backup certificates on full plane instead of single node
https://github.com/rancher/rancher/issues/12873

- Fix etcd listen address edge case when public address is DNATed
  Fix etcd client endpoint address 
https://github.com/rancher/rke/issues/527

### DIFF
--- a/cluster/certificates.go
+++ b/cluster/certificates.go
@@ -22,19 +22,23 @@ func SetUpAuthentication(ctx context.Context, kubeCluster, currentCluster *Clust
 		if currentCluster != nil {
 			kubeCluster.Certificates = currentCluster.Certificates
 		} else {
-			var backupHost *hosts.Host
+			var backupPlane string
+			var backupHosts []*hosts.Host
 			if len(kubeCluster.Services.Etcd.ExternalURLs) > 0 {
-				backupHost = kubeCluster.ControlPlaneHosts[0]
+				backupPlane = ControlPlane
+				backupHosts = kubeCluster.ControlPlaneHosts
 			} else {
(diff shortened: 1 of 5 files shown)


cluster/certificates.go
```diff
@@ -176,3 +181,28 @@ func saveCertToKubernetes(kubeClient *kubernetes.Clientset, crtName string, crt
...
+		errgrp.Go(func() error {
+			return pki.DeployCertificatesOnHost(ctx, runHost, kubeCluster.Certificates, kubeCluster.SystemImages.CertDownloader, pki.TempCertPath, kubeCluster.PrivateRegistriesMap)
+		})
+	}
+	return errgrp.Wait()
+}
+
+func fetchBackupCertificates(ctx context.Context, backupHosts []*hosts.Host, kubeCluster *Cluster) (map[string]pki.CertificatePKI, error) {
+	for _, host := range backupHosts {
+		certificates, err := pki.FetchCertificatesFromHost(ctx, kubeCluster.EtcdHosts, host, kubeCluster.SystemImages.Alpine, kubeCluster.LocalKubeConfigPath, kubeCluster.PrivateRegistriesMap)
```
ibuildthecloud: Only one host has to succeed, it seems like if the first fails, it will give up and not try the second.  Is that true?

moelsayed: Not exactly. `pki.FetchCertificatesFromHost()` handles file not found internally, so `err != nil` will an actual error: something happened while doing a docker command, certificate parsing error..etc.

Issue #454, opened 2018-03-29
Azure Cloud Provider Writing Empty File On System


rossedman:
**RKE version:** v0.1.5-rc1

**Docker version: (`docker version`,`docker info` preferred)** 17.09.1-ce

**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)** 4.14.19-coreos

**Type/provider of hosts: (VirtualBox/Bare-metal/AWS/GCE/DO)** Azure

**cluster.yml file:** 

```
cloud_provider:
  name: azure
  cloud_config:
    aadClientId: <REDACTED>
    aadClientSecret: <REDACTED>
    location: centralus
    resourceGroup: am663-core
    subnetName: core
    subscriptionId: <REDACTED>
    vnetName: am663-core
    tenantId: <REDACTED>
    securityGroupName: am663-core-ssh
```

**Steps to Reproduce:**

Run RKE against an Azure cluster on CoreOS. The file is output to the system at `/etc/kubernetes/cloud-config.json` but nothing is written to the file. Causes API server to fail to start.

**Results:**
API server fails to start causing deployment to fail.

rossedman:
Referencing PR: https://github.com/rancher/rke/pull/449 
Tagging @galal-hussein

rossedman:
@HighwayofLife it didn't work for me but can try again.

galal-hussein:
@rossedman Can you post the logs of kubelet to see what went wrong, thanks.

rossedman:
@galal-hussein @HighwayofLife Still getting multiple failures. The only log entries without flag dumps is this:

```
I0403 15:14:46.681465       1 server.go:121] Version: v1.9.0-rancher2
I0403 15:14:46.681781       1 interface.go:360] Looking for default routes with IPv4 addresses
I0403 15:14:46.681824       1 interface.go:365] Default route transits interface "eth0"
I0403 15:14:46.682074       1 interface.go:174] Interface eth0 is up
I0403 15:14:46.682204       1 interface.go:222] Interface "eth0" has 2 addresses :[172.28.2.10/24 fe80::20d:3aff:fe97:4037/64].
I0403 15:14:46.682272       1 interface.go:189] Checking addr  172.28.2.10/24.
I0403 15:14:46.682308       1 interface.go:196] IP found 172.28.2.10
I0403 15:14:46.682348       1 interface.go:228] Found valid IPv4 address 172.28.2.10 for interface "eth0".
[...]

PR #533, opened 2018-04-20
Always write cloud config

HighwayofLife:
If an empty cloud config was written to the file system, which may be in the case of a previous failed deployment, subsequent runs will not write a good cloud config file. Likewise, currently any changes to a cloud config will not update existing cloud configuration. This PR allows changes to be made to a cloud config.


ibuildthecloud:
Should we be concerned about atomic writes (write to temp and move)?  It is possible that you write a empty or half file to a running and valid cluster.

HighwayofLife:
I'm not sure I understand your question. I do think that maybe write to temp and move may be a better operation than `echo "{json}" > file`.

ibuildthecloud:
@HighwayofLife sorry, i was just saying that both the existing code and this change would be more reliable if they used an atomic write approach. So `t=$(mktemp); echo $i > $t; mv $t file`

HighwayofLife:
@ibuildthecloud Updated with atomic-write approach. Thanks!

galal-hussein:
https://github.com/rancher/rke/issues/454

PR #539, opened 2018-04-23
Add os-release to get proper OS info

superseb:
Fix implemented for https://github.com/rancher/rancher/issues/12776 was to remove Host Info but this would be nice to have.

Only caveat is that if `/etc/os-release` doesn't exist, it will break.

### DIFF
--- a/cluster/plan.go
+++ b/cluster/plan.go
@@ -332,6 +332,8 @@ func (c *Cluster) BuildKubeletProcess(host *hosts.Host, prefixPath string) v3.Pr
 		"/dev:/host/dev:rprivate",
 		fmt.Sprintf("%s:/var/log/containers:z", path.Join(prefixPath, "/var/log/containers")),
 		fmt.Sprintf("%s:/var/log/pods:z", path.Join(prefixPath, "/var/log/pods")),
+		"/usr:/host/usr:ro",
+		"/etc:/host/etc:ro",
 	}
 
 	for arg, value := range c.Services.Kubelet.ExtraArgs {


alena1108:
@superseb `Only caveat is that if /etc/os-release doesn't exist, it will break.` - is there a way to check for that? 
or perhaps we can let user enable this option via extra binds (document it)

alena1108:
marked as WIP so we don't accidentally merge it till we discover the solution that won't break on all the OSs.

ibuildthecloud:
Bind mount `/etc:/host/etc:ro` and create a symlink

superseb:
Changes:

- Added `/usr` as most OS' symlink `/etc/os-release` to `/usr/lib/os-release`
- CentOS doesn't do this, so we still need `/etc`
- Added check if `/host/usr/lib/os-release` exist, else if `/host/etc/os-release` exists

Depends on https://github.com/rancher/rke-tools/pull/4

```
kubectl --kubeconfig kube_config_cluster.yml get nodes -o wide
NAME            STATUS    ROLES                      AGE       VERSION   EXTERNAL-IP   OS-IMAGE                       KERNEL-VERSION               CONTAINER-RUNTIME
206.189.5.143   Ready     controlplane,etcd,worker   1m        v1.10.1   <none>        Fedora 27 (Cloud Edition)      4.13.9-300.fc27.x86_64       docker://1.13.1
206.189.5.175   Ready     controlplane,etcd,worker   1m        v1.10.1   <none>        Ubuntu 16.04.4 LTS             4.4.0-119-generic            docker://17.12.1-ce
206.189.5.195   Ready     controlplane,etcd,worker   1m        v1.10.1   <none>        Ubuntu 17.10                   4.13.0-38-generic            docker://17.12.1-ce
206.189.5.203   Ready     controlplane,etcd,worker   1m        v1.10.1   <none>        CentOS Linux 7 (Core)          3.10.0-693.17.1.el7.x86_64   docker://1.13.1
206.189.5.223   Ready     controlplane,etcd,worker   1m        v1.10.1   <none>        Debian GNU/Linux 9 (stretch)   4.9.0-6-amd64                docker://17.12.1-ce
```

PR #547, opened 2018-04-25
Handle missing backup kube-etcd gracefully

moelsayed:
https://github.com/rancher/rancher/issues/12873
https://github.com/rancher/rancher/issues/12898

### DIFF
--- a/cluster/certificates.go
+++ b/cluster/certificates.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/rsa"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/rancher/rke/hosts"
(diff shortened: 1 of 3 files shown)


cluster/certificates.go
```diff
@@ -38,7 +39,22 @@ func SetUpAuthentication(ctx context.Context, kubeCluster, currentCluster *Clust
 				return err
 			}
 			if kubeCluster.Certificates != nil {
-				log.Infof(ctx, "[certificates] Certificate backup found on[%s] hosts", backupPlane)
+				log.Infof(ctx, "[certificates] Certificate backup found on [%s] hosts", backupPlane)
+
+				// make sure I have all the etcd certs, We need handle dialer failure for etcd nodes https://github.com/rancher/rancher/issues/12898
+				for _, host := range kubeCluster.EtcdHosts {
```
galal-hussein: so basically you regenerate etcd cert for all etcd hosts?

cluster/state.go
```diff
@@ -73,6 +74,21 @@ func (c *Cluster) GetClusterState(ctx context.Context) (*Cluster, error) {
 			if err != nil {
 				return nil, fmt.Errorf("Failed to Get Kubernetes certificates: %v", err)
 			}
+
+			// make sure I have all the etcd certs, We need handle dialer failure for etcd nodes https://github.com/rancher/rancher/issues/12898
+			for _, host := range activeEtcdHosts {
+				certName := pki.GetEtcdCrtName(host.InternalAddress)
+				if (currentCluster.Certificates[certName] == pki.CertificatePKI{}) {
```
galal-hussein: why using different condition than the one above

moelsayed: This one is a completely empty object. The one above is passed through `populateCertMap()` which adds some static values like env name and file path.

Issue #456, opened 2018-03-30
Backup etcd


RhubarbeM:
Set up a rke cluster in no time, simple, fast, just awesome work guys.

What and how should all this be backed up?

Thanks

soumyalj:
Tested with rke version v0.1.7-rc4/v0.1.7-rc4
etcd backup and restore is functional. The following uscases were tested:
1. 
```
1- set up a 2 node cluster, N1 -[c, w] N2 - [etcd]
2- create few pods
3- take backup
4- Power OFF etcd node (N2)
5- Add the snapshot to a new node
6. Run Restore with the new node in the config file
7. Do rke up again

```
2. 
```
1- create rke cluster
2- create few pods
3- take a backup using ./rke etcd backup
4- destroy the cluster using ./rke remove
5- remove all containers
6- add the snapshot in /opt/rke/etcdbackup on all hosts to make them all restore from the same snapshot
7- do ./rke etcd restore
8- do ./rke up again



```

PR #548, opened 2018-04-25
etcd backup/restore

galal-hussein:
https://github.com/rancher/rke-tools/pull/6
https://github.com/rancher/rke/issues/456


cluster/certificates.go
```diff
@@ -228,3 +228,21 @@ func fetchBackupCertificates(ctx context.Context, backupHosts []*hosts.Host, kub
...
+	var err error
+	for _, host := range kubeCluster.EtcdHosts {
+		certificates, err = pki.FetchCertificatesFromHost(ctx, kubeCluster.EtcdHosts, host, kubeCluster.SystemImages.Alpine, kubeCluster.LocalKubeConfigPath, kubeCluster.PrivateRegistriesMap)
+		if certificates != nil {
+			break
+		}
+	}
+	if err != nil {
+		return nil, nil, err
+	}
```
alena1108: can there be a case when certificates are nil?

cluster/cluster.go
```diff
@@ -77,7 +77,10 @@ func (c *Cluster) DeployControlPlane(ctx context.Context) error {
 	if len(c.Services.Etcd.ExternalURLs) > 0 {
 		log.Infof(ctx, "[etcd] External etcd connection string has been specified, skipping etcd plane")
 	} else {
-		if err := services.RunEtcdPlane(ctx, c.EtcdHosts, etcdNodePlanMap, c.LocalConnDialerFactory, c.PrivateRegistriesMap, c.UpdateWorkersOnly, c.SystemImages.Alpine); err != nil {
+		enableBackup := c.Services.Etcd.Backup
```
alena1108: would be nice to create a wrapper for etcdBackup, and pass it as a single variable to RunEtcdPlane

cluster/etcd.go
```diff
@@ -0,0 +1,57 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+func (c *Cluster) RestoreEtcdBackup(ctx context.Context, backupPath string) error {
+	// Stopping all etcd containers
+	for _, host := range c.EtcdHosts {
+		if err := docker.DoRemoveContainer(ctx, host.DClient, services.EtcdContainerName, host.Address); err != nil {
```
alena1108: can we extract cleanup logic to it's own method?

cmd/etcd.go
```diff
@@ -0,0 +1,140 @@
...
+	dockerDialerFactory hosts.DialerFactory,
+	configDir, backupName string) error {
+
+	log.Infof(ctx, "Starting Backup on etcd hosts")
+	kubeCluster, err := cluster.ParseCluster(ctx, rkeConfig, clusterFilePath, configDir, dockerDialerFactory, nil, nil)
+	if err != nil {
+		return err
+	}
+
+	err = kubeCluster.TunnelHosts(ctx, false)
```
alena1108: nit: can we use `if err := <>; err != nil, return err` construct

docker/docker.go
```diff
@@ -266,23 +266,24 @@ func StopRenameContainer(ctx context.Context, dClient *client.Client, hostname s
...
+	if _, err := WaitForContainer(ctx, dClient, hostname, oldContainerName); err != nil {
 		return nil
 	}
 	return RenameContainer(ctx, dClient, hostname, oldContainerName, newContainerName)
 
 }
 
-func WaitForContainer(ctx context.Context, dClient *client.Client, hostname string, containerName string) error {
+func WaitForContainer(ctx context.Context, dClient *client.Client, hostname string, containerName string) (int64, error) {
 	statusCh, errCh := dClient.ContainerWait(ctx, containerName, container.WaitConditionNotRunning)
```
alena1108: what is 1, 0? Can we add some comments explaining the use?

services/etcd.go
```diff
@@ -186,3 +204,87 @@ func IsEtcdMember(ctx context.Context, etcdHost *hosts.Host, etcdHosts []*hosts.
...
+		NetworkMode: container.NetworkMode("host"),
+	}
+
+	if once {
+		if err := docker.DoRunContainer(ctx, etcdHost.DClient, imageCfg, hostCfg, EtcdBackupOnceContainerName, etcdHost.Address, ETCDRole, prsMap); err != nil {
+			return err
+		}
+		status, err := docker.WaitForContainer(ctx, etcdHost.DClient, etcdHost.Address, EtcdBackupOnceContainerName)
+		if err != nil {
+			return nil
```
alena1108: is it by intent - returning nil when error is not nil?

docker/docker.go
alena1108: Also why do we need the status code? looks like in caller stacks we treat 1 as an error

Issue #564, opened 2018-05-02
rke for Windows fails to run


JasonvanBrackel:
**RKE version:**
v.0.1.6

**Docker version: (`docker version`,`docker info` preferred)**
```sh
docker info
Containers: 2
 Running: 0
 Paused: 0
 Stopped: 2
Images: 1
Server Version: 18.03.1-ce
Storage Driver: overlay2
 Backing Filesystem: extfs
 Supports d_type: true
 Native Overlay Diff: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge host macvlan null overlay
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: 773c489c9c1b21a6d78b5c538cd395416ec50f88
runc version: 4fc53a81fb7c994640722ac585fa9ca548971871
init version: 949e6fa
Security Options:
 seccomp
  Profile: default
Kernel Version: 4.9.87-linuxkit-aufs
Operating System: Docker for Windows
OSType: linux
Architecture: x86_64
CPUs: 2
Total Memory: 1.934 GiB
Name: linuxkit-00155d34a401
ID: FWIR:EIFZ:EHEP:TF73:VKMA:QHG5:6CV4:DZAP:AX7D:N7O3:T4BR:SKSU
Docker Root Dir: /var/lib/docker
Debug Mode (client): false
Debug Mode (server): true
 File Descriptors: 19
 Goroutines: 35
 System Time: 2018-05-02T03:40:14.8544Z
 EventsListeners: 1
Registry: https://index.docker.io/v1/
Labels:
Experimental: false
Insecure Registries:
 127.0.0.0/8
Live Restore Enabled: false
```

**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)**
Windows 10

**Type/provider of hosts: (VirtualBox/Bare-metal/AWS/GCE/DO)**
Azure

**cluster.yml file:**
```yaml
# default k8s version: v1.8.9-rancher1-1
# default network plugin: flannel
nodes:
  - address: 168.62.60.216
    user: jvb
    role: [worker]
    ssh_key_path: "/home/jvb/.ssh/id_rsa"
  - address: 168.62.60.204
    user: jvb
    role: [controlplane]
    ssh_key_path: "/home/jvb/.ssh/id_rsa"
  - address: 168.62.58.233
    user: jvb
    role: [etcd]
[...]

PR #569, opened 2018-05-02
Windows binaries

galal-hussein:


moelsayed:
https://github.com/rancher/rke/issues/564

Issue #302, opened 2018-02-02
Add self anti affinity to sky dns 


alena1108:
We should add self anti affinity to sky dns pods in rke config

https://github.com/kubernetes/kubernetes/pull/57683

PR #585, opened 2018-05-07
Add pod antiaffinity to kubedns template

galal-hussein:
#302

### DIFF
--- a/templates/kubedns.go
+++ b/templates/kubedns.go
@@ -113,6 +113,17 @@ spec:
       annotations:
         scheduler.alpha.kubernetes.io/critical-pod: ''
     spec:
+      affinity:
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - weight: 100
+            podAffinityTerm:
+              labelSelector:
+                matchExpressions:
+                  - key: k8s-app
+                    operator: In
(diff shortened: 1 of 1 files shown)


templates/kubedns.go
```diff
@@ -113,6 +113,17 @@ spec:
       annotations:
         scheduler.alpha.kubernetes.io/critical-pod: ''
     spec:
+      affinity:
```
alena1108: I recall seeing k8s issue in 1.6 for upgrades. With limited number of hosts, and self anti affinity, it was possible for the upgrade to fail. Can you check with k8s upstream to see if the issue is still there. If it is, we might need to put some comment to this code, so next time on the addon upgrade, we would change the corresponding label, so the addons scheduling won'g get conflicts during the upgrade.

willmao: @alena1108 @galal-hussein 

why schedule kube-dns on the same node,? would it be a SPOF if the node has network issue ?

alena1108: @willmao it is `anti` affinity rule. Implying the pods won't be on the same node

willmao: sorry, i misunderstood the setting, but one day i saw the kube-dns pods scheduled on the same node, so i found the code here..

thanks for your time!

PR #587, opened 2018-05-07
Fix Openstack cloud config output

alexandrem:
RKE generates a json config when `cloud_config` is defined with a `cloud_provider` name `openstack`, but it should instead generate the appropriate text INI format with a `[Global]` section header using the cloud config map.

This PR separates the cloud config parsing/output block with a different logic when cloud provider is  `openstack`, otherwise falls back to the previous generic cloud config.

Right now it still uses the `cloud-config.json` filename as output. This limits the scope of the change, but perhaps it'd be clearer to have the expected `cloud.conf` instead.

This adds unit tests for the cloud config parsing methods and a new dependency on "testify", which is already used by rancher server codebase. The vendor config and folder was updated accordingly with the new dep.

This addresses https://github.com/rancher/rancher/issues/13338 https://github.com/rancher/rancher/issues/13115


galal-hussein:
@alexandrem Thanks for the PR, i think there are two problems with the implementation:

1- RKE doesn't provide a generic way to configure different formats for the cloud-config, since json is not enough, we need to define a way to add different format based on the cloud provider type, my suggestion is to implement a simple interface that each cloud provider implements on its own, for example:
```
type CloudProvider interface {
Initialize()
GenerateConfigFile() error
ProviderName() string
}
```
2- The PR only adds the [Global] section of the openstack config which means the user can't define [Blockstorage], [Loadbalancer], etc. i get that this is mainly because we don't have static types for each provider, but i think we have to abandon the generic type `cloud-config` and add options in rancher/types for each provider separately like we did in Azure:
https://github.com/rancher/types/blob/master/apis/management.cattle.io/v3/rke_types.go#L348-L417
i think also each cloud provider type can be separated in its own file

alexandrem:
@galal-hussein 
No problem, I expected this to not be totally aligned with future needs.

For #1, what would an `initialize` do in this case exactly? Or do you have more some kind of "register" of the pluggable interface? The latter tends to be more often expressed with golang module init(), like the current cloud provider plugin model in k8s.io/kubernetes.

For #2, I thought as well that a specific provider data struct in the rancher types would be better for Openstack and most other providers. However, I'm concerned about the evolution of those inline struct fields over time. There doesn't seem to be any versioning embedded, so I suppose the general design recommendation is we can add new fields but never change the existing ones. Then there's the possible issue about backward/future compatibility of the Kubernetes cloud config with newer  releases over time, if that ever changes. Maybe I'm just over thinking too much, WDYT?

There's a neat trick I used to do that I borrowed from the cluster-api project, which consists of embedding the apimachinery `TypeMeta` object in nested structs, then you can express it natively in yaml definition, which ultimately can be unserialized correctly with correct type/version using the RawExtension. This is particularly useful to accept any kind of nested apimachinery typed objects in Kubernetes CRDs, like cloud providers. Just an idea, otherwise we can keep it simple with the current inline struct below CloudProvider in rancher types, being aware of its limitations.

Overall, do you think we should separate the design discussion to a bigger issue (e.g https://github.com/rancher/rancher/issues/13338), to address all future integrations, like the rancher server node template, and possibly other new provider types like the vmware one?

alexandrem:
Since the scope of this change is quite limited, could we consider merging this until we come up with a better design for future cloud providers interface and types?

This change allows to create clusters on openstack via both RKE and the rancher server using the current key/value config layout already in place.

The integration is not perfect since it's only the [Global] config, but creation works and we can have 
 persistent volumes and external IPs.

This would also remain backward compatible when we add the new typed struct later on.

greatbn:
@alexandrem I think you should use two levels for Opensack Cloud Provider
Because section `Global`  is for authentication  , there are some section such as `LoadBalancer` and `BlockStorage`

alexandrem:
@greatbn Yes, in that case that requires changing the RKE types for the new structure.

What do you have in mind? Do you mean adding nested fields for those sections?

```yaml
cloud_config:
    global:
        xxx
    load_balancer:
        xxx
```

Doesn't look very pretty. Also, the real problem is that right now the standard in the RKE types is to keep that inlined with struct fields.

I suppose to follow that trend we could just add a new openstack provider struct and have those fields below inlined again:

```yaml
cloud_config:
    auth-url: xxx
    lb-subnet-id: xxx
```

We might have to keep good variable prefixes to ensure nothing clashes.

galal-hussein:
WIll close in favor of https://github.com/rancher/rke/pull/667
The PR will allow the user to provide all sections of openstack, vmware and any custom cloud config file

Issue #103, opened 2017-12-05
Allow use of bastion host/jumphost


superseb:
In some environments (DMZ), you can only SSH through a certain host (usually called bastion host/jump host). Would be nice to be able to configure this.

rossedman:
@superseb Interested to see where this is at. Thanks

HighwayofLife:
I'm currently doing this for my own implementation as well, while I can ssh over port 22 into my hosts, I cannot reach them over port 6443, for example, so I tunnel through a jump host using sshuttle. It would be cool to build this functionality into RKE, just pointing out that there is a decent workaround using sshuttle.

rossedman:
@galal-hussein Just worked for me! Thanks so much, this is great.

rossedman:
@galal-hussein Hitting a couple of errors with this. May be releated to 1.7rc1 or could be my host configuration or something else. 

```
FATA[0089] Can't get Docker containers for host [172.28.2.7]: error during connect: Get http://%2Fvar%2Frun%2Fdocker.sock/v1.24/containers/json?all=1&limit=0: Failed to dial ssh using address [172.28.2.7:22]: Failed to connect to the bastion host [myhost.com:22]: ssh: handshake failed: read tcp 192.168.86.144:63024->104.209.157.247:22: read: connection reset by peer
```

This happens when using the bastion. Seems to time out or terminate connection. Can open another issue if needed.

HighwayofLife:
@rossedman can you ssh directly to your bastion host? Are you referencing the ssh-key path correctly in the cluster.yml file for the bastion host?

rossedman:
@HighwayofLife yeah it makes it into the servers and eventually times out. I'm also getting a lot of other SELinux errors though which is strange. I downgraded to 1.6 and am still having those issues too so I'm not sure if is related.

PR #590, opened 2018-05-08
use Bastion/Jump host

galal-hussein:
Depend on https://github.com/rancher/types/pull/422

#103

### DIFF
--- a/cluster/network.go
+++ b/cluster/network.go
@@ -220,10 +220,13 @@ func (c *Cluster) CheckClusterPorts(ctx context.Context, currentCluster *Cluster
 	if err := c.runServicePortChecks(ctx); err != nil {
 		return err
 	}
-	if c.K8sWrapTransport == nil {
+	// Skip kubeapi check if we are using custom k8s dialer or bastion/jump host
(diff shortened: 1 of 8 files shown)


cluster/network.go
```diff
@@ -231,7 +231,7 @@ func (c *Cluster) CheckClusterPorts(ctx context.Context, currentCluster *Cluster
 	if err := c.runServicePortChecks(ctx); err != nil {
 		return err
 	}
-	if c.K8sWrapTransport == nil {
+	if c.K8sWrapTransport == nil && len(c.BastionHost.Address) == 0 {
```
moelsayed: Should we log something about this ?

hosts/dialer.go
```diff
@@ -19,22 +21,41 @@ type dialer struct {
...
 	dockerSocket    string
 	useSSHAgentAuth bool
+	useBastionHost  bool
+	bastionDialer   *dialer
 }
 
 func newDialer(h *Host, kind string) (*dialer, error) {
+	// Check for Bastion host connection
+	var bastionDialer *dialer
+	var useBastionHost bool
```
moelsayed: Do we need this ? can't we just `if bastionDialer != nil` ?

PR #593, opened 2018-05-09
removing redundant extra binds

aemneina:
Extra binds example prevents kubelet from starting up. The path seems to be already present when inspecting my kubelet.


aemneina:
@superseb @galal-hussein or @moelsayed care to review/validate?

Issue #598, opened 2018-05-10
SELinux relabeling of /var/lib is not allowed: "operation not supported"


rossedman:
**RKE version:**

1.7-rc1

**Docker version: (`docker version`,`docker info` preferred)**

1.12

**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)**
NAME="Container Linux by CoreOS"
ID=coreos
VERSION=1632.3.0
VERSION_ID=1632.3.0
BUILD_ID=2018-02-14-0338
PRETTY_NAME="Container Linux by CoreOS 1632.3.0 (Ladybug)"
ANSI_COLOR="38;5;75"
HOME_URL="https://coreos.com/"
BUG_REPORT_URL="https://issues.coreos.com"
COREOS_BOARD="amd64-usr"

**Type/provider of hosts: (VirtualBox/Bare-metal/AWS/GCE/DO)**
Azure

**cluster.yml file:**
```
#
# this file was autogenerated, do not edit manually
# template generated from: rke.yaml.j2
#

nodes:
  # control plane
  - address: 172.28.2.6
    user: core
    role: ["controlplane", "etcd"]
    hostname_override: am660-kube0-ctl0
  - address: 172.28.2.5
    user: core
    role: ["controlplane", "etcd"]
    hostname_override: am660-kube0-ctl1
  - address: 172.28.2.7
    user: core
    role: ["controlplane", "etcd"]
    hostname_override: am660-kube0-ctl2

  # general compute
  - address: 172.28.2.16
    user: core
    role: ["worker"]
    hostname_override: am660-kube0-gpc1
  - address: 172.28.2.10
    user: core
    role: ["worker"]
    hostname_override: am660-kube0-gpc0
  - address: 172.28.2.18
    user: core
    role: ["worker"]
    hostname_override: am660-kube0-gpc3
  - address: 172.28.2.14
    user: core
    role: ["worker"]
    hostname_override: am660-kube0-gpc2
  - address: 172.28.2.11
    user: core
    role: ["worker"]
    hostname_override: am660-kube0-gpc5
  - address: 172.28.2.12
    user: core
    role: ["worker"]
    hostname_override: am660-kube0-gpc4

  # load balancers
  - address: 172.28.2.15
    user: core
    role: ["worker"]
    hostname_override: am660-kube0-lbe0
  - address: 172.28.2.17
    user: core
    role: ["worker"]
    hostname_override: am660-kube0-lbe1
  - address: 172.28.2.13
[...]

PR #599, opened 2018-05-10
Add prefix path for CoreOS

galal-hussein:
#598


cluster/plan.go
```diff
@@ -656,6 +658,8 @@ func (c *Cluster) getPrefixPath(osType string) string {
 		prefixPath = B2DPrefixPath
 	} else if strings.Contains(osType, ROS) {
 		prefixPath = ROSPrefixPath
+	} else if strings.Contains(osType, CoreOS) {
```
moelsayed: Maybe use switch instead ?

Issue #609, opened 2018-05-14
Ability to upgrade rancher version with a rancher flag


deniseschannon:
Using an existing cluster.yml file for RKE that launched Rancher, we'd like a flag to be able to upgrade the tagged version that you want for Rancher. 

```
rke up --config cluster.yml --rancher-tag <tag of the rancher/rancher image>
```

deniseschannon:
We're going to update rancher using a different method.

PR #617, opened 2018-05-15
Add option to upgrade rancher 2.x deployments

moelsayed:
https://github.com/rancher/rke/issues/609


alena1108:
Not a fan of rancher specific - which is really an addon - parameter...Although we use rke to setup rancher, it is really a generic k8s installer.

moelsayed:
This will be close in favor of a different approach.

Issue #596, opened 2018-05-09
K8 Cluster created with weave don't have CNI loopback driver


YoranSys:
**RKE version:**

`   v0.1.6`

**Docker version: (`docker version`,`docker info` preferred)**

```
Containers: 1741
 Running: 10
 Paused: 0
 Stopped: 1731
Images: 6
Server Version: 17.03.2-ce
Storage Driver: overlay2
 Backing Filesystem: extfs
 Supports d_type: true
 Native Overlay Diff: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins: 
 Volume: local
 Network: bridge host macvlan null overlay
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: 4ab9917febca54791c5f071a9d1f404867857fcc
runc version: 54296cf40ad8143b62dbcaa1d90e520a2136ddfe
init version: 949e6fa
Security Options:
 seccomp
  Profile: default
Kernel Version: 4.9.0-6-amd64
Operating System: Debian GNU/Linux 9 (stretch)
OSType: linux
Architecture: x86_64
CPUs: 2
Total Memory: 7.801 GiB
Name: k8si-1
ID: OCLT:VTKY:MWLA:XF6D:GD72:YHVC:UOXA:G5KP:ZDSZ:YL2R:VVHR:RLCK
Docker Root Dir: /var/lib/docker
Debug Mode (client): false
Debug Mode (server): false
Registry: https://index.docker.io/v1/
Experimental: false
Insecure Registries:
 127.0.0.0/8
Live Restore Enabled: false

WARNING: No swap limit support
```

**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)**

```
PRETTY_NAME="Debian GNU/Linux 9 (stretch)"
NAME="Debian GNU/Linux"
VERSION_ID="9"
VERSION="9 (stretch)"
```

Kernel : `4.9.0-6-amd64`

**Type/provider of hosts: (VirtualBox/Bare-metal/AWS/GCE/DO)**
`
Bare-metal`

**cluster.yml file:**

```yaml
---

network:
  plugin: weave


services:
  kube-api:
    service_cluster_ip_range: 10.43.0.0/16
  kube-controller:
    cluster_cidr: 10.42.0.0/16
    service_cluster_ip_range: 10.43.0.0/16
    extra_args:
      cluster-name: "cluster01"

ingress:
  provider: nginx
[...]

Issue #286, opened 2018-01-30
RKE RHEL-7.4 with weave: "Failed to deploy addon execute job: Failed to get job complete status"


sabiodelhielo:
**RKE version:**
v0.1.0
**Docker version: (`docker version`,`docker info` preferred)**
native docker version for RHEL (1.12.6)
**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)**
NAME="Red Hat Enterprise Linux Server"
VERSION="7.4 (Maipo)"
ID="rhel"
ID_LIKE="fedora"
VARIANT="Server"
VARIANT_ID="server"
VERSION_ID="7.4"
PRETTY_NAME="Red Hat Enterprise Linux Server 7.4 (Maipo)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:7.4:GA:server"
HOME_URL="https://www.redhat.com/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 7"
REDHAT_BUGZILLA_PRODUCT_VERSION=7.4
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="7.4"
3.10.0-693.el7.x86_64
**Type/provider of hosts: (VirtualBox/Bare-metal/AWS/GCE/DO)**
AWS
**cluster.yml file:**
ignore_docker_version: true
ssh_key_path: {{ master_ssh_key_path }}
network:
plugin: canal
nodes:

address: {{ ip_address_0 }}
user: ubuntu
role: [controlplane]
address: {{ ip_address_1 }}
user: ubuntu
role: [etcd]
address: {{ ip_address_2 }}
user: ubuntu
role: [worker]
services:
etcd:
image: quay.io/coreos/etcd:latest
kube-api:
image: rancher/k8s:v1.8.3-rancher2
kube-controller:
image: rancher/k8s:v1.8.3-rancher2
scheduler:
image: rancher/k8s:v1.8.3-rancher2
kubelet:
image: rancher/k8s:v1.8.3-rancher2
kubeproxy:
image: rancher/k8s:v1.8.3-rancher2

Note there is no addons in the yaml

**Steps to Reproduce:**
once the three nodes are provision on AWS with key pair
rke up

**Results:**
[...]

PR #623, opened 2018-05-16
Add loopback cni to weave template

galal-hussein:
https://github.com/rancher/rke/issues/286
https://github.com/rancher/rke/issues/596


templates/weave.go
```diff
@@ -27,7 +27,9 @@ items:
           containers:
             - name: weave
               command:
-                - /home/weave/launch.sh
+              - "sh"
+              - "-c"
+              - "mkdir /host/opt/cni/bin && curl -fsSL 'https://github.com/containernetworking/cni/releases/download/v0.4.0/cni-amd64-v0.4.0.tgz' | tar xvz -C  /host/opt/cni/bin/ && /home/weave/launch.sh"
```
alena1108: what if it's an airgapped setup?

Issue #637, opened 2018-05-18
rke snapshot-restore brings 502 Bad Gateway


harutyundermenjyan:
Hi,
here is my steps:
1. installing rke with HA version: 1 nginx, 3 etcd, 3 controplane, 3 worker (`./rke_linux-amd64 up --config rancher-cluster.yml` (https://rancher.com/docs/rancher/v2.x/en/installation/ha-server-install/)
2. creating deployment with a few replicas(for test purposes)
3. creating etcd backup `./rke_linux-amd64 etcd snapshot-save --name snapshot --config rancher-cluster.yml`
4. removing my cluster in order to test the restore of etcd `./rke_linux-amd64 remove --config rancher-cluster.yml` and then removing all containers and images from hosts separately: 
```./cleanup.sh ; docker stop `docker ps |awk '{print $1}' |paste -sd' '` ; docker rm `docker ps -a |awk '{print $1}' |paste -sd' '` ; docker rmi `docker images |awk '{print $3}' |paste -sd' '` ; service docker stop && rm -rf /var/lib/docker/ ; umount -l -f /var/lib/rancher/volumes/ ; umount -l -f /var/lib/rancher ; rm -rf /var/lib/rancher ; umount -l -f /var/lib/kubelet ; service docker start && docker ps```
I've downloaded the./cleanup.sh script from @superseb repo (https://gist.github.com/superseb/2cf186726807a012af59a027cb41270d)
4. installing the rke with the same way/config as I've described in the 1st step
5. coping the saved snapshot to all etcd hosts in the `/opt/rke/etcd-snapshots/` directory
6. restoring from snapshot `./rke_linux-amd64 etcd snapshot-restore --name snapshot --config rancher-cluster.yml `
7. get 502 Bad Gateway :(
also the pods aren't recreated after delete

```
kubectl get po --all-namespaces --show-all
NAMESPACE       NAME                                      READY     STATUS      RESTARTS   AGE
cattle-system   cattle-84f9cd8589-j7nl2                   1/1       Running     0          4h
cattle-system   cattle-cluster-agent-7657c8c794-rw9k4     1/1       Running     0          4h
cattle-system   cattle-node-agent-2mvhd                   1/1       Running     0          4h
cattle-system   cattle-node-agent-bp872                   1/1       Running     0          4h
cattle-system   cattle-node-agent-flvkt                   1/1       Running     0          4h
default         ubuntutest-94846979c-85g4b                1/1       Running     0          4h
[...]

PR #657, opened 2018-05-29
Add certificate bundle backup

galal-hussein:
https://github.com/rancher/rke/issues/637

### DIFF
--- a/cluster/certificates.go
+++ b/cluster/certificates.go
@@ -11,6 +11,7 @@ import (
 	"github.com/rancher/rke/k8s"
 	"github.com/rancher/rke/log"
 	"github.com/rancher/rke/pki"
+	"github.com/rancher/rke/services"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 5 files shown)


cluster/certificates.go
```diff
@@ -247,3 +248,41 @@ func fetchCertificatesFromEtcd(ctx context.Context, kubeCluster *Cluster) ([]byt
 	clientkey := cert.EncodePrivateKeyPEM(certificates[pki.KubeNodeCertName].Key)
 	return clientCert, clientkey, nil
 }
+
+func (c *Cluster) SaveBackupCertificateBundle(ctx context.Context) error {
+	var backupHosts []*hosts.Host
+	if len(c.Services.Etcd.ExternalURLs) > 0 {
```
alena1108: this code looks very similar to extract* logic (lines 272-*, except for a pki* code. I'd suggest we extract all the common logic

PR #663, opened 2018-05-31
Windows support 

thxCode:
1. Support to configure Flannel as "host-gw" backend;
2. Define the network component yaml and ingress controller yaml only schedule to non-Windows node;
3. Support to configure Docker container's port publishing.


thxCode:
@orangedeng

HighwayofLife:
Adding my voice here, would be wonderful to have this support available. How can we help test this?

olljanat:
@HighwayofLife FYI, IMO biggest issue here is that Kubernetes support for Windows is not even near of production ready and [even AKS does not support Windows nodes yet](https://docs.microsoft.com/en-us/azure/aks/faq#can-i-run-windows-server-containers-on-aks).

Anyway, I was able to get Rancher agent running on Win Srv, version 1803 using instructions given on https://github.com/rancher/rancher/pull/13796
But we did no go decision for all our Kubernetes plans will stay on Docker Swarm at least couple of years until Microsoft gets their stuff more stable. 
I recommend anyone who are interested that progress to ask it from their premier support contact.

What comes to tests RKE, you should be able to do that by building PR sources on Windows.

services/services.go
```diff
@@ -88,12 +89,14 @@ func GetProcessConfig(process v3.Process) (*container.Config, *container.HostCon
 	}
 	// var pidMode container.PidMode
 	// pidMode = process.PidMode
+	_, portBindings, _ := nat.ParsePortSpecs(process.Publish)
```
StrongMonkey: Does this method return an error? If so, do not ignore it. You need to return and handle the error if possible.

StrongMonkey: thanks, make senses.

StrongMonkey:
LGTM. @deniseschannon do you want to get @galal-hussein or @moelsayed to take a look? i am not really familiar with this area.

cjellick:
LGTM

cjellick:
@galal-hussein and @moelsayed when you get back, please do a post mordem review for anything that needs changed

PR #665, opened 2018-05-31
Make control node only unschedulable

galal-hussein:


cluster/reconcile.go
```diff
@@ -72,6 +73,9 @@ func reconcileWorker(ctx context.Context, currentCluster, kubeCluster *Cluster,
 		if host.IsEtcd {
 			host.ToDelTaints = append(host.ToDelTaints, unschedulableEtcdTaint)
 		}
+		if host.IsControl {
```
alena1108: we don't check if the host is worker here?

Issue #584, opened 2018-05-07
Enable addons upgrades


moelsayed:
Currently, during upgrade, we only upgrade core components only. We don't upgrade any of the addons.  We need to support that.

sangeethah:
Tested with rke version v0.1.8-rc5

Deployed cluster with following yml:

```
addon_job_timeout: 60
addons: |-
  ---
  apiVersion: v1
  kind: Service
  metadata:
   name: k8test2
   labels:
     k8s-app: k8test2-service
  spec:
   ports:
   - port: 80
     targetPort: 80
     protocol: TCP
     name: http
   selector:
     k8s-app: k8test2-service
  ---
  apiVersion: v1
  kind: ReplicationController
  metadata:
   name: k8testrc2
  spec:
   replicas: 2
   selector:
     k8s-app: k8test2-service
   template:
     metadata:
       labels:
         k8s-app: k8test2-service
     spec:
       terminationGracePeriodSeconds: 60
       containers:
       - name: testcontainer
         image: nginx
         ports:
         - containerPort: 80
  ---
  apiVersion: extensions/v1beta1
  kind: Ingress
  metadata:
   name: ingress198
   annotations:
     http.port: "9198"
  spec:
   rules:
   -  http:
       paths:
       - path: /name.html
         backend:
           serviceName: test123
           servicePort: 80
```
Add on service/rc and ingress get created as expected.

But when I change the image name from RC and rerun the ```rke up``` command , I see the image change reflected in the RC. But pods dont get recreated.

sangeethah:
When testing with a deployment in addons , I am able to upgrade the container images of the  deployments and see the pods getting recreated.

PR #680, opened 2018-06-09
Add support for addons upgrades

moelsayed:
https://github.com/rancher/rke/issues/584

This PR does two things
- It provides support for updating addons configuration, both system and user defined addons. 
- If fixes an issue where an addon job is skipped on re-runs even if it's not in a completed state. 

Note: It's not possible to change cluster cidr configuration using this. It's not supported by Kubernetes:
- https://github.com/kubernetes/kubernetes/issues/50305
- https://github.com/kubernetes-incubator/kube-aws/issues/1307

### DIFF
--- a/cluster/addons.go
+++ b/cluster/addons.go
@@ -200,7 +200,7 @@ func (c *Cluster) doAddonDeploy(ctx context.Context, addonYaml, resourceName str
 		}
 	}
 
-	err := c.StoreAddonConfigMap(ctx, addonYaml, resourceName)
+	addonUpdated, err := c.StoreAddonConfigMap(ctx, addonYaml, resourceName)
 	if err != nil {
 		return &addonError{fmt.Sprintf("Failed to save addon ConfigMap: %v", err), isCritical}
 	}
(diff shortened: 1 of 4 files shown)


cluster/addons.go
```diff
@@ -219,23 +219,25 @@ func (c *Cluster) doAddonDeploy(ctx context.Context, addonYaml, resourceName str
 	if err != nil {
 		return &addonError{fmt.Sprintf("Failed to generate addon execute job: %v", err), isCritical}
 	}
-	err = c.ApplySystemAddonExcuteJob(addonJob)
+	err = c.ApplySystemAddonExcuteJob(addonJob, addonUpdated)
```
galal-hussein: can you merge the if and the error, `if err := .....; err != nil {`

k8s/job.go
```diff
@@ -24,6 +30,20 @@ func ApplyK8sSystemJob(jobYaml, kubeConfigPath string, k8sWrapTransport WrapTran
...
+	jobStatus, err := getK8sJobStatus(k8sClient, job.Name, job.Namespace)
+	if err != nil {
+		return err
+	}
+	// if the addon configMap is updated, or the previous job is not completed,
+	// I will remove the existing job first, if any
+	if addonUpdated || (jobStatus.Created && !jobStatus.Completed) {
+		logrus.Debugf("[k8s] replacing job %s.. ", job.Name)
+		if err := deleteK8sJob(k8sClient, job.Name, job.Namespace); err != nil && !apierrors.IsNotFound(err) {
+			return err
```
alena1108: if the error is not found, we should not sleep

cluster/addons.go
```diff
@@ -219,23 +219,25 @@ func (c *Cluster) doAddonDeploy(ctx context.Context, addonYaml, resourceName str
...
 	kubeClient, err := k8s.NewClient(c.LocalKubeConfigPath, c.K8sWrapTransport)
 	if err != nil {
-		return err
+		return updated, err
 	}
 	timeout := make(chan bool, 1)
 	go func() {
 		for {
-			err := k8s.UpdateConfigMap(kubeClient, []byte(addonYaml), addonName)
+
```
alena1108: do we ever timeout? What if this loop runs forever?

moelsayed: Yes, we timeout here https://github.com/rancher/rke/pull/680/files#diff-e072bfb1478c2c4b66ce16353b1b3027R254

PR #682, opened 2018-06-10
Possibility to enable PersistentVolumeClaimResize

Itiho:


moelsayed:
@Itiho Unfortunately that's still an alpha feature. We try to enable as few of those as possible. 

If  need to enable this, you can use the `extra_args` option to enable it and it's required feature gate flag.

Itiho:
How can I enable with extra args?

Issue #691, opened 2018-06-13
RKE etcd "snapshot-restore" ] does not show a warning when certificate bundle is not present


soumyalj:
**RKE version:**
rke version v0.1.8-rc1

**Steps to Reproduce:**
1. Create a two node RKE cluster
N1 - control, etcd
N2 - Worker
2. Create few pods
3. Take a one time snapshot
4. Create a new node
5. Add the snapshot to a new node and do not copy the certificate bundle to the new node
6. Run Restore with the new node in the config file

**Results:**
etcd "snapshot-restore" does not provide a warning that the certificate bundle is not present and fails.

soumyalj:
Tested with  rke version v0.1.8-rc5
Create a two node RKE cluster
N1 - control, etcd
N2 - Worker
Create few pods
Take a one time snapshot
Create a new node
Add the snapshot to a new node and do not copy the certificate bundle to the new node
Run Restore with the new node in the config file

etcd "snapshot-restore" provides an error message that certificate bundle is missing and exits
```
LAPTOP-6CQEFUBT:rke soumya$ ./rke_darwin-amd64 etcd snapshot-restore --name soumyasnapshot --config testetcd.yml 
INFO[0000] Starting restoring snapshot on etcd hosts    
INFO[0000] [dialer] Setup tunnel for host abc
INFO[0000] [dialer] Setup tunnel for host def
INFO[0001] [remove/etcd] Successfully removed container on host abc
INFO[0001] [hosts] Cleaning up host def
INFO[0001] [hosts] Running cleaner container on host [abc]
INFO[0002] [etcd] Restoring [soumyasnapshot] snapshot on etcd host abc
INFO[0003] [etcd] Successfully started [etcd-restore] container on hostabc
INFO[0003] [etcd] Building up etcd plane..              
INFO[0003] [etcd] Successfully started [etcd] container on host def
INFO[0004] [etcd] Successfully started [rke-log-linker] container on host def
INFO[0004] [remove/rke-log-linker] Successfully removed container on host abc
INFO[0004] [etcd] Successfully started etcd plane..     
INFO[0005] [certificates] Successfully started [rke-bundle-cert] container on host [abc] 
FATA[0005] Failed to extract certificate bundle on host def, please make sure etcd bundle exist in /opt/rke/etcd-snapshots/pki.bundle.tar.gz: Failed to run certificate bundle extract, exit status is: 1 



```

Issue #690, opened 2018-06-13
RKE etcd "snapshot-restore" fails to restore if it can't find the tmp cert directory


soumyalj:
**RKE version:**rke version v0.1.8-rc1

**Docker version: (`docker version`,`docker info` preferred)** 1.12.6

**cluster.yml file:**
```
auth:
  strategy: x509
network:
  plugin: flannel
nodes:
  - address:abc
    user: root
    role: [etcd, controlplane]
  #  hostname_override: ""
  - address: def
    user: root
    role: [worker]
 services:
  etcd:
    snapshot: true
    creation: 5m0s
    retention: 24h
  kube-api:
    #image: rancher/k8s:v1.9.5-rancher1-1
    service_cluster_ip_range: 10.233.0.0/18
    extra_args:
      v: 4
  kube-controller:
    #image: rancher/k8s:v1.9.5-rancher1-1
    cluster_cidr: 10.233.64.0/18
    service_cluster_ip_range: 10.233.0.0/18
  scheduler:
    #image: rancher/k8s:v1.9.5-rancher1-1
  kubelet:
    #image: rancher/k8s:v1.9.5-rancher1-1
    cluster_domain: cluster.local
    cluster_dns_server: 10.233.0.3
    infra_container_image: gcr.io/google_containers/pause-amd64:3.0
  kubeproxy:
    #image: rancher/k8s:v1.9.5-rancher1-1

  system_images:
    etcd: rancher/etcd:v3.0.17
    kubernetes: rancher/hyperkube:v1.10.1-rancher1
    alpine: rancher/rke-tools:v0.1.4
    nginx_proxy: rancher/rke-tools:v0.1.4
    cert_downloader: rancher/rke-tools:v0.1.4
    kubernetes_services_sidecar: rancher/rke-tools:v0.1.4
    kubedns: rancher/k8s-dns-kube-dns-amd64:1.14.5
    dnsmasq: rancher/k8s-dns-dnsmasq-nanny-amd64:1.14.5
    kubedns_sidecar: rancher/k8s-dns-sidecar-amd64:1.14.5
    kubedns_autoscaler: rancher/cluster-proportional-autoscaler-amd64:1.0.0
    flannel: rancher/coreos-flannel:v0.9.1
    flannel_cni: rancher/coreos-flannel-cni:v0.2.0

  kubernetes_version: v1.9.5-rancher1-1 

```

**Steps to Reproduce:**
1. Create a two node RKE cluster
     N1 - control, etcd
[...]

PR #692, opened 2018-06-13
Fix untar tmp ssl certificates

galal-hussein:
https://github.com/rancher/rke/issues/691
https://github.com/rancher/rke/issues/690

### DIFF
--- a/cluster/certificates.go
+++ b/cluster/certificates.go
@@ -265,14 +265,22 @@ func (c *Cluster) SaveBackupCertificateBundle(ctx context.Context) error {
 func (c *Cluster) ExtractBackupCertificateBundle(ctx context.Context) error {
 	backupHosts := c.getBackupHosts()
 	var errgrp errgroup.Group
-
+	errList := []string{}
 	for _, host := range backupHosts {
 		runHost := host
 		errgrp.Go(func() error {
-			return pki.ExtractBackupBundleOnHost(ctx, runHost, c.SystemImages.Alpine, services.EtcdSnapshotPath, c.PrivateRegistriesMap)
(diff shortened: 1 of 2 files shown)


cluster/certificates.go
```diff
@@ -269,7 +269,10 @@ func (c *Cluster) ExtractBackupCertificateBundle(ctx context.Context) error {
 	for _, host := range backupHosts {
 		runHost := host
 		errgrp.Go(func() error {
-			return pki.ExtractBackupBundleOnHost(ctx, runHost, c.SystemImages.Alpine, services.EtcdSnapshotPath, c.PrivateRegistriesMap)
+			if err := pki.ExtractBackupBundleOnHost(ctx, runHost, c.SystemImages.Alpine, services.EtcdSnapshotPath, c.PrivateRegistriesMap); err != nil {
```
alena1108: @galal-hussein could you explain the nature of the fix? We don't care if the extract fails on all the backup hosts, and just log it in warn?

galal-hussein: yes, there are some user cases where the ssl are already deployed on the machines somehow, for example:
- 1 etcd
- 1 control 
- 1 worker 

and you replace the etcd server, then in this case the certificates are already there on the control plane and in the next run it will be extracted normally

alena1108: @galal-hussein `in the next run it will be extracted normally` what if the next run doesn't happen? Lets say rancher calls create/update on a cluster, and etcd backup fails by some other reason but `ssl are already deployed on the machines somehow`? The create/update call will succeed, and the cluster will remain missing etcd backup till the next update - which might not happen in days

alena1108:
LGTM

PR #709, opened 2018-06-21
kubelet volumeplugins path does not contain prefix

chasebolt:
adjusting of the volumeplugins path was missed in the PR that added prefix support.

### DIFF
--- a/cluster/plan.go
+++ b/cluster/plan.go
@@ -324,7 +324,7 @@ func (c *Cluster) BuildKubeletProcess(host *hosts.Host, prefixPath string) v3.Pr
 		"kubeconfig":                pki.GetConfigPath(pki.KubeNodeCertName),
 		"client-ca-file":            pki.GetCertPath(pki.CACertName),
 		"anonymous-auth":            "false",
-		"volume-plugin-dir":         "/var/lib/kubelet/volumeplugins",
+		"volume-plugin-dir":         path.Join(prefixPath, "/var/lib/kubelet/volumeplugins"),
 		"fail-swap-on":              strconv.FormatBool(c.Services.Kubelet.FailSwapOn),
 		"root-dir":                  path.Join(prefixPath, "/var/lib/kubelet"),
 	}


chasebolt:
is there any changes needed to get this merged in? it currently is not possible to use rook/ceph on coreos.

HighwayofLife:
@chasebolt I'm running rook-ceph on CoreOS.
You can use this in your cluster.yml file to give vlex volume plugin access:
```
services:
  kubelet:
    extra_binds:
      - /var/lib/kubelet/volumeplugins:/var/lib/kubelet/volumeplugins
```

chasebolt:
thanks! right now im just running rke built from this branch.

im surprised `extra_args` actually works as it would define `--volume-plugin-dir` twice on the kubelet args. kubelet may just be ignoring the first one defined by rancher defaults.

which one are you using? because from my understanding the only one that would work would be the `extra_binds` with the correct prefix path on the host side.

```
services
  kubelet:
    extra_binds:
      - /opt/rke/var/lib/kubelet/volumeplugins:/var/lib/kubelet/volumeplugins
```

HighwayofLife:
`extra_args` overrides existing arguments, so that's how it avoids adding duplicates.

yasker:
Wait, what?

Why CoreOS doesn't work with Rook on `/var/lib/kubelet/volumeplugins`?

Here are the Rook documents, which asked for `/var/lib/kubelet/volumeplugins`

https://github.com/rook/rook/blob/master/Documentation/flexvolume.md

Adding another prefix result in user need to change Flexvolume directory based on the distro they're running.

@chasebolt @HighwayofLife @galal-hussein @alena1108

galal-hussein:
@yasker the prefix path was added to make sure its persistent on ROS and b2d, the prefix path was used for CentOS to avoid Selinux errors that happened with some users: https://github.com/rancher/rke/issues/598

yasker:
I don't know about Rook, but Longhorn doesn't need persistency for that directory. It should be installed once the host reboots.

chasebolt:
@HighwayofLife you are changing the pathing one way or another. you are changing it with your `extra_binds`, which really seems wrong since that is defining that path binding twice for kubelet and it just so happens that the 2nd define of it overlays on the first. if you look at the kubelet docker container with inspect you will see what i mean.

defining the correct volume plugins path in Rook imo is the correct way, not by doing an `extra_binds` on the kubelet. maybe there is a way we can make it all working without changing of any configs, kubelet or rook?

yasker:
@chasebolt 

The main issue with this PR is if the user is using RKE with RancherOS or CoreOS, they have to use `/opt/rke/var/lib/kubelet/volumeplugins` as the volume plugin path. But if they're using RKE with other OS, they have to use `/var/lib/kubelet/volumeplugins` instead. It's confusing already for the user have to set the volume plugin path manually (thanks to the K8S removed the API to detect the location), now it's more confusing that one distro will have different volume plugin path for the different host OS.

And the directory `/var/lib/kubelet/volumeplugins` shouldn't need persistence. The volume plugin should automatically clean up the directory when exits (e.g. host reboot), and reinstall it when it's started. Otherwise, the volume plugin will just leave junks on the host and cannot be uninstalled.

RKE should support Flexvolume by default without any user-side configuration (a.k.a no extra bind mount).

We can just bind mount `/var/lib/kubelet/volumeplugins` on the host to the container since persistence is not needed.

chasebolt:
@yasker @HighwayofLife 

#800 reverts the prefix pathing and adds the `extra_bind` that @HighwayofLife was doing manually. this should just make everything work out of the box. i think this should work for everyone, lmk if it needs an additional adjustment.

PR #714, opened 2018-06-22
Switching to docs site

deniseschannon:


chasebolt:
FWIW some of the comments from the `cluster.yaml` were super useful when i started out using RKE. hopefully these will be documented else where.

deniseschannon:
@chasebolt Yes, I've left all the examples from the readme. The docs site took all the content from here and put it in more detailed and organized format.

https://rancher.com/docs/rke/v0.1.x/en/

Issue #606, opened 2018-05-14
Enable API bearer tokens (including service account tokens) to be used to authenticate to the kubelet’s HTTPS endpoint


guangbochen:
**RKE version:** 0.1.7
This is a feature request:

Prometheus operator requires service account tokens authentication of `kubelets` to be enabled for it own and cadvisor metrics, can we please add this config to rke with enabled on default and optional to turn off by cluster admin.
```
--authentication-token-webhook=true
--authorization-mode=Webhook
```

Reference: 
https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet-authentication-authorization/

https://github.com/coreos/prometheus-operator/issues/976

cjellick:
To accomplish this, the k8s authentication webhook framework would need an endpoint to call back to. We would need to expose the Rancher authentication proxy as an an endpoint capable of functioning as one. I've opened the following issue in the rancher repo:
https://github.com/rancher/rancher/issues/13474

gvisca:
I'm working with version 0.1.6 of RKE and I managed to get kubelet metrics with prometheus operator with this kubelet config
```
 kubelet:
   image: rancher/hyperkube:v1.10.1-rancher2
   extra_args:
     authentication-token-webhook: true
```
adding authorization-mode=Webhook in extra-args prevent rke to deploy the cluster

cjellick:
I don't really understand what you did here, @galal-hussein @moelsayed @ibuildthecloud do you any of you understand what they did and how it allow the prom operator to work?

clkao:
It seems `--authorization-mode=Webhook` would make the installer fail, as `nodes/proxy` is not allowed from kube-apiserver and system:nodes. The installer will stop before properly configuring cni.

If you initialize the cluster first and then add  `--authorization-mode=Webhook` to cluster.yml after creating the proper clusterrole and clusterrolebinding, running `rke up` again configures kubelet properly. Unfortunately adding the role manifest to `addons_include` doesn't work, probably because they are created too late. perhaps we need something like `bootstrap_addons`.

jiuchongxiao:
did you solved it?@guangbochen

jiuchongxiao:
@gvisca  how to solve the prevent ?

wangwj:
rke0.1.7 


![Uploading image.png…]()
[...]

PR #715, opened 2018-06-22
enable bearer tokens authentication on default for kubelet

guangbochen:
#606


HighwayofLife:
You can use the extra_args parameter for the kubelet in the cluster.yml file.

Also when modifying a go file please run go fmt

galal-hussein:
I agree with @HighwayofLife, i don't think this is needed since we already use X509 client certificate authentication for kubelet.

pdhung:
I agree with @HighwayofLife and @galal-hussein. Not only that we already have x509 cert authentication + extra_args, but also this PR poses a potential security problem: enabling tokens authentication alone without proper "--authorization-mode=Webhook" settings means that every service account can now connect to kubelet and control it.

HighwayofLife:
@pdhung Is there a doc or blob post about that somewhere? Would be great to have it linked here for further reading.

pdhung:
@HighwayofLife: here it is https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet-authentication-authorization/
"Any request that is successfully authenticated (including an anonymous request) is then authorized. The default authorization mode is AlwaysAllow, which allows all requests.", which make service accounts able to interact with kubelet just as the API server

Flag "--authorization-mode=Webhook" is needed to subdivide and authorize the request:
"To subdivide access to the kubelet API, delegate authorization to the API server:
- ensure the authorization.k8s.io/v1beta1 API group is enabled in the API server
- start the kubelet with the --authorization-mode=Webhook and the --kubeconfig flags
- the kubelet calls the SubjectAccessReview API on the configured API server to determine whether each request is authorized"

PR #720, opened 2018-06-23
Log version

superseb:
https://github.com/rancher/rke/issues/721

### DIFF
--- a/cmd/remove.go
+++ b/cmd/remove.go
@@ -74,6 +74,10 @@ func ClusterRemove(
 }
 
 func clusterRemoveFromCli(ctx *cli.Context) error {
+	clusterFile, filePath, err := resolveClusterFile(ctx)
+	if err != nil {
+		return fmt.Errorf("Failed to resolve cluster file: %v", err)
+	}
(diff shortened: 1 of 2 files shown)


cmd/remove.go
```diff
@@ -74,6 +74,10 @@ func ClusterRemove(
 }
 
 func clusterRemoveFromCli(ctx *cli.Context) error {
+	clusterFile, filePath, err := resolveClusterFile(ctx)
```
galal-hussein: Why exactly did you move this to the top of the file, the local remove actually resolve the cluster file as well

superseb: It's mentioned in the linked issue:

```
Pre fix (on up you get an error when cluster.yml does not exist, on remove you don't)

$ ./build/bin/rke_darwin-amd64 --debug up
FATA[0000] Failed to resolve cluster file: Can not find cluster configuration file: open xx/rke/cluster.yml: no such file or directory
$ ./build/bin/rke_darwin-amd64 --debug remove
Are you sure you want to remove Kubernetes cluster [y/n]: n
```

PR #731, opened 2018-06-26
Add support for API aggregation layer

orangedeng:
Add API server parameters
 `requestheader-client-ca-file`,
`requestheader-allowed-names`,
`proxy-client-key-file` and
`proxy-client-cert-file`.
After we added this parameters into API server, the API aggregation
server will be supported when it is deployed in rke cluster.


orangedeng:
Here is the issue for this pr. https://github.com/rancher/rancher/issues/14231

galal-hussein:
@orangedeng is the CA cert used to configure the aggregation layer is that same that the user should use to configure the api extension service? if yes then how the user will obtain this?

orangedeng:
@galal-hussein If we add this parameter `requestheader-client-ca-file` into API server, k8s will set the ca cert into configmap `extension-apiserver-authentication` in `kube-system` with key `requestheader-client-ca-file`. I will assume that only cluster admin can deploy the API extension service so he has access to it. And as the API extension server design, it will get the configmap `extension-apiserver-authentication` when it starts. refer to [docs](https://github.com/kubernetes-incubator/apiserver-builder/blob/master/docs/concepts/auth.md#client-certificate-authentication). Same mechanism as `client-ca-file`.

galal-hussein:
@orangedeng one last concern is upgrades, when user upgrade the cluster from an older version where certificate already generated, kubeapi will enter an error loop and will keep crashing because it will not be able to find the certificates mentioned.

orangedeng:
@galal-hussein Updated. Should be ok for upgrading.

PR #760, opened 2018-07-07
Set permissions on cloud-config to 644

HighwayofLife:
Resources using the cloud-config get a permission error as the default permission is 600.
```
/etc/kubernetes/cloud-config (Permission denied)
```


superseb:
The path was corrected in v0.1.8, @galal-hussein worked on this refactor AFAIK.

HighwayofLife:
@superseb are you referring to `CloudConfigPath        = "/etc/kubernetes/cloud-config.json"` having a .json extension, or something else?

cluster/cloud-provider.go
```diff
@@ -16,7 +16,7 @@ import (
 const (
 	CloudConfigDeployer    = "cloud-config-deployer"
 	CloudConfigServiceName = "cloud"
-	CloudConfigPath        = "/etc/kubernetes/cloud-config"
+	CloudConfigPath        = "/etc/kubernetes/cloud-config.json"
```
galal-hussein: We are not using json for each cloud provider now, for vsphere and openstack we use .gcfg extension, so its better to remove the extension altogether

HighwayofLife:
@galal-hussein reverted the extension portion of the change and rebased.

PR #763, opened 2018-07-10
Add dind mode to rke

galal-hussein:
https://github.com/rancher/rancher/issues/14266

The dind mode will enable the user to deploy nodes as docker containers locally, each docker container has docker installed within, for example to create 3 node cluster in dind mode:
cluster.yml
```
nodes:
  - address: 172.18.0.9
    user: ubuntu
    role: [controlplane,worker,etcd]
  - address: 172.18.0.10
    user: ubuntu
    role: [controlplane,worker,etcd]
  - address: 172.18.0.11
    user: ubuntu
    role: [controlplane,worker,etcd]
```
and then:
```
rke up --dind
```
RKE will first create a network subnet called `dind-network` with subnet defaults to `172.18.0.0/16` and then deploy 3 nodes as docker containers and start kubernetes installation

### DIFF
--- a/cmd/remove.go
+++ b/cmd/remove.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/rancher/rke/cluster"
+	"github.com/rancher/rke/dind"
 	"github.com/rancher/rke/hosts"
 	"github.com/rancher/rke/k8s"
 	"github.com/rancher/rke/log"
@@ -31,7 +32,11 @@ func RemoveCommand() cli.Command {
 		},
 		cli.BoolFlag{
 			Name:  "local",
-			Usage: "Deploy Kubernetes cluster locally",
+			Usage: "Remove Kubernetes cluster locally",
+		},
+		cli.BoolFlag{
+			Name:  "dind",
+			Usage: "Remove Kubernetes cluster deployed in dind mode",
 		},
 	}
 
@@ -94,6 +99,9 @@ func clusterRemoveFromCli(ctx *cli.Context) error {
 	if ctx.Bool("local") {
 		return clusterRemoveLocal(ctx)
 	}
+	if ctx.Bool("dind") {
+		return clusterRemoveDind(ctx)
+	}
 	clusterFilePath = filePath
 	rkeConfig, err := cluster.ParseConfig(clusterFile)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


cmd/remove.go
```diff
@@ -33,6 +35,10 @@ func RemoveCommand() cli.Command {
 			Name:  "local",
 			Usage: "Deploy Kubernetes cluster locally",
 		},
+		cli.BoolFlag{
+			Name:  "dind",
+			Usage: "Remove Kubernetes cluster deployed in dind mode",
```
alena1108: flag name "dind" doesn't really correlate to it's usage - it is used to remove the cluster?

alena1108: can't we figure out in which mode the cluster was deployed, w/o passing the flag? ideally the flag should be passed on create only

cmd/remove.go
```diff
@@ -33,6 +35,10 @@ func RemoveCommand() cli.Command {
 			Name:  "local",
 			Usage: "Deploy Kubernetes cluster locally",
```
alena1108: I know it's not a part of this pr, but should usage say "Remove" instead of "Deploy"?

cmd/remove.go
```diff
@@ -130,3 +139,31 @@ func clusterRemoveLocal(ctx *cli.Context) error {
...
+	rkeConfig, err = setOptionsFromCLI(ctx, rkeConfig)
+	if err != nil {
+		return err
+	}
+
+	for _, node := range rkeConfig.Nodes {
+		if err = dind.RmoveDindContainer(context.Background(), node.Address); err != nil {
+			return err
+		}
+		time.Sleep(DINDWaitTime * time.Second)
```
alena1108: why do we need to sleep here? should we validate smth after sleep is done?

cmd/up.go
```diff
@@ -205,3 +221,46 @@ func clusterUpLocal(ctx *cli.Context) error {
...
+	if err != nil {
+		return fmt.Errorf("Failed to parse cluster file: %v", err)
+	}
+
+	rkeConfig, err = setOptionsFromCLI(ctx, rkeConfig)
+	if err != nil {
+		return err
+	}
+
+	disablePortCheck := ctx.Bool("disable-port-check")
```
alena1108: do you think it would make sense to extract config setup + dind setup to it's own methods? So clusterupdind would look something like:

* read/setupConfig
* deploy dind
* cluster up

cmd/remove.go
galal-hussein: sure will fix this

cmd/remove.go
galal-hussein: oh that's an error the sleep was supposed to be after creation to give time to the container to be ready, will fix

cmd/remove.go
galal-hussein: there is no way to identify the mode of rke installation of the nodes, each mode has different dialer and healthcheck dialer, that's why we use:
```
rke up --dind
rke remove --dind
```
and for local
```
rke up --local
rke remove --local
```

cmd/up.go
galal-hussein: sure will fix that

PR #774, opened 2018-07-13
Remove ingress controller when disabled

moelsayed:


addons/addons.go
```diff
@@ -1,12 +1,42 @@
...
+func GetAddonsDeleteJob(addonName, nodeName, image string) (string, error) {
+	return getAddonJob(addonName, nodeName, image, true)
+}
+
+func getAddonJob(addonName, nodeName, image string, isDelete bool) (string, error) {
 	jobConfig := map[string]string{
 		"AddonName": addonName,
 		"NodeName":  nodeName,
 		"Image":     image,
+		"DeleteJob": "false",
```
galal-hussein: instead of the if condition you can use
```
"DeleteJob": strconv.FormatBool(isDelete),
```

cluster/addons.go
```diff
@@ -226,6 +229,44 @@ func (c *Cluster) doAddonDeploy(ctx context.Context, addonYaml, resourceName str
...
+	}
+	deleteJob, err := addons.GetAddonsDeleteJob(resourceName, node.Name, c.Services.KubeAPI.Image)
+	if err != nil {
+		return &addonError{fmt.Sprintf("Failed to generate addon delete job: %v", err), isCritical}
+	}
+	if err := k8s.ApplyK8sSystemJob(deleteJob, c.LocalKubeConfigPath, c.K8sWrapTransport, c.AddonJobTimeout, false); err != nil {
+		return &addonError{fmt.Sprintf("%v", err), isCritical}
+	}
+	// At this point, the addon should be deleted. We need to clean up by deleting the deploy and delete jobs.
+	//fmt.Println("melsayed", err)
```
galal-hussein: can you remove this comment

cluster/addons.go
```diff
@@ -226,6 +229,44 @@ func (c *Cluster) doAddonDeploy(ctx context.Context, addonYaml, resourceName str
...
+	tmpJobYaml, err := addons.GetAddonsExcuteJob(resourceName, node.Name, c.Services.KubeAPI.Image)
+	if err != nil {
+		return err
+	}
+	if err := k8s.DeleteK8sSystemJob(tmpJobYaml, k8sClient, c.AddonJobTimeout); err != nil {
+		return err
+	}
+
+	tmpJobYaml, err = addons.GetAddonsDeleteJob(resourceName, node.Name, c.Services.KubeAPI.Image)
```
galal-hussein: this is redundant since you already queried it

cluster/addons.go
```diff
@@ -226,6 +229,39 @@ func (c *Cluster) doAddonDeploy(ctx context.Context, addonYaml, resourceName str
...
+	// At this point, the addon should be deleted. We need to clean up by deleting the deploy and delete jobs.
+	tmpJobYaml, err := addons.GetAddonsExecuteJob(resourceName, node.Name, c.Services.KubeAPI.Image)
+	if err != nil {
+		return err
+	}
+	if err := k8s.DeleteK8sSystemJob(tmpJobYaml, k8sClient, c.AddonJobTimeout); err != nil {
+		return err
+	}
+
+	if err := k8s.DeleteK8sSystemJob(deleteJob, k8sClient, c.AddonJobTimeout); err != nil {
```
alena1108: curious why do we call the same method twice, with diff param?

moelsayed: because we have a k8s job to deploy the addon, and another one to delete it.. when an addon is removed, we delete both jobs to avoid conflict in the future if we want to enable it again..

Issue #768, opened 2018-07-11
Failed to restore etcd snapshot: Failed to run etcd restore container, exit status is: 3


techcto:
rke version v0.1.8

Docker Version:      18.03.1-ce
 API version:  1.37
 Go version:   go1.9.6

4.14.47-64.38.amzn2.x86_64
NAME="Amazon Linux"
VERSION="2"
ID="amzn"
ID_LIKE="centos rhel fedora"
VERSION_ID="2"
PRETTY_NAME="Amazon Linux 2"
ANSI_COLOR="0;33"
CPE_NAME="cpe:2.3:o:amazon:amazon_linux:2"
HOME_URL="https://amazonlinux.com/"

https://s3.amazonaws.com/kub88-rke/config.yaml 

0. Start with 3 instances at AWS
1. ./rke up --config config.yaml (Please feel free to test with my config.  Just let me know so I can stay off. This is a test account, so no sensitive data.)
2. ./rke etcd snapshot-save --name etcdsnapshot --config config.yaml
3. Pick one server, copy etcdsnapshot onto other 2 instances
4. ./rke etcd snapshot-restore --name etcdsnapshot --config config.yaml

Failure:

INFO[0010] [etcd] Restoring [etcdsnapshot] snapshot on etcd host [54.175.184.90]
INFO[0010] [etcd] Successfully started [etcd-restore] container on host [54.175.184.90]
INFO[0011] [etcd] Restoring [etcdsnapshot] snapshot on etcd host [54.174.94.151]
INFO[0011] [etcd] Successfully started [etcd-restore] container on host [54.174.94.151]
FATA[0011] [etcd] Failed to restore etcd snapshot: Failed to run etcd restore container, exit status is: 3

It always stops on second/third server.  I am not sure where the error is.  But this is is a simple backup and restore on 3 existing instances.  Please let me know if I am missing something.

galal-hussein:
can you check the etcd-restore logs on the failed instance:
```
docker logs etcd-restore
```

techcto:
Not sure which server, but I am guessing the second server based on the output.  This is what I get from server 2:

Error:  data-dir "/opt/rke/etcd-snapshots-restore/" exists
Error:  data-dir "/opt/rke/etcd-snapshots-restore/" exists
Error:  data-dir "/opt/rke/etcd-snapshots-restore/" exists
Error:  data-dir "/opt/rke/etcd-snapshots-restore/" exists
[...]

PR #786, opened 2018-07-18
Add etcd snapshot fix and more log messages to certificate bundle

galal-hussein:
https://github.com/rancher/rke/issues/768

### DIFF
--- a/cluster/network.go
+++ b/cluster/network.go
@@ -1,14 +1,12 @@
 package cluster
 
 import (
-	"bytes"
 	"context"
 	"fmt"
 	"net"
 	"strings"
 
 	"github.com/docker/docker/api/types/container"
-	"github.com/docker/docker/pkg/stdcopy"
 	"github.com/docker/go-connections/nat"
 	"github.com/rancher/rke/docker"
 	"github.com/rancher/rke/hosts"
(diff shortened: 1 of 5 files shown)


cluster/network.go
```diff
@@ -435,14 +431,10 @@ func checkPlaneTCPPortsFromHost(ctx context.Context, host *hosts.Host, portList
 		return err
 	}
 
-	clogs, err := docker.ReadContainerLogs(ctx, host.DClient, PortCheckContainer, true, "all")
-	if err != nil {
-		return err
+	containerLog, logsErr := docker.GetContainerLogsStdoutStderr(ctx, host.DClient, PortCheckContainer, "1")
+	if logsErr != nil {
+		logrus.Debugf("[network] Failed to get network port check logs: %v", logsErr)
```
alena1108: should we log it in warn? you won't be able to see it on user's setup where the log level is info

docker/docker.go
```diff
@@ -362,6 +364,22 @@ func ReadContainerLogs(ctx context.Context, dClient *client.Client, containerNam
...
 }
 
+func GetContainerLogsStdoutStderr(ctx context.Context, dClient *client.Client, containerName, tail string) (string, error) {
+	var containerStderr bytes.Buffer
+	var containerStdout bytes.Buffer
+	var containerLog string
+	clogs, logserr := ReadContainerLogs(ctx, dClient, containerName, false, tail)
+	if logserr != nil {
+		logrus.Debug("logserr: %v", logserr)
+		return containerLog, fmt.Errorf("Failed to get gather logs from container [%s]", containerName)
```
alena1108: it might be better to return the error in the error message. The current error doesn't reflect any useful info about the error itself

PR #799, opened 2018-07-20
Revert adjust kubelet volume-plugin-dir to match root-dir

alena1108:
Reverts https://github.com/rancher/rke/pull/709

@yasker could you put the detailed reason for review here


alena1108:
Closing in favor of https://github.com/rancher/rke/pull/800

Issue #59, opened 2017-11-28
Better error message and directions on next steps when control plane node is not reachable.


sangeethah:
Tested with rke version v0.0.5-dev

Steps to reproduce the problem:
1. Create a k8s cluster with 1 node each in controlplane,etcd and worker.
2. Add another node to controlplane.
Power down the host in controlplane from step 1 , which is the api-server configured in .kube_config_cluster.yml

Using cluster up command results in ```Error establishing SSH connection```.
Can this error message be improved and include suggestion on what can be done to get around this state and allow for connecting to the cluster?

```
Sangeethas-MBP:~ sangeethahariharan1$ ./rke_darwin-amd64 cluster up
INFO[0000] Building Kubernetes cluster                  
INFO[0000] [ssh] Checking private key                   
INFO[0000] [ssh] Start tunnel for host [ip-***.us-east-2.compute.internal] 
INFO[0000] [ssh] Start tunnel for host [ip-***.us-east-2.compute.internal] 
INFO[0000] [ssh] Start tunnel for host [ip-***.us-east-2.compute.internal] 
INFO[0000] [ssh] Start tunnel for host [ip-***.us-east-2.compute.internal] 
INFO[0000] [ssh] Start tunnel for host [ip-***.us-east-2.compute.internal] 
INFO[0000] [ssh] Start tunnel for host [ip-***.us-east-2.compute.internal] 
INFO[0000] [state] Found local kube config file, trying to get state from cluster 
INFO[0000] [state] Fetching cluster state from Kubernetes 
INFO[0030] Timed out waiting for kubernetes cluster to get state 
INFO[0030] [certificates] Generating kubernetes certificates 
INFO[0030] [certificates] Generating CA kubernetes certificates 
INFO[0030] [certificates] Generating Kubernetes API server certificates 
INFO[0031] [certificates] Generating Kube Controller certificates 
INFO[0031] [certificates] Generating Kube Scheduler certificates 
INFO[0032] [certificates] Generating Kube Proxy certificates 
INFO[0034] [certificates] Generating Node certificate   
INFO[0034] [certificates] Generating admin certificates and kubeconfig 
INFO[0036] [certificates] Deploying kubernetes certificates to Cluster nodes 
[...]

PR #820, opened 2018-07-27
Add better error when node is unreachable

superseb:
https://github.com/rancher/rke/issues/59


superseb:
This was an uncaught error from https://github.com/rancher/rke/issues/59, can you check @galal-hussein @moelsayed

Issue #821, opened 2018-07-27
Snapshot Save Unable to Archive Certs to /opt/rke/etcd-snapshots/


massimo-cannavo:
**RKE version:** v0.1.9-rc3

**Docker version: (`docker version`,`docker info` preferred)** 18.03.1-ce

**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)** RancherOS v1.4.0

**cluster.yml file:**

```
ignore_docker_version: true

nodes:
  - address: 159.65.102.198
    user: rancher
    role:
      - controlplane
      - etcd
      - worker
    ssh_key_path: ~/.ssh/id_rsa

  - address: 159.65.107.8
    user: rancher
    role:
      - controlplane
      - worker
    ssh_key_path: ~/.ssh/id_rsa

services:
  etcd:
    snapshot: true # enables recurring etcd snapshots
    creation: 1440m # time increment between shapshots
    retention: 168h # time increment before snapshot purge

addons: |-
  ---
  kind: Namespace
  apiVersion: v1
  metadata:
    name: cattle-system

  ---
  kind: ServiceAccount
  apiVersion: v1
  metadata:
    name: cattle-admin
    namespace: cattle-system

  ---
  kind: ClusterRoleBinding
  apiVersion: rbac.authorization.k8s.io/v1
  metadata:
    name: cattle-crb
    namespace: cattle-system
  subjects:
  - kind: ServiceAccount
    name: cattle-admin
    namespace: cattle-system
  roleRef:
    kind: ClusterRole
    name: cluster-admin
    apiGroup: rbac.authorization.k8s.io

  ---
  apiVersion: v1
  kind: Secret
  metadata:
    name: cattle-keys-ingress
    namespace: cattle-system
  type: Opaque
  data:
    tls.crt: 
    tls.key: 

  ---
  apiVersion: v1
  kind: Service
  metadata:
    namespace: cattle-system
    name: cattle-service
    labels:
      app: cattle
  spec:
    ports:
    - port: 80
      targetPort: 80
      protocol: TCP
      name: http
    - port: 443
      targetPort: 443
      protocol: TCP
      name: https
    selector:
      app: cattle

  ---
  apiVersion: extensions/v1beta1
  kind: Ingress
  metadata:
    namespace: cattle-system
    name: cattle-ingress-http
    annotations:
      nginx.ingress.kubernetes.io/proxy-connect-timeout: "30"
[...]

PR #822, opened 2018-07-27
Fixed snapshot restore unable to archive pki bundle

massimo-cannavo:
Refer to issue #821


galal-hussein:
@massimo-cannavo can you also add a fix for extract, it has the same problem, thanks!

massimo-cannavo:
@galal-hussein I believe it should be good to go.

PR #831, opened 2018-08-01
fixes wrong value for soap-roundtrip-count in vsphere provider

iljaweis:
The template for creating the cloud provider config file for vsphere uses a wrong value for soap-roundtrip-count in the VirtualCenter section.
This commit fixes that. It will set the value to 0 if not specified in RKE's cluster config. That should be ok as a value of 0 means "use the default" and the default is the value set in the global section.


iljaweis:
Thanks!

Issue #650, opened 2018-05-23
/opt/rke/var/lib/kubelet mounted thousands of times


HighwayofLife:
This line appears in a `df -h` command over a thousand times:
```
/dev/sda9        494G  2.1G  472G   1% /opt/rke/var/lib/kubelet
```
This is massively slowing down each node and we're encountering problems with the kubelet not passing healthcheck.

RKE version v0.1.7

HighwayofLife:
Update: Once I upgraded from kubernetes_version `v1.10.1-rancher2-1` to `v1.10.1-rancher2-1`, rke-tools stopped remounting the `/opt/rke/var/lib/kubelet` This *might* have also been causing #648 once there was so many mounts, the host became largely unresponsive.

HighwayofLife:
Yes. I hit it again multiple times and again today. This only happens to
existing clusters, and I think the reason might be if something causes the
node to drop off line ( kubelet no longer reporting). I haven't figured out
whether the solution is to Simply restart the node, or to reprovision it
entirely.

>

moelsayed:
@HighwayofLife can you please share kubelet logs ? and debug logs from rke runs ? Which RC did you use ?

HighwayofLife:
I am able to reproduce this with some regularity now, so working on gathering some useful info.

I created / provisioned a series of nodes (6) in Azure, CoreOS Virtual Machines. Rake rke, which runs successfully, then when I update some components, maybe updating the kubelet, the kubelet itself is restarting multiple times, and there are over a thousand mounts to /var/lib/kubelet on at least one of the worker nodes.

```
/dev/sda9        246G  3.3G  233G   2% /opt/rke/var/lib/kubelet

$ df -h | wc -l
1038
```

Kubelet logs --tail 500:
https://gist.github.com/HighwayofLife/93e81a4eb659e35891931a061c7ca74e

RKE error:
```
[...]

PR #836, opened 2018-08-02
Fix sidekick upgrade issue and change services entry point

moelsayed:
https://github.com/rancher/rke/issues/650


rossedman:
@moelsayed when I build this using dapper the binary doesn't work for me. could be something i'm doing wrong. any ideas?

rossedman:
@moelsayed this currently doesn't work for me. started with fresh nodes. the `/opt/rke-tools` folder is not created on host causing failures.

HighwayofLife:
Tested and works on a brand new cluster. Upgrading an existing cluster could be an issue though, we may need to test that.

Issue #810, opened 2018-07-24
Preparing IPVS mode in Calico


sebastien-prudhomme:
**RKE version:**

v0.1.9-rc2

Hi, as IPVS proxy mode is now GA in Kubernetes 1.11, and will be the default mode in future version, i think it's a good idea to prepare RKE to work in this mode.

One on the step is to make sure Calico (and perhaps Canal ?) uses the same node ports configured in Kubernetes:

https://docs.projectcalico.org/v3.1/reference/felix/configuration#ipvs-portranges

As this range is now configurable in RKE, see #672, Calico should use the same range througth the use of FELIX_KUBENODEPORTRANGES env variable.

PR #839, opened 2018-08-03
Improving IPVS mode compatibility #810

sebastien-prudhomme:


cluster/network.go
```diff
@@ -140,6 +141,7 @@ func (c *Cluster) doCalicoDeploy(ctx context.Context) error {
 	calicoConfig := map[string]string{
 		KubeCfg:       clientConfig,
 		ClusterCIDR:   c.ClusterCIDR,
+		NodePortRange: c.Services.KubeAPI.ServiceNodePortRange,
```
galal-hussein: what is reason behind configuring the node port range?

cluster/plan.go
```diff
@@ -437,6 +437,7 @@ func (c *Cluster) BuildKubeProxyProcess(host *hosts.Host, prefixPath string) v3.
 	}
 	Binds := []string{
 		fmt.Sprintf("%s:/etc/kubernetes:z", path.Join(prefixPath, "/etc/kubernetes")),
+		"/lib/modules:/lib/modules:ro",
```
galal-hussein: is it safe to mount the /lib/modules 
CC @moelsayed @yasker

yasker: It should be safe (and it's RO) but I don't know the reason behind this. Why does this container need `/lib/modules/`? Is it going to load kernel module inside the container? If so, this not a good idea in general since it would change the host OS behavior as well.

cluster/network.go
sebastien-prudhomme: In the Calico doc: https://docs.projectcalico.org/v3.1/usage/enabling-ipvs
"Calico needs to be configured with the port range that is assigned to Kubernetes NodePorts. If services do use NodePorts outside Calico’s expected range, Calico will treat traffic to those ports as host traffic instead of pod traffic"

cluster/plan.go
sebastien-prudhomme: Yes, kube-proxy tries to load modules in the container if the modules are not yet loaded.

Search there for modprobe: https://github.com/kubernetes/kubernetes/blob/master/pkg/proxy/ipvs/proxier.go

yasker: I haven't dug deep into this. But what impact it would have on the host os?

willmao:
Any update on this?

willmao:
if we do not mount /lib/modules，kube-proxy will try to load it and print the following warnings:

E0401 14:52:42.399536    9105 proxier.go:473] Failed to read file /lib/modules/4.4.0-142-generic/modules.builtin with error open /lib/modules/4.4.0-142-generic/modules.builtin: no such file or directory
W0401 14:52:42.400774    9105 proxier.go:486] Failed to load kernel module ip_vs with modprobe. You can ignore this message when kube-proxy is running inside container without mounting /lib/modules
W0401 14:52:42.402008    9105 proxier.go:486] Failed to load kernel module ip_vs_rr with modprobe. You can ignore this message when kube-proxy is running inside container without mounting /lib/modules
W0401 14:52:42.403109    9105 proxier.go:486] Failed to load kernel module ip_vs_wrr with modprobe. You can ignore this message when kube-proxy is running inside container without mounting /lib/modules
W0401 14:52:42.404108    9105 proxier.go:486] Failed to load kernel module ip_vs_sh with modprobe. You can ignore this message when kube-proxy is running inside container without mounting /lib/modules
W0401 14:52:42.405196    9105 proxier.go:486] Failed to load kernel module nf_conntrack_ipv4 with modprobe. You can ignore this message when kube-proxy is running inside container without mounting /lib/modules
W0401 14:52:42.406156    9105 proxier.go:486] Failed to load kernel module nf_conntrack with modprobe. You can ignore this message when kube-proxy is running inside container without mounting /lib/modules
time="2019-04-01T14:52:42Z" level=warning msg="Running modprobe ip_vs failed with message: `modprobe: ERROR: ../libkmod/libkmod.c:586 kmod_search_moddep() could not open moddep file '/lib/modules/4.4.0-142-generic/modules.dep.bin'\nmodprobe: WARNING: Module ip_vs not found in directory /lib/modules/4.4.0-142-generic`, error: exit status 1"

PR #842, opened 2018-08-04
Handle entrypoint upgrade across versions

moelsayed:
Note: For this to work in rancher we need to add the semver package to vendor.conf.


galal-hussein:
https://github.com/rancher/rke/issues/650

Issue #266, opened 2018-01-25
Do not have put the generated CA Private Key on Every Node


ghost:
**RKE version:**
0.0.12-dev

**Docker version: (`docker version`,`docker info` preferred)**
17.06.2-ce

**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)**
RancherOS v1.1.3

**Type/provider of hosts: (VirtualBox/Bare-metal/AWS/GCE/DO)**
Baremetal (KVM)

**cluster.yml file:**
Not required

**Steps to Reproduce:**
* Run RKE on any cluster.yml configuration
* SSH into any node, controlpane or worker

**Results:**
* Generated CA private key is stored on every node
* Other kube services certificates are very generic, issued as a blanket certificate for every controlpane and/or node

**Expected Result:**
* Sensitive secrets should be exported after the provisioning, similar to tectonic installer, and not stored on every node.
* Other kube services certificates should be issued for each individual node/apiserver/controller/scheduler/proxy combination, so it cannot be reused on another node

sebastien-prudhomme:
The other security issue i see is that kubelet, kube-apiserver, etc all use a mount bind for /etc/kubernetes directory which has all the certificates and keys.

Each certificate or key should be mounted individualy.

Will try to propose a PR for that.

PR #848, opened 2018-08-08
Improve private keys security #266

sebastien-prudhomme:
Also change the client certificate used to contact Kubelets. It's better to use the node certificate than the API server certificate.

### DIFF
--- a/cluster/plan.go
+++ b/cluster/plan.go
@@ -126,8 +126,8 @@ func (c *Cluster) BuildKubeAPIProcess(prefixPath string) v3.Process {
 		"client-ca-file":                     pki.GetCertPath(pki.CACertName),
 		"tls-cert-file":                      pki.GetCertPath(pki.KubeAPICertName),
 		"tls-private-key-file":               pki.GetKeyPath(pki.KubeAPICertName),
-		"kubelet-client-certificate":         pki.GetCertPath(pki.KubeAPICertName),
-		"kubelet-client-key":                 pki.GetKeyPath(pki.KubeAPICertName),
+		"kubelet-client-certificate":         pki.GetCertPath(pki.KubeNodeCertName),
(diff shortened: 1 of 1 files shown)


cluster/plan.go
```diff
@@ -126,8 +126,8 @@ func (c *Cluster) BuildKubeAPIProcess(prefixPath string) v3.Process {
 		"client-ca-file":                     pki.GetCertPath(pki.CACertName),
 		"tls-cert-file":                      pki.GetCertPath(pki.KubeAPICertName),
 		"tls-private-key-file":               pki.GetKeyPath(pki.KubeAPICertName),
-		"kubelet-client-certificate":         pki.GetCertPath(pki.KubeAPICertName),
-		"kubelet-client-key":                 pki.GetKeyPath(pki.KubeAPICertName),
+		"kubelet-client-certificate":         pki.GetCertPath(pki.KubeNodeCertName),
```
galal-hussein: we use kubeapi cert because its a authentication from kubeapi to kubelet, not the other way around

cluster/plan.go
```diff
@@ -174,7 +174,28 @@ func (c *Cluster) BuildKubeAPIProcess(prefixPath string) v3.Process {
 		services.SidekickContainerName,
 	}
 	Binds := []string{
-		fmt.Sprintf("%s:/etc/kubernetes:z", path.Join(prefixPath, "/etc/kubernetes")),
+		fmt.Sprintf("%s:%s:ro", path.Join(prefixPath, pki.GetCertPath(pki.CACertName)), pki.GetCertPath(pki.CACertName)),
```
galal-hussein: can you explain how this will improve the security of managing certificates, this solution will also break the ability to use custom certificates deployed in /etc/kubernetes

PR #853, opened 2018-08-08
Handled non-semver rke-tools image versions

moelsayed:


moelsayed:
Closed in favor of #856

Issue #450, opened 2018-03-28
RFE - Ease certificate rotation in kubernetes


aemneina:
Add the ability to rotate certificates out within kubernetes:
- user should be able to rotate the certificates using rke command `rke cert rotate` or similar

- user should be able to provide certificates as an option - Removed this request to a different issue: https://github.com/rancher/rke/issues/1037

zedtux:
Any progress on this?

techninja1008:
Any Progress?

matthax:
Hey I've got a cluster with a bad SSL certificate and I'm not able to update the cert using `rke up`. Is there a short term workaround that I can use until this is solved?

cjellick:
Possible ui copoments for showing and managing certs for rke:
- upload on create
- renew and/or auto rotate

alena1108:
@galal-hussein also explore the possibility of using a different key for service account vs k8s components

deniseschannon:
@galal-hussein we should make sure this works for air gapped setups

mtchuyen:
my certificate cluster will expire soon. Hurry up pls!

bmdepesa:
First pass of testing has been completed.

PR #865, opened 2018-08-18
Add Rotate certificates option to rke

galal-hussein:
https://github.com/rancher/rke/issues/450

Certificate Rotation PR includes:

- Add `rke cert rotate` command for:
  - all certificates
  - ca certificates (Master CA - Request header CA) `--rotate-ca`
  - specific component (kubelet, kube-apiserver, kube-proxy, kube-controller-manager, kube-scheduler, etcd) eg. `--service etcd`
- Reconcile the service account token key and use different key instead of kubeapi key
 

Tested Scenarios:

- Run cert rotation for all certificates
- Run cert rotation for all certs and ca certs
- Run cert rotation for specific component
- Run cert rotation command on cluster created by rke v0.1.11
- Upgrade from rke v0.1.11 to master
- Run cert rotation multiple times

### DIFF
--- a/cluster/certificates.go
+++ b/cluster/certificates.go
@@ -71,7 +71,7 @@ func SetUpAuthentication(ctx context.Context, kubeCluster, currentCluster *Clust
 					if err := rebuildLocalAdminConfig(ctx, kubeCluster); err != nil {
 						return err
 					}
-					kubeCluster.Certificates, err = regenerateAPICertificate(kubeCluster, kubeCluster.Certificates)
+					err = pki.GenerateKubeAPICertificate(ctx, kubeCluster.Certificates, kubeCluster.RancherKubernetesEngineConfig, "", "")
 					if err != nil {
 						return fmt.Errorf("Failed to regenerate KubeAPI certificate %v", err)
 					}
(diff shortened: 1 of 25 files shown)


cluster/certificates.go
```diff
@@ -361,3 +364,54 @@ func regenerateAPIAggregationCerts(c *Cluster, certificates map[string]pki.Certi
...
+	}
+	for _, k8sComponent := range components {
+		genFunc := componentsCertsFuncMap[k8sComponent]
+		if genFunc != nil {
+			if err := genFunc(ctx, c.Certificates, c.RancherKubernetesEngineConfig, configPath, configDir); err != nil {
+				return err
+			}
+		}
+	}
+	if len(components) == 0 {
```
moelsayed: Should this be moved to 385 ? logic seems related?

services/controlplane.go
```diff
@@ -80,6 +80,56 @@ func RemoveControlPlane(ctx context.Context, controlHosts []*hosts.Host, force b
...
+					errList = append(errList, err)
+				}
+
+				// restart scheduler
+				err := restartScheduler(ctx, runHost)
+				if err != nil {
+					errList = append(errList, err)
+				}
+
+				// check if the host already is a worker
```
moelsayed: Do you think it would be simpler to just restart CP and Etcd components on their respective planes and then restart worker components on _all_ nodes?

cluster/certificates.go
galal-hussein: hmmm, i can't do that because the user may need to only restart all components without ca which is the default option

moelsayed: got it :+1:

PR #877, opened 2018-08-24
Fixed health check timing issue on slow platforms

daskanu:
When using RKE to deploy Kubernetes clusters on Azure, I often run into the issue that the health checks fail, because some containers (e.g. kube-apiserver, kubelet) take too long for coming up.

Therefore I increased the maximum amount of retries from 10 to 50 and this solved the problem for me.

I think the best solution would be to make this configurable using the cluster.yml so that the default value can stick to 10.

### DIFF
--- a/services/healthcheck.go
+++ b/services/healthcheck.go
@@ -53,7 +53,7 @@ func runHealthcheck(ctx context.Context, host *hosts.Host, serviceName string, l
 	if err != nil {
 		return fmt.Errorf("Failed to initiate new HTTP client for service [%s] for host [%s]: %v", serviceName, host.Address, err)
 	}
-	for retries := 0; retries < 10; retries++ {
+	for retries := 0; retries < 50; retries++ {
 		if err = getHealthz(client, serviceName, host.Address, url); err != nil {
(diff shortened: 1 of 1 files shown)


jomeier:
@daskanu:
Same problems here. Your fix helped us in getting rke running without timeouts!

alena1108:
@daskanu we should make the number of attempts configurable. 50 for default sounds too excessive. @galal-hussein @moelsayed could you point what is the best place for defining such config.

moelsayed:
@daskanu That's probably not the best way to fix this. I think it would be better if we used a configurable timeout like. Ideally you could use [this approach](https://github.com/rancher/rke/blob/8f47078988da92b57df6e8abe9c27a2ca6b74f43/cluster/state.go#L110-L128), and make the timeout configurable with a default like we do with addon_job_timeout:
- https://github.com/rancher/types/blob/2dcb95d094ec01fefa8a8736b4b74b09f389d559/apis/management.cattle.io/v3/rke_types.go#L38-L39
- https://github.com/rancher/types/blob/master/apis/management.cattle.io/v3/rke_types.go#L38-L39

Let me know if you have any questions.

Issue #588, opened 2018-05-07
Pre-Pull Pause images to be able to use private registries with basic auth


galal-hussein:
Using private registries with basic auth in rke should allow kubelet to use the pause image, one solution is to pre pull the image with docker credentials like we do in kubernetes images. also using ImagePullSecret will be necessary to be added to the templates

https://github.com/rancher/rancher/issues/13339

alena1108:
https://github.com/rancher/rancher/issues/13339

bmdepesa:
rancher/rancher:master 9/13
rke version v0.1.10-rc2

* Created private registry and tagged/pushed system images (`rke config --system-images`)
* Created cluster in Rancher, with edited yaml to include private registry
```
private_registries:
- url: <registry url>
  user: <password>
  password: <password>
  is_default: true
```
* The created cluster comes up successfully and is using images from the private registry
* Deployed workload w/ ingress (nginx:latest) - functions propertly

PR #882, opened 2018-08-28
[WIP] Add docker config file for kubelet

moelsayed:
https://github.com/rancher/rke/issues/588


moelsayed:
This solution will not work with rancher-agent in worker only mode. It will require the agent bind mount several directories depending on the node OS, which is information we don't have while generating the registration command.

Closing this in favor of: https://github.com/rancher/rke/pull/884

PR #886, opened 2018-08-29
Move admission controller options to ServiceOptions

moelsayed:
https://github.com/rancher/rancher/issues/15064
requires: https://github.com/rancher/types/pull/554


galal-hussein:
LGTM

PR #906, opened 2018-09-13
Save state to the nodes backup path

galal-hussein:
https://github.com/rancher/rancher/issues/15422

### DIFF
--- a/cluster/state.go
+++ b/cluster/state.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"path"
 	"time"
 
+	"github.com/rancher/rke/hosts"
 	"github.com/rancher/rke/k8s"
 	"github.com/rancher/rke/log"
 	"github.com/rancher/rke/pki"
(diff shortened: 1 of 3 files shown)


cluster/state.go
```diff
@@ -128,7 +149,19 @@ func saveStateToKubernetes(ctx context.Context, kubeClient *kubernetes.Clientset
...
 
-func getStateFromKubernetes(ctx context.Context, kubeClient *kubernetes.Clientset, kubeConfigPath string) *Cluster {
+func saveStateToNodes(ctx context.Context, uniqueHosts []*hosts.Host, clusterState *v3.RancherKubernetesEngineConfig, alpineImage string, prsMap map[string]v3.PrivateRegistry) error {
+	log.Infof(ctx, "[state] Saving cluster state to cluster nodes")
+	clusterFile, err := yaml.Marshal(*clusterState)
+	if err != nil {
+		return err
+	}
+	for _, host := range uniqueHosts {
+		pki.DeployStateOnPlaneHost(ctx, host, alpineImage, prsMap, string(clusterFile))
```
alena1108: doesn't look like we check for error here? should we? given it's in the method signature

cluster/state.go
```diff
@@ -151,13 +184,37 @@ func getStateFromKubernetes(ctx context.Context, kubeClient *kubernetes.Clientse
...
+		if err == nil {
+			break
+		}
+	}
+	if len(clusterFile) == 0 {
 		return nil
 	}
+	err = yaml.Unmarshal([]byte(clusterFile), &currentCluster)
+	if err != nil {
+		return nil
```
alena1108: should we at least log the error?

PR #910, opened 2018-09-18
Publish archives in release

superseb:
* Synced the build script with `rancher/cli`
* Removed publishing `./bin/rke`

Probably best to merge this after releasing v0.1.10, then we can release v0.1.11 with binaries and archives and make a note that it will be only archives in v0.1.12.

@cloudnautique to check


superseb:
https://github.com/rancher/rke/issues/911

PR #914, opened 2018-09-19
Container stop timeout

galal-hussein:
https://github.com/rancher/rancher/issues/15422
https://github.com/rancher/rancher/issues/15695


docker/docker.go
```diff
@@ -28,6 +28,7 @@ import (
 const (
 	DockerRegistryURL = "docker.io"
 	RestartTimeout    = 30
+	StopTimeout       = 30
```
alena1108: nit: although it's kinda obvious, lets put a comment that it's in seconds

docker/docker.go
```diff
@@ -225,15 +220,17 @@ func UseLocalOrPull(ctx context.Context, dClient *client.Client, hostname string
...
 		return fmt.Errorf("Can't remove Docker container [%s] for host [%s]: %v", containerName, hostname, err)
 	}
 	return nil
 }
 
 func StopContainer(ctx context.Context, dClient *client.Client, hostname string, containerName string) error {
-	err := dClient.ContainerStop(ctx, containerName, nil)
+	// define the stop timeout
+	stopTimeoutDuration := StopTimeout * time.Second
+	err := dClient.ContainerStop(ctx, containerName, &stopTimeoutDuration)
```
alena1108: what will docker do if the timeout is hit?

docker/docker.go
```diff
@@ -225,15 +220,17 @@ func UseLocalOrPull(ctx context.Context, dClient *client.Client, hostname string
 }
 
 func RemoveContainer(ctx context.Context, dClient *client.Client, hostname string, containerName string) error {
-	err := dClient.ContainerRemove(ctx, containerName, types.ContainerRemoveOptions{})
+	err := dClient.ContainerRemove(ctx, containerName, types.ContainerRemoveOptions{Force: true})
```
alena1108: what does force option do? If the container failed to stop, will it remove this container? Any implications?

galal-hussein: the force option will remove the container even if its running, i followed the same way its implemented in rancher repo https://github.com/rancher/rancher/blob/master/pkg/rkeworker/docker.go#L137-L142

docker/docker.go
galal-hussein: will do

cluster/network.go
```diff
@@ -262,30 +262,18 @@ func (c *Cluster) checkKubeAPIPort(ctx context.Context) error {
 func (c *Cluster) deployTCPPortListeners(ctx context.Context, currentCluster *Cluster) error {
 	log.Infof(ctx, "[network] Deploying port listener containers")
 
-	etcdHosts := []*hosts.Host{}
```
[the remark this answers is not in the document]
galal-hussein: because we handle this part already after deploying the listener:

- deploy listener on host
- if the address already bind to the host, then skip deploying the listener

it doesn't make sense to only deploy on new hosts added to the cluster, this way we deploy on everything which will mitigate any edge case including the one in issue rancher/rancher#15422

docker/docker.go
galal-hussein: docker will force stop the container, the default is 10 seconds

PR #919, opened 2018-09-26
Handle deleting all controlplane nodes when nodes are unreachable

galal-hussein:
https://github.com/rancher/rancher/issues/15810

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -315,7 +315,20 @@ func (c *Cluster) deployAddons(ctx context.Context) error {
 	return nil
 }
 
-func (c *Cluster) SyncLabelsAndTaints(ctx context.Context) error {
+func (c *Cluster) SyncLabelsAndTaints(ctx context.Context, currentCluster *Cluster) error {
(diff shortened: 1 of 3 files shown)


cluster/cluster.go
```diff
@@ -383,3 +394,21 @@ func ConfigureCluster(
 	}
 	return nil
 }
+
+func cleanControlNodes(ctx context.Context, kubeCluster, currentCluster *Cluster, cpToDelete []*hosts.Host) error {
+	for _, toDeleteHost := range cpToDelete {
+		kubeClient, err := k8s.NewClient(kubeCluster.LocalKubeConfigPath, kubeCluster.K8sWrapTransport)
```
alena1108: it feels like a repetition of https://github.com/rancher/rke/blob/5b3fec01657570642dd3ae61029b1fe270679af9/cluster/reconcile.go#L102. Can we re-use the same cleanup code logic?

alena1108: You can pass a custom message for the control node use case

PR #924, opened 2018-10-01
advertise address on kube-apiserver by internal_address

code0x9:
on certain circumstance, kubernetes apiserver cannot detect IP address from default routes. 
this PR uses internal_address from cluster.yml to propergate "advertise-address" for mitigate this.

excerpt from kube-apiserver error
```
I1001 09:38:56.390931       1 interface.go:360] Looking for default routes with IPv4 addresses
I1001 09:38:56.390938       1 interface.go:365] Default route transits interface "eth0.100"
I1001 09:38:56.392911       1 interface.go:174] Interface eth0.100 is up
I1001 09:38:56.392970       1 interface.go:222] Interface "eth0.100" has 1 addresses :[fe80::266e:96ff:fed4:954e/64].
I1001 09:38:56.392983       1 interface.go:189] Checking addr  fe80::266e:96ff:fed4:954e/64.
I1001 09:38:56.392989       1 interface.go:202] fe80::266e:96ff:fed4:954e is not an IPv4 address
I1001 09:38:56.392999       1 interface.go:360] Looking for default routes with IPv6 addresses
I1001 09:38:56.393003       1 interface.go:365] Default route transits interface "eth0.100"
I1001 09:38:56.394297       1 interface.go:174] Interface eth0.100 is up
I1001 09:38:56.394345       1 interface.go:222] Interface "eth0.100" has 1 addresses :[fe80::266e:96ff:fed4:954e/64].
I1001 09:38:56.394355       1 interface.go:189] Checking addr  fe80::266e:96ff:fed4:954e/64.
I1001 09:38:56.394360       1 interface.go:199] Non-global unicast address found fe80::266e:96ff:fed4:954e
I1001 09:38:56.394366       1 interface.go:365] Default route transits interface "eth1.100"
I1001 09:38:56.395641       1 interface.go:174] Interface eth1.100 is up
I1001 09:38:56.395692       1 interface.go:222] Interface "eth1.100" has 1 addresses :[fe80::266e:96ff:fed4:9550/64].
I1001 09:38:56.395701       1 interface.go:189] Checking addr  fe80::266e:96ff:fed4:9550/64.
I1001 09:38:56.395706       1 interface.go:199] Non-global unicast address found fe80::266e:96ff:fed4:9550
I1001 09:38:56.395712       1 interface.go:376] No active IP found by looking at default routes
Error: Unable to find suitable network address.error='unable to select an IP from default routes.'. Try to set the AdvertiseAddress directly or provide a valid BindAddress to fix this.
Usage:
  kube-apiserver [flags]```

### DIFF
--- a/cluster/plan.go
+++ b/cluster/plan.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/md5"
 	"fmt"
+	"net"
 	"path"
 	"strconv"
 	"strings"
@@ -69,7 +70,7 @@ func BuildRKEConfigNodePlan(ctx context.Context, myCluster *Cluster, host *hosts
 		processes[services.NginxProxyContainerName] = myCluster.BuildProxyProcess()
 	}
 	if host.IsControl {
-		processes[services.KubeAPIContainerName] = myCluster.BuildKubeAPIProcess(prefixPath)
+		processes[services.KubeAPIContainerName] = myCluster.BuildKubeAPIProcess(host, prefixPath)
 		processes[services.KubeControllerContainerName] = myCluster.BuildKubeControllerProcess(prefixPath)
 		processes[services.SchedulerContainerName] = myCluster.BuildSchedulerProcess(prefixPath)
 
@@ -97,7 +98,7 @@ func BuildRKEConfigNodePlan(ctx context.Context, myCluster *Cluster, host *hosts
 	}
 }
 
-func (c *Cluster) BuildKubeAPIProcess(prefixPath string) v3.Process {
+func (c *Cluster) BuildKubeAPIProcess(host *hosts.Host, prefixPath string) v3.Process {
 	// check if external etcd is used
 	etcdConnectionString := services.GetEtcdConnString(c.EtcdHosts)
 	etcdPathPrefix := EtcdPathPrefix
@@ -191,6 +192,10 @@ func (c *Cluster) BuildKubeAPIProcess(prefixPath string) v3.Process {
 
 	if c.Authorization.Mode == services.RBACAuthorizationMode {
 		CommandArgs["authorization-mode"] = "Node,RBAC"
+	}
+
+	if len(host.InternalAddress) > 0 && net.ParseIP(host.InternalAddress) != nil {
+		CommandArgs["advertise-address"] = host.InternalAddress
 	}
 
 	// PodSecurityPolicy

### CODE REVIEW

**Comment by cuchac** [comment]
Hello,
my use-case also requires to set "advertise-address" on kube-apiserver. But in my case I need to set it to "external" address, not internal. I prepared very similar patch but using "hostname_override" to populate "advertise-address".

Should we introduce new host configuration option to set only "advertise-address"? Or hostname_override makes also sense for you?

**Comment by code0x9** [comment]
@cuchac I think kube-apiserver advertise-address and bind-address is for internal access and needs 'IP', not hostname.

from kubernetes reference document(https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/)
```
--advertise-address ip
--
  | The IP address on which to advertise the apiserver to members of the cluster. This address must be reachable by the rest of the cluster. If blank, the --bind-address will be used. If --bind-address is unspecified, the host's default interface will be used.

--bind-address ip     Default: 0.0.0.0
--
  | The IP address on which to listen for the --secure-port port. The associated interface(s) must be reachable by the rest of the cluster, and by CLI/web clients. If blank, all interfaces will be used (0.0.0.0 for all IPv4 interfaces and :: for all IPv6 interfaces).


```

**Comment by code0x9** [comment]
advertise-address **ip**
- This address must be **reachable by the rest of the cluster**.

File: cluster/plan.go [hunk 1]
```diff
@@ -171,6 +171,11 @@ func (c *Cluster) BuildKubeAPIProcess(prefixPath string) v3.Process {
 	if c.Authorization.Mode == services.RBACAuthorizationMode {
 		CommandArgs["authorization-mode"] = "Node,RBAC"
 	}
+
+	if len(host.InternalAddress) > 0 {
```
cuchac: I think becuse of https://github.com/rancher/rke/blob/22dc19205f794f19dc156d16695ff86c07a501e6/cluster/defaults.go#L76-L78 there could be hostname in InternalAddress causing apiserver to fail to start - it requires IP. It happens when you do not specify internal IP for node and specified address is hostname.

code0x9:
@cuchac Node annotation ```flannel.alpha.coreos.com/public-ip-overwrite=<internal_ip>``` can make only flannel user's life easier. Due to the nature of rke's network stack(which is pluggable), I think patching flannel plugin so that it can use advertise-address if defined is better way to solve your problem.

cluster/plan.go
code0x9: If InternalAddress meant to specify IP, there shouldn't be the hostname. 

see https://github.com/rancher/rke/blob/master/cluster/plan.go#L94 and https://github.com/rancher/rke/blob/master/k8s/node.go#L20

miyakelp:
I am interested in this PR. It is useful for multi-master cluster using non default interfaces.

Example:
```
- masterA
eth0 192.168.10.10 (default)
eth1 192.168.50.10

- masterB
eth0 192.168.10.11 (default)
eth1 192.168.50.11

- worker
eth0 192.168.20.20 (default)
eth1 192.168.50.20
```
In this case, we can use eth1 for a cluster by `flannel_iface` option, but `advertise-address` of each master node is assigned an address of default interface eth0.

This patch is one of the solution for this issue.

cluster/plan.go
code0x9: @galal-hussein seems reasonable. applied. thanks.

willmao:
this patch can solve my issue, thanks! since the apiservers use the default route ip as the advertise address, currently i have to use the nodes without public ip(default route ip) as the api server.

cluster/plan.go
willmao: I just found adding the condition `host.InternalAddress != host.Address` can't solve my problem. In my case, the address is the same as the internal address, which is the ip of the private network interface, i want to specify the advertise-address arg to it since my public network ip is not accessible  by other nodes due to network policy.

code0x9: @willmao okay then, where do you put your accessable IP?

willmao: @code0x9 the internal ip address provided by my cloud provider can be accessible by all nodes in the cluster, i think the advertise address should be set to the internal address once the `internal_address ` field is set and the format is ipv4 address(escape the address fqdn case)

sorry for the late reply!

code0x9: @willmao right. changed check logic from ```host.InternalAddress != host.Address``` to ```net.ParseIP(host.InternalAddress) != nil```, using golang builtin net package. thanks.

willmao: @code0x9 my pleasure! hoping this pr could be merged before our kubernetes goes to production three month later...

willmao:
@code0x9 pls resolve the conflicts, thanks!

willmao:
@galal-hussein could you spare some time to review this pr again? thanks!

Issue #922, opened 2018-09-28
Make k8s v1.11.3 default version.


sangeethah:
rke version v0.1.10-rc4

When installing a Kubernetes cluster using rke, cluster get created with  k8s version v1.11.2.

We should support k8s v1.11.3 as default version.

moelsayed:
fixed in https://github.com/rancher/rke/pull/927

soumyalj:
Tested with rke version v0.1.10-rc5.
Created rke cluster. Cluster shows version v1.11.3 as below
```
soumyas-MacBook-Pro:rke soumya$ kubectl --kubeconfig kube_config_testabc.yml version
Client Version: version.Info{Major:"1", Minor:"10", GitVersion:"v1.10.0", GitCommit:"fc32d2f3698e36b93322a3465f63a14e9f0eaead", GitTreeState:"clean", BuildDate:"2018-03-26T16:55:54Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.3", GitCommit:"a4529464e4629c21224b3d52edfe0ea91b072862", GitTreeState:"clean", BuildDate:"2018-09-09T17:53:03Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"}
soumyas-MacBook-Pro:rke soumya$ kubectl --kubeconfig kube_config_testabc.yml get node
NAME             STATUS    ROLES                      AGE       VERSION
a.b.c.d   Ready     controlplane,etcd,worker   4m        v1.11.3
```

PR #927, opened 2018-10-01
K8s 1.12

moelsayed:
https://github.com/rancher/types/pull/584


moelsayed:
@alena1108 #922 is fixed automatically by updating types, since that's where rke gets the default version from.

PR #928, opened 2018-10-02
Add supported Docker versions for 1.12

superseb:
```
The list of validated docker versions was updated to
1.11.1, 1.12.1, 1.13.1, 17.03, 17.06, 17.09, 18.06. (#68495)
```

https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.12.md#external-dependencies


superseb:
This is already in https://github.com/rancher/rke/pull/927, nice

Issue #938, opened 2018-10-06
Allow us to specify service.<kubelet, kubeproxy>.extra_args so that we can specify different configuration for each node


ukinau:
Sometimes we want to use different configuration for each node. But currently rke doesn't allow us to specify different configuration for each node like enabling load-reader on some nodes and disabling others, configuration override (extra_args) is always need to be configured as a cluster configuration.

This is useful when we try to enable some option on only specific node first and ensure it works then use this as cluster configuration,

vincent99:
I think this is going to need design discussion (@cjellick), creating mixed clusters is something of an anti-pattern and this seems like it creates an awful lot of new ways to break things.

alena1108:
`This is useful when we try to enable some option on only specific node first and ensure it works then use this as cluster configuration` - scenarios similar to this is better to be tested on a separate cluster. Introducing arguments discrepancy between nodes having the same role is highly undesirable

ukinau:
that's make sense and actually we decided not to use this feature. so let me close my PR.
Thanks for checking.

PR #939, opened 2018-10-06
Allow to override option of kubelet, kubeproxy for each node

ukinau:
One kubernetes cluster can accomodate multiple node having kubelet and
kube-proxy running with different configuration because they are only
affecting just own node.
So we should allow user to define/override command argument for each node.

Current rke implementation only allow us to have same command arguments
for all kube-proxy, kubelet deployment which will cause some problem
when we try to build a cluster with nodes from multiple cloud.

So this commit added the point to override command argument if
rkeconfignode have service.<kubelet, kubeproxy>.extra_args is defined

This implement following feature request:
https://github.com/rancher/rke/issues/938


ukinau:
This commit is depending on following PR for rancher/types
https://github.com/rancher/types/pull/588

Issue #962, opened 2018-10-16
nginx ingress controller extra_args definition does not work and causes rke-ingress-controller-deploy-job to fail


Oats87:
**RKE version:**
0.1.10

**cluster.yml file:**
```
ingress: 
  extra_args: 
    default-backend-service: "ingress-nginx/pretty-error"
  options: 
    custom-http-errors: "404"
  provider: "nginx"
```
**Steps to Reproduce:**
Attempt to set any `extra_args` in the RKE cluster.yml, when cluster is brought up, the nginx-deploy job will fail because of something similar to 
```
namespace/ingress-nginx configured
configmap/nginx-configuration configured
configmap/tcp-services unchanged
configmap/udp-services unchanged
serviceaccount/nginx-ingress-serviceaccount unchanged
clusterrole.rbac.authorization.k8s.io/nginx-ingress-clusterrole configured
role.rbac.authorization.k8s.io/nginx-ingress-role unchanged
rolebinding.rbac.authorization.k8s.io/nginx-ingress-role-nisa-binding unchanged
clusterrolebinding.rbac.authorization.k8s.io/nginx-ingress-clusterrole-nisa-binding configured
error: error parsing /etc/config/rke-ingress-controller.yaml: error converting YAML to JSON: yaml: line 50: did not find expected key
```

superseb:
See https://github.com/rancher/rke/issues/945 / https://github.com/rancher/rke/pull/954

sangeethah:
This issue shows up as rke-ingress-controller addons  not being able to start successfully when creating cluster having the following yml configurations for ingress using rancher server version - v2.1.1-rc4.


```
ingress: 
  extra_args: 
    default-backend-service: "ingress-nginx/pretty-error"
  options: 
    custom-http-errors: "404"
  provider: "nginx"
```
Logs from  rke-ingress-controller-deploy-job:

```
clusterrole.rbac.authorization.k8s.io/nginx-ingress-clusterrole configured
role.rbac.authorization.k8s.io/nginx-ingress-role unchanged
rolebinding.rbac.authorization.k8s.io/nginx-ingress-role-nisa-binding unchanged
clusterrolebinding.rbac.authorization.k8s.io/nginx-ingress-clusterrole-nisa-binding configured
error: error parsing /etc/config/rke-ingress-controller.yaml: error converting YAML to JSON: yaml: line 50: did not find expected key```

sangeethah:
[...]

PR #963, opened 2018-10-16
Moving extra_args iteration to allow extra_args to work successfully

Oats87:
https://github.com/rancher/rke/issues/962


alena1108:
LGTM

Issue #616, opened 2018-05-15
Add S3 for etcd backup/restore


galal-hussein:
Currently etcd backup/restore uses local disk for backup storage, it will be a nice enhancement to etcd backup to save and restore backups from Amazon S3 buckets to both the regular backups and onetime backups, for example:
cluster.yml
```
services:
  etcd:
    backup: true
    s3: true
    s3_bucket_name: "etcd-backup"
```
and for one time backup:
```
./rke etcd backup --name snapshot --config cluster.yml --s3 --s3-bucket-name "etcd-backup"
```

curx:
or on others, like minio ?

techcto:
I +1 this to s3

It would also be nice to have a feature to control etcd backup size. On my test server, it is growing rapidly. Are there any better ways to control it's size?  

I see this page:  https://coreos.com/etcd/docs/latest/op-guide/maintenance.html

iwilltry42:
Hi there,
I created two PRs in https://github.com/rancher/rke/pull/968 and https://github.com/rancher/rke-tools/pull/27 to solve this.
This is my first contribution to a larger project, so I hope I didn't do too much wrong... happy about reviews :)

mitchellmaler:
Will this support other S3 providers and not just AWS? We have a local minio cluster that runs with path based buckets. Would be nice to be able to specify the URL and bucket (path) with auth token, etc.

guangbochen:
Addressed in https://github.com/rancher/rke/pull/1036.

soumyalj:
Tested with rke version v0.2.0-rc5.  Backup restore using S3 is functional
1. Create a cluster using RKE CLI specifying S3 configuration in the cluster.yml file
```
auth:
  strategy: x509
nodes:
  - address: a.b.cd
    user: root
    role: [etcd,controlplane,worker]
  - address: e.f.g.h
    user: root
    role: [etcd,controlplane,worker]
  - address: i.j.k.l
    user: root
    role: [etcd,controlplane,worker]
services:
  etcd:
    snapshot: true
    backup_config:
      s3BackupConfig:
        access_key: "efgh"
        secret_key: "abcd"
        bucket_name: "testbucket2"
        endpoint: s3.amazonaws.com
```

2. Create few pods.
3. Create a one time backup using the command: 
```
[...]

PR #968, opened 2018-10-17
WIP Add functionality to upload etcd snapshots to S3

iwilltry42:
Hi there,

This PR adds the functionality to RKE to upload etcd snapshots directly to AWS S3.
Feature request this targets: https://github.com/rancher/rke/issues/616
It adds the following new flags:

-  `--s3`: feature toggle
- `--s3-bucket`: where to upload it (can be `<bucket>/<sub-path>`)
- `--s3-credentials FILE`: path to AWS credentials file
- `--s3-config FILE`: path to AWS config file (for region)

By default, the credentials + config from `~/.aws/credentials` + `~/.aws/config` will be used respectively.

Link to related PR in rke-tools: https://github.com/rancher/rke-tools/pull/27

### DIFF
--- a/cmd/etcd.go
+++ b/cmd/etcd.go
@@ -26,6 +26,24 @@ func EtcdCommand() cli.Command {
 			Value:  pki.ClusterConfig,
 			EnvVar: "RKE_CONFIG",
 		},
+		cli.BoolTFlag{
+			Name:  "s3",
+			Usage: "Upload Snapshot to AWS S3",
+		},
+		cli.StringFlag{
+			Name:  "s3-bucket",
+			Usage: "Specify an AWS S3 bucket name",
+		},
+		cli.StringFlag{
+			Name:  "s3-credentials",
+			Usage: "Load AWS S3 credentials from `FILE`",
+			Value: "~/.aws/credentials",
+		},
+		cli.StringFlag{
+			Name:  "s3-config",
+			Usage: "Load AWS S3 configuration from `FILE`",
+			Value: "~/.aws/config",
+		},
 	}
 
 	snapshotFlags = append(snapshotFlags, commonFlags...)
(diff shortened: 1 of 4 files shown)


services/etcd.go
```diff
@@ -256,6 +257,34 @@ func RunEtcdSnapshotSave(ctx context.Context, etcdHost *hosts.Host, prsMap map[s
 		imageCfg.Cmd = append(imageCfg.Cmd, "--retention="+retention)
 		imageCfg.Cmd = append(imageCfg.Cmd, "--creation="+creation)
 	}
+	// boolean flag to export to AWS S3 is true -> set parameters for login and backup location
+	if s3 {
+		imageCfg.Cmd = append(imageCfg.Cmd, "--s3")
+		imageCfg.Cmd = append(imageCfg.Cmd, "--s3bucket="+s3bucket)
```
galal-hussein: I think you should add some kind of validation in validation code to make sure that the user passes at least the s3bucket if s3 option is passed

cmd/etcd.go
```diff
@@ -65,7 +84,7 @@ func SnapshotSaveEtcdHosts(
 	if err := kubeCluster.TunnelHosts(ctx, false); err != nil {
 		return err
 	}
-	if err := kubeCluster.SnapshotEtcd(ctx, snapshotName); err != nil {
+	if err := kubeCluster.SnapshotEtcd(ctx, snapshotName, s3bucket, s3credsfile, s3configfile, s3); err != nil {
```
galal-hussein: it could be useful to include these parameters in one structure

galal-hussein:
@iwilltry42 I think the main reason from using s3 is to make a centralized part for not only backup but for restore as well which is more time consuming when using files, can you check if it is possible to also restore from a snapshot placed on s3, it also may be necessary to include the certificate bundle on s3 as well

iwilltry42:
@galal-hussein thanks for the feedback, will add the following functionality tomorrow:
- parse flag arguments into a single struct and pass this on instead of single values 
- enable periodic S3 backup in cluster.yaml (forgot about that)
- restore from S3 option
- validate required user input

iwilltry42:
@galal-hussein: enabling periodic backups via specification in the `rancher-cluster.yml` will require another PR to update the `ETCDService` struct in [types/.../rke_types.go](https://github.com/rancher/types/blob/master/apis/management.cattle.io/v3/rke_types.go)

iwilltry42:
@alena1108 , ok, so I discontinue this draft for now.
I "only" had cloud storage providers in mind, which is why I chose file-based authentication, as all the major cloud providers support this (with similar file setups) with their client SDKs.
Also, it might be interesting, if one could merge the snapshots from the different nodes into a single archive before uploading them to the backup destination.

alena1108:
Closing for now; S3 backup work is being tracked in https://github.com/rancher/rke/pull/1036

PR #1005, opened 2018-11-07
RKE remote k8s auth

erikwilson:
For rancher/rancher/issues/13698


cluster/plan.go
```diff
@@ -119,34 +120,36 @@ func (c *Cluster) BuildKubeAPIProcess(prefixPath string) v3.Process {
...
+		"etcd-servers":                             etcdConnectionString,
+		"etcd-prefix":                              etcdPathPrefix,
+		"requestheader-client-ca-file":             pki.GetCertPath(pki.RequestHeaderCACertName),
+		"requestheader-allowed-names":              pki.APIProxyClientCertName,
+		"proxy-client-key-file":                    pki.GetKeyPath(pki.APIProxyClientCertName),
+		"proxy-client-cert-file":                   pki.GetCertPath(pki.APIProxyClientCertName),
+		"requestheader-extra-headers-prefix":       "X-Remote-Extra-",
+		"requestheader-group-headers":              "X-Remote-Group",
+		"requestheader-username-headers":           "X-Remote-User",
+		"authentication-token-webhook-config-file": AuthnWebhookConfigFile,
```
galal-hussein: I don't think we should allow the webhook token by default, user must have the choice not to use it.

pki/util.go
```diff
@@ -248,6 +248,7 @@ func getControlCertKeys() []string {
 		KubeSchedulerCertName,
 		KubeProxyCertName,
 		KubeNodeCertName,
+		KubeAdminCertName,
```
galal-hussein: why do you deploy the kubeadmin certs on the control nodes?

cluster/authn-webhook.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/docker/docker/api/types/container"
+	"github.com/rancher/rke/docker"
+	"github.com/rancher/rke/hosts"
+	"github.com/rancher/rke/log"
+	"github.com/rancher/types/apis/management.cattle.io/v3"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	AuthnWebhookConfigDeployer    = "authn-webhook-config-deployer"
```
galal-hussein: i think we should add that to the pki package, the problem is that we will end up with a lot of containers to deploy artifacts, so i think there are two choices:
- use one container to deploy all artifacts:
   - cloud config file
   - auth token file

or 
- add the auth token config file to the x509 cert and config list

cluster/plan.go
erikwilson: Logic has been modified to conditionally add webhook params if config is defined.

pki/util.go
erikwilson: Removed, deployment is now controller by Rancher server template.

cluster/authn-webhook.go
erikwilson: Have created one container deployer for cloud & webhook config files.

cluster/defaults.go
```diff
@@ -268,6 +272,28 @@ func (c *Cluster) setClusterNetworkDefaults() {
 	}
 }
 
+func (c *Cluster) setClusterAuthnDefaults() {
+	setDefaultIfEmpty(&c.Authentication.Strategy, DefaultAuthStrategy)
+
+	for _, S := range strings.Split(c.Authentication.Strategy, "|") {
```
cjellick: @erikwilson why did you have to use`|` again? Would really prefer `,` separated list.

cjellick: Using capital `S` vs `s` is a bad pattern. Can you use more meaningful names?

erikwilson: `|` was used because default values in golang tags can not contain `,` - webhook is not currently set as a default strategy so we can change it back to `,` if desired, but may cause issues for a default delimitated value in the future.

Will fix variable naming.

cluster/validation.go
```diff
@@ -39,8 +40,18 @@ func (c *Cluster) ValidateCluster() error {
...
 
 func validateAuthOptions(c *Cluster) error {
-	if c.Authentication.Strategy != DefaultAuthStrategy {
-		return fmt.Errorf("Authentication strategy [%s] is not supported", c.Authentication.Strategy)
+	for S, v := range c.AuthnStrategies {
+		if !v {
+			continue
+		}
+		s := strings.ToLower(S)
+		logrus.Info("Has authn strategy: ", s)
```
galal-hussein: can we add this as a debug instead of info

erikwilson: cleaned up variable names & removed logging

cjellick:
@erikwilson can you improve the `cluster auth token update` commit message to have an explanation of what you are changing and why. The best guide ive found:
https://chris.beams.io/posts/git-commit/

cjellick:
LGTM, other than commit message

PR #1009, opened 2018-11-08
Refactoring State Management

galal-hussein:


cluster/state.go
```diff
@@ -337,10 +150,10 @@ func RebuildState(ctx context.Context, rkeConfig *v3.RancherKubernetesEngineConf
 			return nil, fmt.Errorf("Failed to generate certificate bundle: %v", err)
 		}
 		// Convert rke certs to v3.certs
```
moelsayed: nit: remove the comment

cluster/certificates.go
```diff
@@ -19,86 +19,10 @@ import (
 	"k8s.io/client-go/util/cert"
 )
 
-func SetUpAuthentication(ctx context.Context, kubeCluster, currentCluster *Cluster) error {
+func SetUpAuthentication(ctx context.Context, kubeCluster, currentCluster *Cluster, fullState *RKEFullState) error {
```
alena1108: can we drop RKE from the name? and call the parameter `clusterState`?

cluster/certificates.go
```diff
@@ -117,8 +41,13 @@ func regenerateAPICertificate(c *Cluster, certificates map[string]pki.Certificat
 	return certificates, nil
 }
 
-func getClusterCerts(ctx context.Context, kubeClient *kubernetes.Clientset, etcdHosts []*hosts.Host) (map[string]pki.CertificatePKI, error) {
+func GetClusterCertsFromKubernetes(ctx context.Context, localConfigPath string, k8sWrapTransport k8s.WrapTransport, etcdHosts []*hosts.Host) (map[string]pki.CertificatePKI, error) {
```
alena1108: do you think it would make sense to wrap localConfigPath and k8sWrapTransport into a struct? and pass the struct instead of passing the params separately

cluster/cluster.go
```diff
@@ -54,6 +55,7 @@ type Cluster struct {
 const (
 	X509AuthenticationProvider = "x509"
 	StateConfigMapName         = "cluster-state"
+	FullStateConfigMapName     = "full-cluster-state"
```
alena1108: what is the diff between StateConfigMapName (cluster-state) and FullStateConfigMapName (full-cluster-state)?

moelsayed: StateConfigMapName is the name we used with "legacy" cluster state. FullStateConfigMapName is the name of the new Desired/Current state we save on kubernetes. We renamed it because we actually delete StateConfigMapName when we handle the legacy cluster upgrade. 

StateConfigMapName will be removed in the future.

cmd/up.go
```diff
@@ -61,6 +67,73 @@ func UpCommand() cli.Command {
...
+	rkeState := cluster.RKEFullState{
+		DesiredState: fullState.DesiredState,
+		CurrentState: fullState.CurrentState,
+	}
+	return rkeState.WriteStateFile(ctx, stateFilePath)
+}
+
 func ClusterUp(
 	ctx context.Context,
 	rkeConfig *v3.RancherKubernetesEngineConfig,
```
alena1108: I guess we don't need rkeConfig as it's supposed to be initialized with init?

alena1108: I guess we don't need rkeConfig as it's supposed to be initialized with init?

cluster/state.go
```diff
@@ -249,3 +134,106 @@ func GetK8sVersion(localConfigPath string, k8sWrapTransport k8s.WrapTransport) (
 	}
 	return fmt.Sprintf("%#v", *serverVersion), nil
 }
+
+func RebuildState(ctx context.Context, rkeConfig *v3.RancherKubernetesEngineConfig, oldState *RKEFullState, configPath, configDir string) (*RKEFullState, error) {
```
alena1108: can we reflect the diff between RebuildState and UpdateClusterState?

cluster/state.go
```diff
@@ -249,3 +134,106 @@ func GetK8sVersion(localConfigPath string, k8sWrapTransport k8s.WrapTransport) (
...
+		if err := pki.GenerateKubeAdminCertificate(ctx, pkiCertBundle, *rkeConfig, configPath, configDir); err != nil {
+			return nil, err
+		}
+		newState.DesiredState.CertificatesBundle = pkiCertBundle
+	}
+	newState.CurrentState = oldState.CurrentState
+	return newState, nil
+}
+
+func (s *RKEFullState) WriteStateFile(ctx context.Context, statePath string) error {
```
alena1108: lets make all the methods that are called within the package only, start with lowercase letter.

cluster/state.go
```diff
@@ -249,3 +134,106 @@ func GetK8sVersion(localConfigPath string, k8sWrapTransport k8s.WrapTransport) (
...
+	}
+	rkeFullState.DesiredState.CertificatesBundle = pki.TransformPEMToObject(rkeFullState.DesiredState.CertificatesBundle)
+	rkeFullState.CurrentState.CertificatesBundle = pki.TransformPEMToObject(rkeFullState.CurrentState.CertificatesBundle)
+	return rkeFullState, nil
+}
+
+func RemoveStateFile(ctx context.Context, statePath string) {
+	log.Infof(ctx, "Removing state file: %s", statePath)
+	if err := os.Remove(statePath); err != nil {
+		logrus.Warningf("Failed to remove state file: %v", err)
```
alena1108: what will happen if the file failed to remove, and user initiates a new cluster from the same directory?

cluster/state.go
```diff
@@ -249,3 +134,106 @@ func GetK8sVersion(localConfigPath string, k8sWrapTransport k8s.WrapTransport) (
...
+	}
+	log.Infof(ctx, "State file removed successfully")
+}
+
+func RemoveLegacyStateFromKubernets(ctx context.Context, kubeConfigPath string, k8sWrapTransport k8s.WrapTransport) error {
+	k8sClient, err := k8s.NewClient(kubeConfigPath, k8sWrapTransport)
+	if err != nil {
+		return fmt.Errorf("Failed to create Kubernetes Client: %v", err)
+	}
+	return k8s.DeleteConfigMap(k8sClient, StateConfigMapName)
```
alena1108: what are the implications of leaving the legacy state around? Thinking of situations when user upgrades rancher, triggers rke cluster update, and then rollbacks rancher to the older version where the state was expected to exist on the cluster side.

cluster/state.go
```diff
@@ -2,142 +2,77 @@ package cluster
...
 
-			currentCluster.Certificates, err = regenerateAPICertificate(c, currentCluster.Certificates)
-			if err != nil {
-				return nil, fmt.Errorf("Failed to regenerate KubeAPI certificate %v", err)
-			}
-		}
+func (c *Cluster) GetClusterState(ctx context.Context, fullState *RKEFullState, configDir string) (*Cluster, error) {
+	var err error
+	if fullState.CurrentState.RancherKubernetesEngineConfig == nil {
+		return nil, nil
```
alena1108: how do we handle nil cluster being returned?

cluster/state.go
```diff
@@ -249,3 +134,106 @@ func GetK8sVersion(localConfigPath string, k8sWrapTransport k8s.WrapTransport) (
...
+		if err := pki.GenerateKubeAPICertificate(ctx, pkiCertBundle, *rkeConfig, "", ""); err != nil {
+			return nil, err
+		}
+		// Regenerating kubeadmin certificates/config
+		if err := pki.GenerateKubeAdminCertificate(ctx, pkiCertBundle, *rkeConfig, configPath, configDir); err != nil {
+			return nil, err
+		}
+		newState.DesiredState.CertificatesBundle = pkiCertBundle
+	}
+	newState.CurrentState = oldState.CurrentState
```
alena1108: do we need to do a deep copy? Do oldState being updated anywhere?

cmd/up.go
```diff
@@ -61,6 +67,73 @@ func UpCommand() cli.Command {
...
+	if err != nil {
+		log.Warnf(ctx, "[state] can't fetch legacy cluster state from Kubernetes")
+	}
+
+	fullState, err := cluster.RebuildState(ctx, &kubeCluster.RancherKubernetesEngineConfig, rkeFullState, clusterFilePath, configDir)
+	if err != nil {
+		return err
+	}
+
+	rkeState := cluster.RKEFullState{
```
alena1108: where do we use rkeState?

cmd/up.go
```diff
@@ -70,7 +143,21 @@ func ClusterUp(
 
 	log.Infof(ctx, "Building Kubernetes cluster")
 	var APIURL, caCrt, clientCert, clientKey string
-	kubeCluster, err := cluster.ParseCluster(ctx, rkeConfig, clusterFilePath, configDir, dockerDialerFactory, localConnDialerFactory, k8sWrapTransport)
+
+	// is tehre any chance we can store the cluster object here instead of the rke config ?
```
alena1108: Do we intend to leave these comments around?

cluster/state.go
galal-hussein: we do call this method from cmd package

cluster/state.go
galal-hussein: we can't change this to error, because it will generate an error if the user runs rke remove twice

galal-hussein: the same case applies to the kube config file

cluster/state.go
galal-hussein: that's true, i will keep the config map even after upgrade

cmd/up.go
galal-hussein: removing this comment

cmd/up.go
galal-hussein: fixed

cluster/certificates.go
galal-hussein: fixed

cluster/certificates.go
galal-hussein: fixed

cluster/state.go
galal-hussein: fixed

Issue #992, opened 2018-10-30
Rolling Snapshots are not done after docker host reboot


RedPepper:
**RKE version:**
`
rke version v0.1.9
`
**Docker version: (`docker version`,`docker info` preferred)**
`
Containers: 48
 Running: 26
 Paused: 0
 Stopped: 22
Images: 17
Server Version: 17.03.2-ce
Storage Driver: overlay2
 Backing Filesystem: extfs
 Supports d_type: true
 Native Overlay Diff: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins: 
 Volume: local
 Network: bridge host macvlan null overlay
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: 4ab9917febca54791c5f071a9d1f404867857fcc
runc version: 54296cf40ad8143b62dbcaa1d90e520a2136ddfe
init version: 949e6fa
Security Options:
 seccomp
  Profile: default
Kernel Version: 4.9.0-8-amd64
Operating System: Debian GNU/Linux 9 (stretch)
OSType: linux
Architecture: x86_64
CPUs: 6
Total Memory: 9.743 GiB
Name: node1.k8s.dev.de.esl.systems
ID: KKQ2:DUQ2:IJFN:HFFI:O6G7:EPKW:FHLD:LWJZ:AK7V:RMMR:F2OT:LM77
Docker Root Dir: /var/lib/docker
Debug Mode (client): false
Debug Mode (server): false
Registry: https://index.docker.io/v1/
Experimental: false
Insecure Registries:
 127.0.0.0/8
Live Restore Enabled: false

WARNING: No swap limit support

`
**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)**
`
cat /etc/os-release 
PRETTY_NAME="Debian GNU/Linux 9 (stretch)"
NAME="Debian GNU/Linux"
VERSION_ID="9"
VERSION="9 (stretch)"
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
uname -r
4.9.0-8-amd64

`
**Type/provider of hosts: (VirtualBox/Bare-metal/AWS/GCE/DO)**
Nutanix VM
**cluster.yml file:**
`
---
ssh_agent_auth: true
nodes:
  - address: node1
    user: root
    role: [controlplane,etcd,worker]

[...]

PR #1012, opened 2018-11-08
Fix broken etcd rolling snapshot

galal-hussein:
https://github.com/rancher/rancher/issues/16511


galal-hussein:
https://github.com/rancher/rke/issues/992

Issue #722, opened 2018-06-25
CoreDNS as an alternative to kube-dns


bootc:
This is a feature request. Please provide an option to deploy [CoreDNS](https://coredns.io/) instead of kube-dns. Bonus points for providing a migration path between them.

https://kubernetes.io/docs/tasks/administer-cluster/coredns/

mkjoerg:
This would be awesome, if included directly into installation.

hiscal2015:
+ 1

LiouPinChen:
Thank you for helping me. I want to understand RKE more and try to deal with core-dns. My codes seems workable. However, they still have some problems. Pls enlighten me on this issue and help me. Thank you very much.

<img width="1274" alt="2018-08-30 9 21 26" src="https://user-images.githubusercontent.com/38847157/44826619-4af21c00-ac42-11e8-9f95-7a28be241794.png">

<img width="499" alt="2018-08-30 9 23 39" src="https://user-images.githubusercontent.com/38847157/44826644-5cd3bf00-ac42-11e8-97e8-c26ba165433e.png">

<img width="680" alt="2018-08-30 9 23 02" src="https://user-images.githubusercontent.com/38847157/44826658-66f5bd80-ac42-11e8-91c3-1e4ee5300a30.png">

superseb:
Put some time into this today, and had a first deploy. Hopefully it can make it in the next release as experimental option

thearifismail:
@superseb what is the expected timeframe?  Anything available for kicking tires?

superseb:
PR should be up this week

thearifismail:
@superseb Great. I have been waiting for it to try it out.  How will users know when it becomes available?

pulberg:
@superseb Any update on this?

sdelrio:
@pulberg, I'm also interested in this feature. I saw a PR https://github.com/rancher/rke/pull/1018 about this, but seems CI test is failing

superseb:
We are reviewing it currently, CI is failing due to types not being merged yet (https://github.com/rancher/types/pull/621)

alena1108:
Can be tested in v0.2.0-rc5

sdelrio:
Nice with  v0.2.0-rc5, it smoothly upgrades, just a small warning but probably a timeout, just less than second after it says deployed successfully
```
...
INFO[0207] [dns] removing DNS provider kubedns
INFO[0249] [dns] DNS provider kubedns removed successfully
[...]

PR #1018, opened 2018-11-12
Add CoreDNS as addon

superseb:
https://github.com/rancher/rke/issues/722

Depends on: https://github.com/rancher/types/pull/621

Ready to review

### DIFF
--- a/cluster/addons.go
+++ b/cluster/addons.go
@@ -20,7 +20,6 @@ import (
 )
 
 const (
-	KubeDNSAddonResourceName      = "rke-kubedns-addon"
 	UserAddonResourceName         = "rke-user-addon"
 	IngressAddonResourceName      = "rke-ingress-controller"
 	UserAddonsIncludeResourceName = "rke-user-includes-addons"
@@ -31,7 +30,12 @@ const (
 	NginxIngressAddonAppName       = "ingress-nginx"
 	KubeDNSAddonAppName            = "kube-dns"
 	KubeDNSAutoscalerAppName       = "kube-dns-autoscaler"
+	CoreDNSAutoscalerAppName       = "coredns-autoscaler"
+
+	CoreDNSProvider = "coredns"
 )
+
+var DNSProviders = []string{"kubedns", "coredns"}
 
 type ingressOptions struct {
 	RBACConfig     string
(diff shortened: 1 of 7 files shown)


cluster/addons.go
```diff
@@ -377,3 +433,69 @@ func (c *Cluster) deployIngress(ctx context.Context) error {
...
+		}
+		if err := c.deployKubeDNS(ctx); err != nil {
+			if err, ok := err.(*addonError); ok && err.isCritical {
+				return err
+			}
+			log.Warnf(ctx, "Failed to deploy addon execute job [%s]: %v", getAddonResourceName(c.DNS.Provider), err)
+		}
+		log.Infof(ctx, "[dns] DNS provider %s deployed successfully", c.DNS.Provider)
+		return nil
+	case "coredns":
```
galal-hussein: can the name be a constant ?

cluster/addons.go
```diff
@@ -377,3 +433,69 @@ func (c *Cluster) deployIngress(ctx context.Context) error {
...
+}
+
+func (c *Cluster) deployDNS(ctx context.Context) error {
+	switch DNSProvider := c.DNS.Provider; DNSProvider {
+	case DefaultDNSProvider:
+		for _, dnsprovider := range DNSProviders {
+			if strings.EqualFold(dnsprovider, DefaultDNSProvider) {
+				continue
+			}
+			if err := c.removeDNSProvider(ctx, dnsprovider); err != nil {
```
moelsayed: Do you think it would be easier to do this once before the switch ?
```
       for _, dnsprovider := range DNSProviders {
		if strings.EqualFold(dnsprovider, c.DNS.Provider) {
			continue
		}
		if err := c.removeDNSProvider(ctx, dnsprovider); err != nil {
			return err
		}
	}
```

cluster/addons.go
```diff
@@ -377,3 +433,69 @@ func (c *Cluster) deployIngress(ctx context.Context) error {
...
+				return nil
```
moelsayed: return err

cluster/addons.go
```diff
@@ -377,3 +433,69 @@ func (c *Cluster) deployIngress(ctx context.Context) error {
...
+		}
+		log.Infof(ctx, "[dns] DNS provider %s deployed successfully", c.DNS.Provider)
+		return nil
+	case "coredns":
+		for _, dnsprovider := range DNSProviders {
+			if strings.EqualFold(dnsprovider, "coredns") {
+				continue
+			}
+			if err := c.removeDNSProvider(ctx, dnsprovider); err != nil {
+				return nil
```
moelsayed: return err

cluster/addons.go
moelsayed: also err for the other remove blocks

alena1108:
@superseb could you remove vendor update commit? looks like it's already been synced to the rke master

PR #1021, opened 2018-11-13
Add wait time after starting etcd snapshot container

galal-hussein:
https://github.com/rancher/rancher/issues/16553

### DIFF
--- a/services/etcd.go
+++ b/services/etcd.go
@@ -22,10 +22,11 @@ import (
 )
 
 const (
-	EtcdSnapshotPath = "/opt/rke/etcd-snapshots/"
-	EtcdRestorePath  = "/opt/rke/etcd-snapshots-restore/"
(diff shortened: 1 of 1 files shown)


services/etcd.go
```diff
@@ -308,6 +309,7 @@ func RunEtcdSnapshotSave(ctx context.Context, etcdHost *hosts.Host, prsMap map[s
 	if err != nil {
 		return err
 	}
+	time.Sleep(EtcdSnapshotWaitTime * time.Second)
```
alena1108: I feel like we might need to add validation for snapshot related parameters prior to starting a container. But at this point, waiting might be the least impactful way in terms of qa

galal-hussein: I agree, but adding a paremeter validation now will impact the way this feature has been tested, also if we decides that we shouldn't exit on snapshot error then it won't make any difference

PR #1022, opened 2018-11-13
Add wait time after starting etcd snapshot container for v0.1

galal-hussein:
https://github.com/rancher/rancher/issues/16553

### DIFF
--- a/services/etcd.go
+++ b/services/etcd.go
@@ -20,10 +20,11 @@ import (
 )
 
 const (
-	EtcdSnapshotPath = "/opt/rke/etcd-snapshots/"
-	EtcdRestorePath  = "/opt/rke/etcd-snapshots-restore/"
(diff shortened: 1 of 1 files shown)


services/etcd.go
```diff
@@ -281,6 +282,7 @@ func RunEtcdSnapshotSave(ctx context.Context, etcdHost *hosts.Host, prsMap map[s
 	if err := docker.DoRunContainer(ctx, etcdHost.DClient, imageCfg, hostCfg, EtcdSnapshotContainerName, etcdHost.Address, ETCDRole, prsMap); err != nil {
 		return err
 	}
+	time.Sleep(EtcdSnapshotWaitTime * time.Second)
```
alena1108: I feel like we might need to add validation for snapshot related parameters prior to starting a container. But at this point, waiting might be the least impactful way in terms of qa. Let me know what you think

PR #1025, opened 2018-11-13
Fix flannel startup in new cluster

annymsMthd:
In a new cluster the nodes will be NotReady until the network cni plugin is configured. Since the flannel daemonset is responsible for setting this up it needs to be able to schedule on nodes that are NotReady

Signed-off-by: Joshua Benjamin <joshuabnjmn@gmail.com>

### DIFF
--- a/templates/flannel.go
+++ b/templates/flannel.go
@@ -170,6 +170,9 @@ spec:
         operator: Exists
         effect: NoExecute
       {{- end }}
+      - key: node.kubernetes.io/not-ready
+        effect: NoSchedule
+        operator: Exists
       volumes:
         - name: run
           hostPath:


annymsMthd:
hmm I see the configuration for ClusterVersion 1.12 but when deploying this for version 1.12 I dont see those. Maybe there is something else going on here

annymsMthd:
rke version
`Server Version: version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.0", GitCommit:"0ed33881dc4355495f623c6f22e7dd0b7632b7c0", GitTreeState:"clean", BuildDate:"2018-09-27T16:55:41Z", GoVersion:"go1.10.4", Compiler:"gc", Platform:"linux/amd64"}`

perhaps `{{- if eq .ClusterVersion "v1.12" }}` should be `{{- if eq .ClusterVersion "v1.12.0" }}` ?

superseb:
We should probably just use 
```
      tolerations:
      - operator: Exists
        effect: NoSchedule
```

That's what upstream went with as well: https://raw.githubusercontent.com/coreos/flannel/master/Documentation/kube-flannel.yml

galal-hussein:
https://github.com/rancher/rancher/issues/18053

PR #1032, opened 2018-11-15
Return api and client certs to rotate certs

galal-hussein:


cmd/common.go
```diff
@@ -52,3 +55,39 @@ func setOptionsFromCLI(c *cli.Context, rkeConfig *v3.RancherKubernetesEngineConf
 
 	return rkeConfig, nil
 }
+
+func ClusterInit(ctx context.Context, rkeConfig *v3.RancherKubernetesEngineConfig, dialersOptions hosts.DialersOptions, flags cluster.ExternalFlags) error {
+	log.Infof(ctx, "Initiating Kubernetes cluster")
+	var fullState *cluster.FullState
+	stateFilePath := cluster.GetStateFilePath(flags.ClusterFilePath, flags.ConfigDir)
+	rkeFullState, _ := cluster.ReadStateFile(ctx, stateFilePath)
```
alena1108: we don't care about the error ReadStateFile returns?

galal-hussein: no not really, this is just to check if there is a legacy cluster, if not then it will create a new state file

PR #1036, opened 2018-11-19
added etcd backup and restore from s3

guangbochen:
PRs: 
1. rke-tools: https://github.com/rancher/rke-tools/pull/30

Feature Updates:
- rke-cli works same as the [rke doc](https://rancher.com/docs/rke/v0.1.x/en/etcd-snapshots/)
- Example of cluster.yml for using s3 backup:
```
services:
  etcd:
    snapshot: true
    # creation: 5m0s
    # retention: 24h
    backup_target:
      # interval_hours: 12  # Backup interval in hours, default to 12hrs
      # retention: 6  # Number of backups to keep, default to 6 = (12*6) = 72 hrs
      s3backupconfig:
        endpoint: s3.amazonaws.com
        access_key: xxxxx
        secret_key: xxxx
        bucket_name: my-bucketname
```

- additional cli commands for s3 includes:
```
NAME:
   rke etcd snapshot-save - Take snapshot on all etcd hosts

USAGE:
   rke etcd snapshot-save [command options] [arguments...]

OPTIONS:
   --name value             Specify snapshot name
   --config value           Specify an alternate cluster YAML file (default: "cluster.yml") [$RKE_CONFIG]
   --s3                     Enabled backup to s3, set true or false
   --s3-endpoint value      Specify s3 endpoint url (default: "s3.amazonaws.com")
   --access-key value        Specify s3 accessKey
   --secret-key value        Specify s3 secretKey
   --bucket-name value       Specify s3 bucket name
   --region value      Specify the s3 bucket location (optional)
   --ssh-agent-auth         Use SSH Agent Auth defined by SSH_AUTH_SOCK
   --ignore-docker-version  Disable Docker version check
```


cluster/etcd.go
```diff
@@ -13,14 +13,21 @@ import (
 
 func (c *Cluster) SnapshotEtcd(ctx context.Context, snapshotName string) error {
 	for _, host := range c.EtcdHosts {
-		if err := services.RunEtcdSnapshotSave(ctx, host, c.PrivateRegistriesMap, c.SystemImages.Alpine, c.Services.Etcd.Creation, c.Services.Etcd.Retention, snapshotName, true); err != nil {
+		if err := services.RunEtcdSnapshotSave(ctx, host, c.PrivateRegistriesMap, c.SystemImages.Alpine, c.Services.Etcd.Creation, c.Services.Etcd.Retention, snapshotName, true, c.Services.Etcd.BackupBackend); err != nil {
```
galal-hussein: I think we can just send the whole snapshot etcd structure instead of sending each option separately

cluster/etcd.go
```diff
@@ -13,14 +13,21 @@ import (
...
 			return err
 		}
 	}
 	return nil
 }
 
 func (c *Cluster) RestoreEtcdSnapshot(ctx context.Context, snapshotPath string) error {
+	// get etcd snapshots from s3 via rke-tools container
+	for _, host := range c.EtcdHosts {
+		if err := services.DownloadEtcdSnapshot(ctx, host, c.PrivateRegistriesMap, c.SystemImages.Alpine, c.Services.Etcd.Creation, c.Services.Etcd.Retention, snapshotPath, true, c.Services.Etcd.BackupBackend); err != nil {
```
galal-hussein: Same thing here, we send the snapshot structure

services/etcd.go
```diff
@@ -318,6 +324,53 @@ func RunEtcdSnapshotSave(ctx context.Context, etcdHost *hosts.Host, prsMap map[s
...
+			"/opt/rke-tools/rke-etcd-backup",
+			"etcd-backup",
+			"download",
+			"--name", name,
+			"--endpoints=" + etcdHost.InternalAddress + ":2379",
+		},
+		Image: etcdSnapshotImage,
+	}
+
+	log.Infof(ctx, "invoking restore s3, bucketName=%s, endpoint:%s", s3Backend.BucketName, s3Backend.Endpoint)
```
alena1108: this code looks similar to the code block on lines 280-286, any chance we can extract it and re-use?

services/etcd.go
```diff
@@ -373,3 +414,14 @@ func RestoreEtcdSnapshot(ctx context.Context, etcdHost *hosts.Host, prsMap map[s
 	}
 	return docker.RemoveContainer(ctx, etcdHost.DClient, etcdHost.Address, EtcdRestoreContainerName)
 }
+
+func configS3BackupImgCmd(ctx context.Context, imageCfg *container.Config, s3Backend *v3.S3BackupBackend) *container.Config {
+	log.Infof(ctx, "invoking s3 backup server cmd config, bucketName:%s, endpoint:%s", s3Backend.BucketName, s3Backend.Endpoint)
+	imageCfg.Cmd = append(imageCfg.Cmd, "--s3-backup=true")
```
moelsayed: I think it would be easier if you group these in a single slice first:
```
cmd := []string{
"--s3-backup=true",
--s3-endpoint="+s3Backend.Endpoint,
.
.
"--s3-region="+s3Backend.Region,
}
return append(imageCfg.Cmd, cmd...)
```

cmd/etcd.go
```diff
@@ -14,20 +14,46 @@ import (
...
 			Usage:  "Specify an alternate cluster YAML file",
 			Value:  pki.ClusterConfig,
 			EnvVar: "RKE_CONFIG",
 		},
+		cli.BoolFlag{
+			Name:  "s3",
+			Usage: "Enabled backup to s3, set true or false",
+		},
+		cli.StringFlag{
+			Name:  "endpoint",
```
moelsayed: Do you think we should prepend the s3 related options with `s3` ?

cmd/etcd.go
```diff
@@ -14,20 +14,46 @@ import (
...
+			Usage: "Specify s3 endpoint url",
+			Value: s3Endpoint,
+		},
+		cli.StringFlag{
+			Name:  "accessKey",
```
moelsayed: I am not sure how the cli library will handle camelCase for flags, maybe we should go with a dash `access-key` ?

cmd/etcd.go
guangbochen: @alena1108 and you have diffrent opion of the `endpoint` name, but I would prefer to go with the name of `s3-endpoint`.

guangbochen:
@moelsayed all resloved, thanks.

cmd/common.go
```diff
@@ -91,3 +98,20 @@ func ClusterInit(ctx context.Context, rkeConfig *v3.RancherKubernetesEngineConfi
 	}
 	return rkeState.WriteStateFile(ctx, stateFilePath)
 }
+
+func setS3OptionsFromCLI(c *cli.Context) *v3.S3BackupTarget {
```
moelsayed: I think we still need to pass the access and secret keys here, if they are set in the command line, right ?

services/etcd.go
```diff
@@ -318,6 +318,47 @@ func RunEtcdSnapshotSave(ctx context.Context, etcdHost *hosts.Host, prsMap map[s
 	return nil
 }
 
+func DownloadEtcdSnapshot(ctx context.Context, etcdHost *hosts.Host, prsMap map[string]v3.PrivateRegistry, etcdSnapshotImage string, name string, once bool, es v3.ETCDService) error {
+	log.Infof(ctx, "[etcd] Get snapshot [%s] on host [%s]", name, etcdHost.Address)
+	s3Backend := es.BackupTarget.S3BackupTarget
+	if len(s3Backend.Endpoint) == 0 || len(s3Backend.AccessKey) == 0 || len(s3Backend.SecretKey) == 0 || len(s3Backend.BucketName) == 0 {
```
moelsayed: This looks like it would break, since you don't pass the access/secret key anymore. Also, this would break if you use an IAM role.

services/etcd.go
```diff
@@ -318,6 +318,47 @@ func RunEtcdSnapshotSave(ctx context.Context, etcdHost *hosts.Host, prsMap map[s
...
+		log.Warnf(ctx, "[etcd] Failed to get snapshot [%s] from s3 on host [%s], invalid s3 configurations", name, etcdHost.Address)
+		return nil
```
moelsayed: Why return nil instead of err ?

cmd/common.go
guangbochen: yep, I have brought it back.

services/etcd.go
guangbochen: resolved.

services/etcd.go
guangbochen: addressed.

services/etcd.go
```diff
@@ -318,6 +318,46 @@ func RunEtcdSnapshotSave(ctx context.Context, etcdHost *hosts.Host, prsMap map[s
...
 
+func DownloadEtcdSnapshot(ctx context.Context, etcdHost *hosts.Host, prsMap map[string]v3.PrivateRegistry, etcdSnapshotImage string, name string, once bool, es v3.ETCDService) error {
+	log.Infof(ctx, "[etcd] Get snapshot [%s] on host [%s]", name, etcdHost.Address)
+	s3Backend := es.BackupTarget.S3BackupTarget
+	if len(s3Backend.Endpoint) == 0 || len(s3Backend.BucketName) == 0 {
+		return fmt.Errorf("failed to get snapshot [%s] from s3 on host [%s], invalid s3 configurations", name, etcdHost.Address)
+	}
+	imageCfg := &container.Config{
+		Cmd: []string{
+			"/opt/rke-tools/rke-etcd-backup",
```
moelsayed: You need to handle credentials here, as well.

services/etcd.go
```diff
@@ -318,6 +318,46 @@ func RunEtcdSnapshotSave(ctx context.Context, etcdHost *hosts.Host, prsMap map[s
...
+			"etcd-backup",
+			"download",
+			"--name", name,
+			"--endpoints=" + etcdHost.InternalAddress + ":2379",
```
moelsayed: Why do you need the endpoints for download ?

cluster/etcd.go
```diff
@@ -21,6 +21,16 @@ func (c *Cluster) RestoreEtcdSnapshot(ctx context.Context, snapshotPath string)
 	if isEqual := c.etcdSnapshotChecksum(ctx, snapshotPath); !isEqual {
 		return fmt.Errorf("etcd snapshots are not consistent")
 	}
+
+	// get etcd snapshots from s3 if backup backend server is set
```
moelsayed: This needs to _before_ the checksum check.

services/etcd.go
guangbochen: removed.

cluster/etcd.go
guangbochen: resolved.

cluster/defaults.go
```diff
@@ -179,6 +181,15 @@ func (c *Cluster) setClusterServicesDefaults() {
 	if _, ok := c.Services.Etcd.ExtraArgs[DefaultEtcdHeartbeatIntervalName]; !ok {
 		c.Services.Etcd.ExtraArgs[DefaultEtcdHeartbeatIntervalName] = DefaultEtcdHeartbeatIntervalValue
 	}
+
+	if c.Services.Etcd.BackupConfig != nil {
+		if c.Services.Etcd.BackupConfig.IntervalHours == 0 {
+			c.Services.Etcd.BackupConfig.IntervalHours = DefaultEtcdBackupConfigIntervalHours
```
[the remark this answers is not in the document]
guangbochen: if the `BackupConfig` is not set, we will use `Etcd`'s values, and we already have the default value set for both `c.Services.Etcd.Creation` and `c.Services.Etcd.Retention` at line 169.

Issue #873, opened 2018-08-22
Allow custom CNI deployments by using "network: none"


dnoland1:
Currently RKE supports Flannel, Calico, Canal, and Weave as network plugins. For users who have their own (custom) CNI solution, would be good to have a none or "opt-out" option. Could look something like this in the cluster.yml:

```
network:
    plugin: none
```
or
```
network:
    plugin: custom
```

dnoland1:
@alena1108 Here's a little more information on the use case. Before RKE is run, a custom CNI configuration file will be placed in `/etc/cni/net.d/` and a custom CNI binary in `/opt/cni/bin/` on the host. If needed or if it's not the default, can add this into RKE's cluster.yml:
```
services:
  kubelet:
    extra_args:
      cni-conf-dir: "/etc/cni/net.d"
      cni-bin-dir: "/opt/cni/bin"
```
May need bind-args as well to map the above directories on the host into kubelet's container.
So I believe if the above is in place and RKE does not deploy it's own network plugin, pod-to-pod networking should work, including DNS.

alena1108:
Although the feature is added earlier, it can be tested as a part of `v0.2.0-rc3` rke

dnoland1:
Tested with `rke version v0.2.0-rc5`. Added this to my cluster.yml:
```
network:
  plugin: none
```
Confirmed that no CNI plugin was deployed. Kubelet logs correctly showed me:
```
W0207 01:40:44.457124    8182 cni.go:203] Unable to update cni config: No networks found in /etc/cni/net.d
E0207 01:40:44.457291    8182 kubelet.go:2192] Container runtime network not ready: NetworkReady=false reason:NetworkPluginNotReady message:docker: network plugin is not ready: cni config uninitialized
```
Which was expected since I didn't manually install a custom CNI.

PR #1047, opened 2018-11-30
Allow creating Kubernetes clusters without a cluster network

Oats87:
Create an RKE config file like:
```
nodes:
    - address: 192.168.100.129
      user: docker
      role:
        - controlplane
        - etcd
    - address: 192.168.100.130
      user: docker
      role:
        - worker
    - address: 192.168.100.131
      user: docker
      role:
        - worker
network:
  plugin: none
```

https://github.com/rancher/rke/issues/873

### DIFF
--- a/cluster/network.go
+++ b/cluster/network.go
@@ -39,6 +39,8 @@ const (
 	ProtocolTCP = "TCP"
 	ProtocolUDP = "UDP"
 
+	NoNetworkPlugin = "none"
+
 	FlannelNetworkPlugin = "flannel"
 	FlannelIface         = "flannel_iface"
 	FlannelBackendType   = "flannel_backend_type"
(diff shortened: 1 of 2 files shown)


galal-hussein:
@Oats87 in case the user added a `none` network plugin do we expect for the subsequent addons to be deployed successfully, also are the nodes will be in a READY state?

Oats87:
> @Oats87 in case the user added a `none` network plugin do we expect for the subsequent addons to be deployed successfully, also are the nodes will be in a READY state?

Subsequent addon creation should be successful. For 1.12, nodes will not be ready until CNI is initialized, but the workaround for this if you want ready nodes at the end of RKE up would be to define your own CNI in the user addons section and allow RKE to initialize the cluster network

PR #1056, opened 2018-12-07
CIS Benchmark recommended options

jgreat:
First in a set of recommended options for CIS Kubernetes Benchmark guidelines.

Command Options: (sorry reordered the lists alphabetically to make it easier to find options)
kubelet:
* streaming-connection-idle-timeout: "30m"
* make-iptables-util-chains: "true"
* event-qps: "0"

kubeapi:
* anonymous-auth: "false"
* unset insecure-bind-address
* profiling: "false"
* repair-malformed-updates: false
* service-account-lookup: true
* enable-admission-plugins: ...NodeRestriction...

kube-scheduler:
* profiling: "false"
* address: "127.0.0.1" ("0.0.0.0" when DinD)

kube-controller-manager:
* terminated-pod-gc-threshold: "1000"
* profiling: "false"
* address: "127.0.0.1" ("0.0.0.0" when DinD)

kube-proxy:
* healthz-bind-address: "127.0.0.1" ("0.0.0.0" when DinD)

---

Moved cadvisor-port out of global default into k8s_defaults for various versions. 

---

Modified admission-plugins to work as a []string and then write out to a comma separated string for easier add/remove admissions plugins.

---

Created `scripts/integration` test. Grabs list of the latest "major.minor" k8s versions from rke and creates a DinD instance for each version (1.8 -> 1.12). If build succeeds and rke exits 0, tests pass.  If not it fails. Either way cleans up after its self when tests are done.    

This test does do all 5 instances simultaneously, so its resource heavy (especially network since it has to pull all the hyperkube images).  I think its valuable, but if we don't want it in the normal "ci" set I can pull it out.

---

Fixed typo `kubeletDcokerConfig`. It was bugging me :)

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -37,6 +37,7 @@ type Cluster struct {
 	ClusterDomain                    string
 	ClusterCIDR                      string
 	ClusterDNSServer                 string
+	DinD                             bool
 	DockerDialerFactory              hosts.DialerFactory
 	EtcdHosts                        []*hosts.Host
 	EtcdReadyHosts                   []*hosts.Host
@@ -150,6 +151,7 @@ func InitClusterObject(ctx context.Context, rkeConfig *v3.RancherKubernetesEngin
 		RancherKubernetesEngineConfig: *rkeConfig,
 		ConfigPath:                    flags.ClusterFilePath,
 		ConfigDir:                     flags.ConfigDir,
+		DinD:                          flags.DinD,
 		StateFilePath:                 GetStateFilePath(flags.ClusterFilePath, flags.ConfigDir),
 		PrivateRegistriesMap:          make(map[string]v3.PrivateRegistry),
 	}
--- a/cluster/defaults.go
+++ b/cluster/defaults.go
@@ -54,6 +54,7 @@ const (
 type ExternalFlags struct {
 	ConfigDir        string
 	ClusterFilePath  string
+	DinD             bool
 	DisablePortCheck bool
 	Local            bool
 	UpdateOnly       bool
--- a/cluster/plan.go
+++ b/cluster/plan.go
@@ -117,36 +117,48 @@ func (c *Cluster) BuildKubeAPIProcess(prefixPath string) v3.Process {
 		c.getRKEToolsEntryPoint(),
 		"kube-apiserver",
 	}
-
+	baseEnabledAdmissionPlugins := []string{
+		"DefaultStorageClass",
+		"DefaultTolerationSeconds",
+		"LimitRanger",
+		"NamespaceLifecycle",
+		"NodeRestriction",
+		"PersistentVolumeLabel",
+		"ResourceQuota",
+		"ServiceAccount",
+	}
 	CommandArgs := map[string]string{
-		"insecure-bind-address":              "127.0.0.1",
-		"bind-address":                       "0.0.0.0",
-		"insecure-port":                      "0",
-		"secure-port":                        "6443",
-		"cloud-provider":                     c.CloudProvider.Name,
 		"allow-privileged":                   "true",
-		"kubelet-preferred-address-types":    "InternalIP,ExternalIP,Hostname",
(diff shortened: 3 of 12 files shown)


jgreat:
https://github.com/rancher/types/pull/648

cluster/cluster.go
```diff
@@ -154,6 +155,7 @@ func InitClusterObject(ctx context.Context, rkeConfig *v3.RancherKubernetesEngin
 		ConfigDir:                     flags.ConfigDir,
 		StateFilePath:                 GetStateFilePath(flags.ClusterFilePath, flags.ConfigDir),
 		PrivateRegistriesMap:          make(map[string]v3.PrivateRegistry),
+		CLIFlags:                      &flags,
```
galal-hussein: can we just include the dind flag instead of passing all flags to the cluster object

cluster/plan.go
```diff
@@ -236,21 +270,27 @@ func (c *Cluster) BuildKubeControllerProcess(prefixPath string) v3.Process {
...
-		"v":                                "2",
-		"allocate-node-cidrs":              "true",
-		"cluster-cidr":                     c.ClusterCIDR,
-		"service-cluster-ip-range":         c.Services.KubeController.ServiceClusterIPRange,
-		"service-account-private-key-file": pki.GetKeyPath(pki.ServiceAccountTokenKeyName),
+		"profiling":                        "false",
 		"root-ca-file":                     pki.GetCertPath(pki.CACertName),
+		"service-account-private-key-file": pki.GetKeyPath(pki.ServiceAccountTokenKeyName),
+		"service-cluster-ip-range":         c.Services.KubeController.ServiceClusterIPRange,
+		"terminated-pod-gc-threshold":      "1000",
```
galal-hussein: can you explain the main reason why is this 1000 instead of the default 12500

cluster/defaults.go
```diff
@@ -293,11 +294,12 @@ func (c *Cluster) setCloudProvider() error {
 	return nil
 }
 
-func GetExternalFlags(local, updateOnly, disablePortCheck bool, configDir, clusterFilePath string) ExternalFlags {
+func GetExternalFlags(local, updateOnly, disablePortCheck, dind bool, configDir, clusterFilePath string) ExternalFlags {
```
galal-hussein: Can we avoid updating this function and just enable dind in the dind function


`externalFlags.dind = true`

galal-hussein: this same function is used in kontainer engine and you will have to update the kontainer engine as well

cluster/plan.go
jgreat: Here's the CIS Benchmark justification:

###1.3.1 Ensure that the --terminated-pod-gc-threshold argument is set as appropriate (Scored)

####Description:
Activate garbage collector on pod termination, as appropriate.

####Rationale:
Garbage collection is important to ensure sufficient resource availability and avoiding
degraded performance and availability. In the worst case, the system might crash or just be
unusable for a long period of time. The current setting for garbage collection is 12,500
terminated pods which might be too high for your system to sustain. Based on your system
resources and tests, choose an appropriate threshold value to activate garbage collection.

####References:
1. https://kubernetes.io/docs/admin/kube-controller-manager/
2. https://github.com/kubernetes/kubernetes/issues/28484

####CIS Controls:
Version 6

14 Controlled Access Based on the Need to Know
Controlled Access Based on the Need to Know

cluster/defaults.go
jgreat: I'd rather not use this function, but all the other places that pass if flags use this.  How do we pass in flags with out this function?

(personally I don't understand why this function exists. If you have to pass in all the values to it, why not just pass an ExternalFlags object, the effect is the same and you don't have to maintain the function)

cluster/cluster.go
galal-hussein: thats true, but for consistency then we should remove other flags like configDir and configPath from the cluster object definition, either way are correct, its up to you.

PR #1064, opened 2018-12-14
WIP - CIS Benchmark - Secrets encryption config

jgreat:
Add support for secrets encryption config.

* Default to generating an encryption config using recommended `asecbc` with a random key.
* User can specify their own provider.
* Warn and require acknowledgement (I know what I'm doing) if user tries to change or provide custom provider after cluster already has a provider set up. If you mess your config up: at best the kube-apiserver won't start, at worst you can lock yourself out of your secrets. 
* add `--force` flag to `up` to bypass warning prompt.
* Restart `controlplane` services if config is changed. 
* Replace secrets to apply new encryption if config changes, or when config is applied to a cluster built with previous version of RKE.

---

k8s/secrets.go
* Add function to get all secrets in a cluster.
* Add function to replace all secrets in a cluster.

--- 

templates/encryption.go
* add templates for 1.8 -> 1.13

---

cluster/write-file.go
* more generic way to write and permission files to a host
* will create timestamped back up of files it overwrites (just in case)

---

cmd/up.go
* add `--force` - I know what I'm doing - flag 
* add in triggers for creating the generated config and reconcile secretes to apply encryption.

---

cluster/defaults.go
cluster/state.go
cmd/remove.go
cmd/etcd.go
cmd/cert.go
* Add support for `--force` flag 

---

### Testing: 

* New cluster generated config
* New cluster provided config
* Update cluster with provided config (rotate keys)
* Previous RKE cluster is now encrypted.

#### General procedure:

1. Create Cluster with previous version of RKE
2. Create secret
```
kubectl create secret generic secret1 -n default --from-literal=mykey=mydata
secret/secret1 created
```
3. Confirm secret
```
kubectl get secret secret1 -o jsonpath --template='{ .data.mykey }' | base64 -d

mydata%
```
4. Update cluster with new RKE (this should generate, apply config and recreate all the secrets so they are encrypted)
5. Confirm you can still read secret
 ```
kubectl get secret secret1 -o jsonpath --template='{ .data.mykey }' | base64 -d

mydata%
```
5. Exec into etcd container and confirm encryption has been applied. Look for `enc:aescbc:v1:key1`. This will tell you the encryption provider and which key is being used for the secret.
```
etcd etcdctl get /registry/secrets/default/secret1 | hexdump -C

00000000  2f 72 65 67 69 73 74 72  79 2f 73 65 63 72 65 74  |/registry/secret|
00000010  73 2f 64 65 66 61 75 6c  74 2f 73 65 63 72 65 74  |s/default/secret|
00000020  31 0a 6b 38 73 3a 65 6e  63 3a 61 65 73 63 62 63  |1.k8s:enc:aescbc|
00000030  3a 76 31 3a 6b 65 79 31  3a 37 01 ba 0c 82 30 f8  |:v1:key1:7....0.|
00000040  9a 57 3a ff 7f 5d a9 49  9d 4b 84 09 c2 43 b4 4a  |.W:..].I.K...C.J|
00000050  4d 93 d3 43 98 bf 48 62  21 f1 c8 1e c6 68 bc 05  |M..C..Hb!....h..|
00000060  41 d6 a0 11 3a 39 e9 55  c4 8d 3a 20 84 44 94 99  |A...:9.U..: .D..|
00000070  63 bf ea 3c 44 db ae 0f  18 14 d9 6f d2 84 37 12  |c..<D......o..7.|
00000080  26 72 35 96 ab 38 22 6f  88 c1 12 5a 66 90 87 3d  |&r5..8"o...Zf..=|
00000090  96 dd a7 cb b8 e7 8b 3d  7e 4a 8f c6 b1 e6 5f ac  |.......=~J...._.|
000000a0  73 3c de 44 a6 d0 ac fe  0d d7 aa a0 d9 5d c9 79  |s<.D.........].y|
000000b0  ec ef 12 3b d4 d6 a0 d9  91 f0 ab 73 0b 3f 85 e3  |...;.......s.?..|
000000c0  f6 e2 d5 54 b7 bc c3 ad  aa 0a                    |...T......|
000000ca
```

#### Example provided config `cluster.yml`
Secret is 32 byte base64 encoded string.
```yaml
 services:
  kube-api:
    secrets_encryption_config: |-
      apiVersion: v1
      kind: EncryptionConfig
      resources:
        - resources:
          - secrets
          providers:
          - aescbc:
              keys:
              - name: key1
                secret: qsJXfHuNtQV2othYw6ihJLGj9U5XPi1ThQYcS1XJJ5A=
          - identity: {}
```

#### Example rotating keys
Previous key must be included in the provider list so encrypted secrets can still be read.
```yaml
 services:
  kube-api:
    secrets_encryption_config: |-
      apiVersion: v1
      kind: EncryptionConfig
      resources:
        - resources:
          - secrets
          providers:
          - aescbc:
              keys:
              - name: key2
                secret: UIt8QL3hRjrWZJWV5hZsNWFNFtQyKZovKxESplYiFq0=
              - name: key1
                secret: qsJXfHuNtQV2othYw6ihJLGj9U5XPi1ThQYcS1XJJ5A=
          - identity: {}
```

### DIFF
--- a/cluster/encryption.go
+++ b/cluster/encryption.go
@@ -1,0 +1,152 @@
+package cluster
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"crypto/rand"
+	"encoding/base64"
+	"fmt"
+	"os"
+	"strings"
+	"text/template"
+
+	"github.com/rancher/rke/k8s"
+	"github.com/rancher/rke/services"
+	"github.com/rancher/rke/templates"
+	"github.com/sirupsen/logrus"
+)
+
+// DeploySecretsEncryptionConfig -  Deploy encryption.yaml to ControlPlane hosts.
+func (c *Cluster) DeploySecretsEncryptionConfig(ctx context.Context, currentCluster *Cluster) error {
+	if currentCluster == nil {
+		currentCluster = &Cluster{}
+	}
+
+	if c.Services.KubeAPI.SecretsEncryptionConfig == currentCluster.Services.KubeAPI.SecretsEncryptionConfig {
+		logrus.Debug("[controlplane] Secrets encryption config matches state. No changes needed.")
+		return nil
+	}
+
+	for _, host := range c.ControlPlaneHosts {
+		err := c.WriteFile(ctx, host, "controlplane", "/etc/kubernetes/encryption.yaml", "0600", c.Services.KubeAPI.SecretsEncryptionConfig)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// ReconcileSecretsEncryptionConfig - Restart controlplane and replace all secrets in the cluster when the encryption config changes.
+func (c *Cluster) ReconcileSecretsEncryptionConfig(ctx context.Context, currentCluster *Cluster) error {
+	if currentCluster != nil {
+		if c.Services.KubeAPI.SecretsEncryptionConfig == currentCluster.Services.KubeAPI.SecretsEncryptionConfig {
+			logrus.Debug("[reconcile] Secrets encryption config matches. No changes needed.")
+			return nil
+		}
+
+		logrus.Info("[reconcile] New secrets encryption config. Updating existing secrets to apply encryption.")
+
+		err := services.RestartControlPlane(ctx, c.ControlPlaneHosts)
+		if err != nil {
+			return err
+		}
+
+		kubeClient, err := k8s.NewClient(c.LocalKubeConfigPath, c.K8sWrapTransport)
+		if err != nil {
+			return fmt.Errorf("Failed to initialize new kubernetes client: %v", err)
+		}
+
+		secrets, err := k8s.GetAllSecrets(kubeClient)
+		if err != nil {
+			return err
+		}
+
+		for _, secret := range secrets {
+			err = k8s.UpdateSecretObject(kubeClient, &secret)
+			if err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// Need some logic to keep users from shooting themselves in the foot.
+// If you rotate keys without the old key in the list you will lock yourself out of the existing secrets.
+// Ways people will break this:
+// - Add a config after a cluster is up with a generated config
+// - Change a config without putting the old key in the list before re-create of secrets.
+
+// SetSecretsEncryptionConfig - Set a default encryption config and secret if not provided.
+func (c *Cluster) SetSecretsEncryptionConfig(currentCluster *Cluster) error {
+	if currentCluster == nil {
+		currentCluster = &Cluster{}
+	}
+	// cluster.yaml not defined or populated
+	if len(c.Services.KubeAPI.SecretsEncryptionConfig) == 0 {
+		// Generate from template if not populated in current config.
+		if len(currentCluster.Services.KubeAPI.SecretsEncryptionConfig) == 0 {
+			clusterMajorVersion := getTagMajorVersion(c.Version)
+			encSecretKey := make([]byte, 32)
+			_, err := rand.Read(encSecretKey)
+			if err != nil {
+				return err
+			}
+			encConfig := &bytes.Buffer{}
+			secret := struct {
+				Secret string
+			}{
+				Secret: base64.StdEncoding.EncodeToString(encSecretKey),
+			}
+
+			_, ok := templates.EncryptionConfigurationTemplate[clusterMajorVersion]
+			if !ok {
+				return fmt.Errorf("Unsupported Kubernetes version, no encryption config template available: %s", c.Version)
+			}
+			encTemplate := template.New("encryption.yaml")
+			encTemplate, err = encTemplate.Parse(templates.EncryptionConfigurationTemplate[clusterMajorVersion])
+			if err != nil {
+				return err
+			}
+			err = encTemplate.Execute(encConfig, secret)
+			if err != nil {
+				return err
+			}
+
+			c.Services.KubeAPI.SecretsEncryptionConfig = encConfig.String()
+		} else {
+			// set previous cluster config (c is populated with cluster.yml, so currently empty)
+			c.Services.KubeAPI.SecretsEncryptionConfig = currentCluster.Services.KubeAPI.SecretsEncryptionConfig
+		}
+		// cluster.yml is defined.
+	} else {
+		// Should be a new cluster - just let it pass
+		if len(currentCluster.Services.KubeAPI.SecretsEncryptionConfig) == 0 {
+			return nil
+		}
+		// cluster.yml equals current config
+		if c.Services.KubeAPI.SecretsEncryptionConfig == currentCluster.Services.KubeAPI.SecretsEncryptionConfig {
+			return nil
+		}
+		// cluster.yml is different than current config
+		// Are you sure you know what you are doing?
+		if !c.Force {
+			reader := bufio.NewReader(os.Stdin)
+			logrus.Warn("A new secrets encryption configuration has been detected!")
+			logrus.Warn("Improper configuration may result in kube-apiserver failing to start and permanently losing access to stored secrets.")
+			logrus.Warn("For details on using or changing a custom encryption configuration see:")
+			logrus.Warn("https://rancher.com/docs/rke/v0.1.x/en/config-options/services/kube-api")
+			fmt.Print("Are you sure you want to continue [y/n]: ")
+			input, err := reader.ReadString('\n')
+			input = strings.TrimSpace(input)
+			if err != nil {
+				return err
+			}
+			if input != "y" && input != "Y" {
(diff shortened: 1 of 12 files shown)


cluster/plan.go
```diff
@@ -328,29 +368,31 @@ func (c *Cluster) BuildKubeletProcess(host *hosts.Host, prefixPath string) v3.Pr
...
-		"allow-privileged":             "true",
-		"cloud-provider":               c.CloudProvider.Name,
-		"kubeconfig":                   pki.GetConfigPath(pki.KubeNodeCertName),
-		"client-ca-file":               pki.GetCertPath(pki.CACertName),
-		"anonymous-auth":               "false",
-		"volume-plugin-dir":            "/var/lib/kubelet/volumeplugins",
-		"fail-swap-on":                 strconv.FormatBool(c.Services.Kubelet.FailSwapOn),
-		"root-dir":                     path.Join(prefixPath, "/var/lib/kubelet"),
-		"authentication-token-webhook": "true",
+		"address":                           "0.0.0.0",
```
galal-hussein: is it possible to avoid rearranging the fields and do it in a separate PR, and add only the modifications that needed to be done

cluster/encryption.go
```diff
@@ -0,0 +1,152 @@
...
+		}
+		// cluster.yml is different than current config
+		// Are you sure you know what you are doing?
+		if !c.Force {
+			reader := bufio.NewReader(os.Stdin)
+			logrus.Warn("A new secrets encryption configuration has been detected!")
+			logrus.Warn("Improper configuration may result in kube-apiserver failing to start and permanently losing access to stored secrets.")
+			logrus.Warn("For details on using or changing a custom encryption configuration see:")
+			logrus.Warn("https://rancher.com/docs/rke/v0.1.x/en/config-options/services/kube-api")
+			fmt.Print("Are you sure you want to continue [y/n]: ")
```
moelsayed: Can we extract this into it's own function ?

cluster/write-file.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	b64Content := base64.StdEncoding.EncodeToString([]byte(content))
+	pathDir := filepath.Dir(path)
+	imageCfg := &container.Config{
+		Image: c.SystemImages.Alpine,
+		Cmd: []string{
+			"sh",
+			"-c",
+			fmt.Sprintf("cp -f %s %s-$(date -Iseconds -u); echo \"%s\" | base64 -d > %s && chmod %s %s", path, path, b64Content, path, mode, path),
```
moelsayed: We have been thinking about using docker cp API to write files to nodes. It should be faster, much easier to report errors, and you don't need to start() the container, you just need to create() it. Do you think it would make sense to start using that here?

cmd/up.go
```diff
@@ -210,6 +225,11 @@ func ClusterUp(ctx context.Context, dialersOptions hosts.DialersOptions, flags c
 		return APIURL, caCrt, clientCert, clientKey, nil, err
 	}
 
+	err = kubeCluster.ReconcileSecretsEncryptionConfig(ctx, currentCluster)
```
moelsayed: This should be done in cluster/reconcile.go:ReconcileCluster().

cluster/encryption.go
```diff
@@ -0,0 +1,152 @@
...
+		if err != nil {
+			return fmt.Errorf("Failed to initialize new kubernetes client: %v", err)
+		}
+
+		secrets, err := k8s.GetAllSecrets(kubeClient)
+		if err != nil {
+			return err
+		}
+
+		for _, secret := range secrets {
```
moelsayed: This should be done in a concurrent worker loop like this: https://github.com/rancher/rke/blob/master/cluster/cluster.go#L366-L384

cluster/encryption.go
```diff
@@ -0,0 +1,152 @@
...
+// SetSecretsEncryptionConfig - Set a default encryption config and secret if not provided.
+func (c *Cluster) SetSecretsEncryptionConfig(currentCluster *Cluster) error {
+	if currentCluster == nil {
+		currentCluster = &Cluster{}
+	}
+	// cluster.yaml not defined or populated
+	if len(c.Services.KubeAPI.SecretsEncryptionConfig) == 0 {
+		// Generate from template if not populated in current config.
+		if len(currentCluster.Services.KubeAPI.SecretsEncryptionConfig) == 0 {
+			clusterMajorVersion := getTagMajorVersion(c.Version)
```
galal-hussein: getting the encryption config should be separated in its own method like setting the cloud config file and the result should be one of the cluster's object field that can be used later, for example https://github.com/rancher/rke/blob/master/cluster/cluster.go#L179-L181

cluster/encryption.go
```diff
@@ -0,0 +1,152 @@
...
+			encTemplate, err = encTemplate.Parse(templates.EncryptionConfigurationTemplate[clusterMajorVersion])
+			if err != nil {
+				return err
+			}
+			err = encTemplate.Execute(encConfig, secret)
+			if err != nil {
+				return err
+			}
+
+			c.Services.KubeAPI.SecretsEncryptionConfig = encConfig.String()
```
galal-hussein: Setting the service fields should go in cluster/plan.go in generating the kubeapi plan, it shouldn't be part of this function

cluster/write-file.go
```diff
@@ -0,0 +1,68 @@
...
+	err := c.doWriteFile(ctx, host, role, path, mode, content)
+	if err != nil {
+		return fmt.Errorf("[%s] Failed to file %s on node [%s]: %v", role, path, host.Address, err)
+	}
+	logrus.Debugf("[%s] Successfully deployed %s on node [%s]", role, path, host.Address)
+
+	return nil
+}
+
+func (c *Cluster) doWriteFile(ctx context.Context, host *hosts.Host, role string, path string, mode string, content string) error {
```
galal-hussein: there is an already function that does this https://github.com/rancher/rke/blob/master/cluster/file-deployer.go#L32, i think this should be reused

k8s/secret.go
```diff
@@ -11,6 +11,25 @@ func GetSecret(k8sClient *kubernetes.Clientset, secretName string) (*v1.Secret,
 	return k8sClient.CoreV1().Secrets(metav1.NamespaceSystem).Get(secretName, metav1.GetOptions{})
 }
 
+// GetAllSecrets - Iterate through namespaces and return a slice of secrets.
+func GetAllSecrets(k8sClient *kubernetes.Clientset) ([]v1.Secret, error) {
+	secrets := []v1.Secret{}
+	namespaces, err := k8sClient.CoreV1().Namespaces().List(metav1.ListOptions{})
```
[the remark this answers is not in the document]
jgreat: I see that this works, but I'm trying to figure out how I could have known if namespace is empty it would return all the namespaces.

cluster/write-file.go
jgreat: That's an interesting idea. I think the one advantage to this method is that I'm creating a backup of the previous file before writing over it. I'm not sure you can could you do that with docker cp.

cluster/write-file.go
jgreat: Do you mind if I enhance file-deployer? 
I think its cleaner to use base64 encoding (so no escaping issues) and create a backup copy before writing over an existing file. Also most of these files should probably not be 0644 (or at least not world readable) so mode should be configurable.

cluster/encryption.go
jgreat: If you have potentially 1000s of secrets, should they be done concurrently. Is there a potential to hit rate limits? Enabling EventRateLimiter is on the list of recommended security options.

jgreat: I'll admit that I don't have a lot of experience with concurrency. I'm confused by the code you reference.  

It looks to me like each thread is going to iterate through the entire hostQueue and its going to repeat the creation of a thread and iterating through the hostQueue SyncWorkers(10) times.

What am I missing here?

cluster/write-file.go
galal-hussein: that's totally fine but can you combine both of these methods

alena1108:
It looks like this pr was put in before rke refactor, and we might need to rethink the flow to fit the refactored code. Something along the steps below should be done:

```
1) Write secrets encryption config is passed- should happen in the same place where we do deploy cloud config. https://github.com/rancher/rke/blob/master/cluster/plan.go#L84
2) Reconcile if checksum is diff similar to (https://github.com/rancher/rke/blob/master/cluster/plan.go#L171-L174)
3) Update all secrets in the place where we sync all the labels and annotations. https://github.com/rancher/rke/blob/master/cluster/cluster.go#L338:19
```

alena1108:
https://github.com/rancherlabs/rancher-security/issues/10

alena1108:
Closing as the feature is being re-worked in https://github.com/rancher/rke/pull/1654

Issue #677, opened 2018-06-07
Port check should make sure that ports for network plugins are open


galal-hussein:
Currently port checker only check for ports between k8s components, but doesn't check for network plugin ports:

**Flannel:**

8472/UDP 

**Canal:**

8472/UDP
9099/TCP

**Calico:**

9099/TCP
179/TCP

**Weave:**

6781-6782 TCP
6783-6784 TCP, UDP

superseb:
9099/TCP is a liveness/readiness port, that does not need to be open between nodes.

PR #1077, opened 2019-01-08
Add port checks for network plugins

guangbochen:
**Problem:**
The port check should make sure that ports for network plugins are open.

**Solution:**
Add port checks for network plugins and check by network type:

Flannel:
8472 | UDP | VXLAN

Canal:
8472 | UDP | VXLAN

Calico:
179 | TCP | BGP

Weave:
6781-6782 | TCP | Metrics
8673 | TCP | Network
8783-6784 UDP | Network

**Issues:**
https://github.com/rancher/rke/issues/677


loganhz:
@moelsayed @galal-hussein Could you help to review this PR, please? Thanks
cc @alena1108

cluster/network.go
```diff
@@ -454,9 +524,19 @@ func (c *Cluster) runServicePortChecks(ctx context.Context) error {
...
 	var hosts []string
 
+	cmd := []string{
+		"sh",
+		"-c",
+	}
+	if tcp {
+		cmd = append(cmd, "for host in $HOSTS; do for port in $PORTS ; do echo \"Checking host ${host} on port ${port}\" >&1 & nc -w5 -z $host $port > /dev/null || echo \"${host}:${port}\" >&2 & done; wait; done")
+	} else {
+		cmd = append(cmd, "for host in $HOSTS; do for port in $PORTS ; do echo \"Checking host ${host} on port ${port}\" >&1 & nc -w5 -uzv $host $port > /dev/null & done; wait; done")
```
moelsayed: Why did you omit the fail check `|| echo \"${host}:${port}\"` ?

guangbochen: Because `UDP` port scans using the `-uz` combination of flags will always report success irrespective of the target machine's state, so instead we use `-uzv` to log verbose output if the checking port is not open or unreachable, and it will print nothing if it succeeds.

moelsayed: @guangbochen Can you please add that as a comment ?

guangbochen: addressed @moelsayed.

PR #1082, opened 2019-01-14
Add restart components to custom certs

galal-hussein:


cluster/reconcile.go
```diff
@@ -243,3 +250,69 @@ func cleanControlNode(ctx context.Context, kubeCluster, currentCluster *Cluster,
...
+		pki.APIProxyClientCertName:     []services.RestartFunc{services.RestartKubeAPI},
+		pki.KubeControllerCertName:     []services.RestartFunc{services.RestartKubeController},
+		pki.KubeSchedulerCertName:      []services.RestartFunc{services.RestartScheduler},
+		pki.KubeProxyCertName:          []services.RestartFunc{services.RestartKubeproxy},
+		pki.KubeNodeCertName:           []services.RestartFunc{services.RestartKubelet},
+	}
+	for certName, changed := range AllCertsMap {
+		if changed {
+			for _, host := range allHosts {
+				for _, restartFunc := range AllCertsFuncMap[certName] {
```
moelsayed: Can we extract this in it's own function ? maybe we can wrap it in a errgroup for parallel rune.

Issue #1074, opened 2019-01-02
running rke up multiple times results with lots of orphaned container volumes


aemneina:
**RKE version:**
rke version v0.1.11

**Docker version: (`docker version`,`docker info` preferred)**
17.06.2

**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)**
ubuntu 16.04

**Type/provider of hosts: (VirtualBox/Bare-metal/AWS/GCE/DO)**
AWS

**cluster.yml file:**

**Steps to Reproduce:**
just run rke up a couple times, didnt even need to change anything in the yaml and saw a growth of 10 new volumes.

**Workaround:**
`docker volume prune` should do the trick to reclaim some space

**Results:**
could run out of disk space if one has a smaller volume and runs rke up a few times.

galal-hussein:
@aemneina I was able to find the root cause of this problem, the sidecar container is upgraded 5 times everytime rke runs, the sidecar container is used with k8s components to provide several helper scripts for cloud provider configuration, etc.

sangeethah:
Reproduced this issue with v0.1.11

This issue does not happen when testing with v0.2.0-rc5.

PR #1087, opened 2019-01-21
Avoid creating dangling volumes with rke tools

galal-hussein:
https://github.com/rancher/rke/issues/1074

### DIFF
--- a/cluster/plan.go
+++ b/cluster/plan.go
@@ -691,6 +691,7 @@ func (c *Cluster) BuildSidecarProcess() v3.Process {
 		Labels: map[string]string{
 			ContainerNameLabel: services.SidekickContainerName,
 		},
+		Command: []string{"/bin/bash"},
 	}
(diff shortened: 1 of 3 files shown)


cluster/plan.go
```diff
@@ -691,6 +691,7 @@ func (c *Cluster) BuildSidecarProcess() v3.Process {
 		Labels: map[string]string{
 			ContainerNameLabel: services.SidekickContainerName,
 		},
+		Command: []string{"/bin/bash"},
```
alena1108: is it to address some different issue?

alena1108: Confirmed: it is to address some slightly different issue when the upgrade was *always* performed for the sidekick container as on rke side the command verification was done incorrectly. And it contributed to the mass of the volumes being left around.

alena1108:
LGTM

PR #1096, opened 2019-01-25
Add checks for nil docker client

erikwilson:
To prevent against accidental SIGSEGV errors we should verify the use
of a non-nil docker client.

For rancher/rancher/issues/17519


docker/docker.go
```diff
@@ -98,6 +101,9 @@ func DoRunContainer(ctx context.Context, dClient *client.Client, imageCfg *conta
 }
 
 func DoRollingUpdateContainer(ctx context.Context, dClient *client.Client, imageCfg *container.Config, hostCfg *container.HostConfig, containerName, hostname, plane string, prsMap map[string]v3.PrivateRegistry) error {
+	if dClient == nil {
+		return fmt.Errorf("[%s] DoRollingUpdateContainer client is nil for container [%s] on host [%s]", plane, containerName, hostname)
```
galal-hussein: Can you make the error more descriptive, like `Failed to roll update the container: docker client is nil`

PR #1105, opened 2019-01-28
Remove network plugin port check

galal-hussein:


deniseschannon:
https://github.com/rancher/rke/issues/677

PR #1126, opened 2019-02-08
Add AWS cloudprovider config

superseb:
https://github.com/rancher/rancher/issues/17345

Depends on https://github.com/rancher/types/pull/709 && https://github.com/rancher/types/pull/714


templates/aws.go
```diff
@@ -0,0 +1,44 @@
+package templates
```
galal-hussein: is it possible to avoid using the template and just marshal the object to an ini file like in openstack https://github.com/rancher/rke/blob/master/cloudprovider/openstack/openstack.go#L40-L51

PR #1136, opened 2019-02-14
Handle legacy clusters restore

moelsayed:
https://github.com/rancher/rancher/issues/18035

### DIFF
--- a/templates/canal.go
+++ b/templates/canal.go
@@ -559,7 +559,7 @@ const CanalTemplateV113 = `
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1beta1
 metadata:
-  name: calico-node
+  name: calico
 rules:
   # The CNI plugin needs to get pods, nodes, and namespaces.
   - apiGroups: [""]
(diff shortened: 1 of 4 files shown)


templates/canal.go
```diff
@@ -559,7 +559,7 @@ const CanalTemplateV113 = `
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1beta1
 metadata:
-  name: calico-node
+  name: calico
```
galal-hussein: What is the reason behind changing this from `calico` to `calico-node` and what is the impact of upgrading cluster with this change?

moelsayed: The network addon job would fail with this error:
```
The ClusterRoleBinding "canal-calico" is invalid: roleRef: Invalid value: rbac.RoleRef{APIGroup:"rbac.authorization.k8s.io", Kind:"ClusterRole", Name:"calico-node"}: cannot change roleRef
```
`roleRef` is an immutable part of the spec. Check https://github.com/kubernetes/kubernetes/issues/66353 and https://github.com/kubernetes/website/pull/11254 for details.

galal-hussein: got it

PR #1142, opened 2019-02-18
rke canal cni flannel options

rafal-prasal:
generally it should make possible to set ipsec as flannel backend for flannel and canal

### DIFF
--- a/cluster/network.go
+++ b/cluster/network.go
@@ -41,16 +41,22 @@ const (
 
 	NoNetworkPlugin = "none"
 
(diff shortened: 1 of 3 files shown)


galal-hussein:
@rafal-prasal I think now the templates has moved to rancher/kontainer-driver-metadata repo, can you add the PR there and rebase this one with the right vendor

alena1108:
Closing the pr. If still relevant, please open it against rancher/kontainer-driver-metadata repo

Issue #1149, opened 2019-02-21
Manifests fromats


blaggacao:
https://github.com/rancher/rke/blob/40cd80a20864aad8003575c89c2ec74e996ee118/cluster/addons.go#L149-L177

Manifest such as https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/storage-class/local/default.yaml
do neither have `---` prefix nor do they terminate with `\n`.

Both should be checked and auto-completed when missing to avoid obscure error due to malformed concatenated yaml bytestring.
This should work for both, http and local files.

azbpa:
Can confirm this issue when using i.e. https://raw.githubusercontent.com/rancher/local-path-provisioner/master/deploy/local-path-storage.yaml in `cluster.yml` after another addon:

```yaml
addons_include:
  - /etc/rke/addons/myapp.yaml
  - https://raw.githubusercontent.com/rancher/local-path-provisioner/master/deploy/local-path-storage.yaml
```

Result is something like this where `Service` is the last resource of the first addon and `Namespace` is the first resource of the second file:

```yaml
...
---
apiVersion: v1
kind: Service
metadata:
  name: myapp
  namespace: default
  labels:
    app: myapp
spec:
  type: NodePort
  ports:
  - port: 8000
    targetPort: 8000
    nodePort: 30000
    name: http
  selector:
    app: myapp
apiVersion: v1
kind: Namespace
metadata:
  name: local-path-storage
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: local-path-provisioner-service-account
  namespace: local-path-storage
...
```

deniseschannon:
Available with v0.3.0-rc6

bmdepesa:
rke 0.3.0-rc6

Tested with multiple manifests:
* No ---, No \n, admin-user-1
* No ---, \n, admin-user-2
* ---, No \n, admin-user-3
* ---, \n, admin-user-4

cluster.yml
```
nodes:
  - address: node_address
    user: ubuntu
    role: [controlplane, etcd, worker]

addons_include:
  - manifest1.yml
  - manifest2.yml
  - manifest3.yml
  - manifest4.yml
```

manifest.yml (without dashes or newlines)
```
apiVersion: v1
kind: ServiceAccount
metadata:
  name: admin-user-x
  namespace: kube-system
```

All service accounts are created successfully:
```
kubectl get serviceaccounts -n kube-system | grep admin-user
admin-user-1                         1         2m17s
[...]

PR #1150, opened 2019-02-21
Fix manifest concat

blaggacao:
https://github.com/rancher/rke/issues/1149

### DIFF
--- a/cluster/addons.go
+++ b/cluster/addons.go
@@ -157,6 +157,14 @@ func (c *Cluster) deployAddonsInclude(ctx context.Context) error {
 			log.Infof(ctx, "[addons] Adding addon from url %s", addon)
 			logrus.Debugf("URL Yaml: %s", addonYAML)
 
+			// make sure we properly separated manifests
+			addonYAMLStr := string(addonYAML)
+			if !strings.HasPrefix(addonYAMLStr, "---") {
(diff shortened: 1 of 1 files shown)


blaggacao:
Haven't got a local git-flow set up. Therefore, sorry for the go fmt failure.

blaggacao:
@superseb rc7? It would come in very handy for me :wink:

blaggacao:
@superseb CI fixed. Green now. Merge?

superseb:
This was merged a while back addressing partly the same issue: https://github.com/rancher/rke/pull/1259. This now adds the same logic in another area, this should be cleaned up to have one proper implementation.

Secondly, can you explain why the newline is added in the if block for checking prefix being `---`? I don't see the relation between something not having a prefix and adding a newline to the end if it does.

blaggacao:
If it doesn't have a newline (not all follow good practice) the next file could continue on the same line as the last finished.

This breaks the yaml parser.

superseb:
I get why it's done, I don't understand why the newline should be added if it does not start with `---`. I refactored it in https://github.com/rancher/rke/pull/1388, let me know if that works for you.

alena1108:
Closing as the original issue is addressed with https://github.com/rancher/rke/pull/1388. @blaggacao please open an additional pr if you are interested with that extra new line fix.

blaggacao:
Looks perfect thanks!

PR #1153, opened 2019-02-21
Fix k8s version to allow upgrade

galal-hussein:
https://github.com/rancher/rancher/issues/18277


cluster/validation.go
```diff
@@ -84,7 +94,7 @@ func validateHostsOptions(c *Cluster) error {
 	return nil
 }
 
-func validateServicesOptions(c *Cluster) error {
+func validateServicesOptions(ctx context.Context, c *Cluster) error {
```
moelsayed: context is not used in this function, and the other ones.

vendor.conf
```diff
@@ -28,4 +28,4 @@ github.com/matttproud/golang_protobuf_extensions c12348ce28de40eed0136aa2b644d0e
 github.com/mattn/go-colorable            efa589957cd060542a26d2dd7832fd6a6c6c3ade
 github.com/mattn/go-isatty               6ca4dbf54d38eea1a992b3c722a76a5d1c4cb25c
 github.com/rancher/norman                0557aa4ff31a3a0f007dcb1b684894f23cda390c
-github.com/rancher/types                 5b7ed44f928712b6c0c7a1f3f19a93f32e2ee206
+github.com/rancher/types                 df7205653c84cd67ac2b21202b987c05d01873e9
```
moelsayed: can you please update this to the latest

PR #1159, opened 2019-02-27
Handle missing backups

moelsayed:
https://github.com/rancher/rancher/issues/18419


cluster/etcd.go
```diff
@@ -41,11 +42,21 @@ func (c *Cluster) PrepareBackup(ctx context.Context, snapshotPath string) error
 			if backupServer == nil { // start the download server, only one node should have it!
 				if err := services.StartBackupServer(ctx, host, c.PrivateRegistriesMap, backupImage, snapshotPath); err != nil {
 					log.Warnf(ctx, "failed to start backup server on host [%s]: %v", host.Address, err)
+					errors = append(errors, err)
 					continue
 				}
 				backupServer = host
```
galal-hussein: Should we break here if the backup server succeed to start

moelsayed: fixed.

util/util.go
```diff
@@ -7,13 +7,13 @@ import (
...
 	"github.com/coreos/go-semver/semver"
-	"github.com/rancher/types/apis/management.cattle.io/v3"
+	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
 )
 
 const (
 	WorkerThreads = 50
 
-	SupportedSyncToolsVersion = "0.1.22"
+	SupportedSyncToolsVersion = "0.1.25"
```
alena1108: I feel like we should use min version. Otherwise it's hard to keep up

moelsayed: It's an after/before version. We don't need to update this once it's released.

PR #1187, opened 2019-03-09
Upgrade legacy kubeapi service in certificate rotation

galal-hussein:
https://github.com/rancher/rancher/issues/18748


alena1108:
LGTM

PR #1188, opened 2019-03-09
Convert aarch64 to arm64 for ETCD_UNSUPPORTED_ARCH env var

jianghang8421:
On the ARM64 platform, ETCD_UNSUPPORTED_ARCH env var cannot recognize `aarch64`, so we need to convert `aarch64` to `arm64`.

https://github.com/rancher/rancher/issues/16518


loganhz:
LGTM

PR #1197, opened 2019-03-12
 Handle dnsconfig being a pointer

alena1108:
https://github.com/rancher/rancher/issues/18845

types pr https://github.com/rancher/types/pull/766


alena1108:
@superseb i see the field being referenced in validation.go as well, but assume validation gets called after the defaulting is done. Can you confirm.

cjellick:
tested the fix. lgtm

PR #1200, opened 2019-03-13
Handle mixed providers

moelsayed:
https://github.com/rancher/rancher/issues/18838

### DIFF
--- a/cluster/etcd.go
+++ b/cluster/etcd.go
@@ -3,6 +3,7 @@ package cluster
 import (
 	"context"
 	"fmt"
+	"regexp"
 
 	"github.com/rancher/rke/docker"
 	"github.com/rancher/rke/hosts"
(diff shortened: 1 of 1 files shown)


cluster/etcd.go
```diff
@@ -156,3 +171,14 @@ func (c *Cluster) getBackupImage() string {
 	}
 	return util.GetDefaultRKETools()
 }
+
+func IsLocalSnapshot(name string) bool {
+	// name is fmt.Sprintf("%s-%s%s-", cluster.Name, typeFlag, providerFlag)
```
alena1108: will it take care of the legacy snapshots taken prior to 2.2?

alena1108: You might want to make the reverse check. IsS3Snapshot.

moelsayed: Legacy snapshots are not affected in this case.

PR #1201, opened 2019-03-14
Restart Cattle agent pods and kubeapi auth pods

galal-hussein:
https://github.com/rancher/rancher/issues/18805


alena1108:
LGTM

Issue #911, opened 2018-09-18
Move to publishing archives in release


superseb:
* Makes it possible to have archives with $arch names, and just `rke` as binary name in each of them
* Saves data to transfer (~8MB vs ~30MB)

cjellick:
@galal-hussein to validate, right?

galal-hussein:
Verified with `v0.2.0-rc7` tag, rke binary was not released as expected.

superseb:
The goal is to have archives with the binaries, to be the same as how the CLI is released

cjellick:
drone isnt doing what we want, need to move out of the release

deniseschannon:
I believe this specific request is asking that we make specific versioned RKE CLI releases like we do in rancher CLI, where the artifacts are tagged with `-version`

Example: https://github.com/rancher/cli/releases/tag/v2.2.0

PR #1204, opened 2019-03-14
Add listing of artifacts in Drone similar to CLI

superseb:
First debug step for https://github.com/rancher/rke/issues/911


alena1108:
LGTM

PR #1225, opened 2019-03-22
Fix flannel issue with k8s 1.13

senivel:
This is a small update to the change made in https://github.com/rancher/rke/commit/24a84659415ae23aa4085e190c3238964ef0ad97#diff-7276c3d6a641862f48aed87f62158904, applying the updated tolerations to k8s 1.12 and above, to account for the default k8s 1.13 install on RKE v0.1.17.

### DIFF
--- a/templates/flannel.go
+++ b/templates/flannel.go
@@ -157,7 +157,7 @@ spec:
           mountPath: /host/opt/cni/bin/
       hostNetwork: true
       tolerations:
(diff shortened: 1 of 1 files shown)


janeczku:
Thanks! It looks like this already landed via the commit here in the meantime: https://github.com/rancher/rke/commit/21f3a3eff9467aac518e12daf5e13cd3ea35ef7f

PR #1257, opened 2019-04-04
Do *preliminary* nc KubeAPI port check using HTTP (not raw TCP)

### LINKED ISSUE #1256: RKE should perform preliminary KubeAPI port check over HTTP - not raw TCP
Many corporate and enterprise networks do not allow raw TCP connections, but most allow HTTP/HTTPs (even if through a HTTP proxy).

To increase compatibility, reduce installation errors, and for consistency, RKE should either switch to doing its preliminary KubeAPI port checks over HTTP or at least try and use HTTP as a fall-back before declaring the port as inaccessible.

**RKE version:**

rke version v0.2.1

**Docker version: (`docker version`,`docker info` preferred)**

18.09

**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)**

**Type/provider of hosts: (VirtualBox/Bare-metal/AWS/GCE/DO)**

**cluster.yml file:**

**Steps to Reproduce:**

Be on a network that only allows HTTP/HTTPs connections and try to bring up brand new cluster. Because `6443` is _initially_ TCP via `nc -kl -p 1337 -e echo` the installation process will fail like so. Note that if you somehow manage to get past this error, subsequent calls to `up` will succeed since `6443` will subsequently be HTTP once the API server is up.

`rke_darwin-amd64 up  --config generated/rke/cluster.yml`

**Results:**

```
INFO[0018] [network] Running control plane -> etcd port checks 
INFO[0019] [network] Successfully started [rke-port-checker] container on host [xxx.xxx.xxx.xxx] 
INFO[0019] [network] Running control plane -> worker port checks 
INFO[0020] [network] Successfully started [rke-port-checker] container on host [xxx.xxx.xxx.xxx] 
INFO[0020] [network] Running workers -> control plane port checks 
INFO[0021] [network] Successfully started [rke-port-checker] container on host [xxx.xxx.xxx.xxx] 
INFO[0021] [network] Checking KubeAPI port Control Plane hosts 
FATA[0097] [network] Can't access KubeAPI port [6443] on Control Plane host: xxx.xxx.xxx.xxx 
```

jlamillan:
This PR updates RKE's preliminary KubeAPI port check to fall back to HTTP when raw TCP fails to avoid potential firewall connectivity problems. 

Signed-off-by: Jesse Millan jlamillan@gmail.com

### DIFF
--- a/cluster/network.go
+++ b/cluster/network.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"net/http"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


galal-hussein:
I don't think this is possible to do, even if we allowed http port check we still can't access https, which means kubeconfig will basically not work because it uses https not http and will fail to update the cluster during rke installation and will fail.

jlamillan:
@galal-hussein

> I don't think this is possible to do, even if we allowed http port check we still can't access https, which means kubeconfig will basically not work because it uses https not http and will fail to update the cluster during rke installation and will fail.

I think you misunderstand the issue and/or this PR. Both HTTP and HTTPs requests, kubectl, etc. to the _actual_ API server (once it is up) work fine because those responses are HTTP and therefore can go through a HTTP proxy. This PR does not change the anything to do with the actual API server.

The issue we are seeing is very specific and only has to do with this initial and _temporary_ listener (6443->1337) process RKE spawns on `rancher/rke-tools` using the `nc -kl -p 1337 -e echo` during the _initial_ up command before the actual k8s API server is running. I suspect it does this as a  preliminary and rudimentary check just to see if port `6443` (the future API server port) _will_ be accessible from the RKE host.

That's why this (trivial) fix is only to modify the `nc` command to return a HTTP response. Without this fix, we cannot use RKE to bring up a new cluster for the first time. It's a total blocker. Also note, we have been using a custom build of RKE that contains this fix for a few days now and it works like a charm. Integration tests also passed.

jnicholls:
This should be merged, thanks @jlamillan for preparing it.

PR #1275, opened 2019-04-09
Fix fetching state with prefix path from nodes

galal-hussein:
https://github.com/rancher/rancher/issues/17965

The fix makes sure there is prefix path included in the file path because we already bind mount the volume /etc/kubernetes with prefix path


alena1108:
LGTM

PR #1277, opened 2019-04-09
[release/v0.1] Fix save and fetch state in clusters with prefix path

galal-hussein:
https://github.com/rancher/rancher/issues/19484


alena1108:
LGTM

PR #1281, opened 2019-04-10
Fix racy iptables between kube-proxy and CNI provider

janeczku:
This PR should mitigate the issue where the CNI provider in an RKE cluster fails to access iptables with the error: `iptables: Resource temporarily unavailable` leading to pod network disruption on the node. See 
https://github.com/rancher/rancher/issues/18637

#### Details:
`kube-proxy` and network plugins such as Flannel, Calico, etc. are all accessing iptables on the host.

Currently the kube-proxy container does not mount the common, host-level locking file (`/run/xtables.lock`), while the CNI plugins do. See https://github.com/rancher/rke/blob/master/templates/canal.go#L985-L987

This leads to racy/concurrent iptables access that may result in iptable calls to fail (`iptables: Resource temporarily unavailable` ) while CNI provider tries to ensure the required masquerade rules. In case of Flannel this may even cause pod networking to stop working until the particular host is rebooted.

Due to the racy nature of the issue the fix is difficult to validate. But it should be obvious that we need to fix the current race condition between multiple processes accessing iptables.

References:
- https://github.com/kubernetes-sigs/kubespray/issues/3192
- https://github.com/coreos/flannel/issues/988#issuecomment-453224071

### DIFF
--- a/cluster/plan.go
+++ b/cluster/plan.go
@@ -544,6 +544,7 @@ func (c *Cluster) BuildKubeProxyProcess(host *hosts.Host, prefixPath string) v3.
 	}
 	Binds := []string{
 		fmt.Sprintf("%s:/etc/kubernetes:z", path.Join(prefixPath, "/etc/kubernetes")),
+		"/run:/run",
 	}
 
 	for arg, value := range c.Services.Kubeproxy.ExtraArgs {


cluster/plan.go
```diff
@@ -544,6 +544,8 @@ func (c *Cluster) BuildKubeProxyProcess(host *hosts.Host, prefixPath string) v3.
 	}
 	Binds := []string{
 		fmt.Sprintf("%s:/etc/kubernetes:z", path.Join(prefixPath, "/etc/kubernetes")),
+		"/var/run:/var/run:rw,rprivate",
```
galal-hussein: is there a reason behind using rprivate mount propagation?

janeczku: tbh, i mirrored how `/run` is mounted for the kubelet but i guess in this case `/run:/run` should to it.

PR #1283, opened 2019-04-11
Decrease calico/canal logging for K8s 1.13 templates

axeal:
The Canal and Calico templates for K8s 1.13 introduce a regression in the use of info vs warning log level. Initial decrease of log level in the templates was made in https://github.com/rancher/rke/commit/c00e6d2e9d272ea76ce8147fb7c163a34fb369ad#diff-3497e44efe766fa9ca127075b285d611L118

This decreases log verbosity to warning in the 1.13 templates.

https://github.com/rancher/rancher/issues/19472

### DIFF
--- a/templates/calico.go
+++ b/templates/calico.go
@@ -675,7 +675,7 @@ data:
       "plugins": [
         {
           "type": "calico",
-          "log_level": "info",
+          "log_level": "WARNING",
           "datastore_type": "kubernetes",
           "nodename": "__KUBERNETES_NODE_NAME__",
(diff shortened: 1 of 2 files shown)


superseb:
This needs careful consideration before applying as it will trigger new pods. We should discuss how we want to adjust this and to be honest, discuss templates in general as current templates leads to confusion.

superseb:
I created https://github.com/rancher/rke/issues/1320 for managing the templates better, this can go in as is and backported to 2.2

superseb:
LGTM

PR #1292, opened 2019-04-16
[release/v0.1] Add cluster certificate rotation

galal-hussein:


cmd/cert.go
```diff
@@ -0,0 +1,122 @@
...
+	log.Infof(ctx, "Finished Rotating Kubernetes cluster certificates successfully")
+	return APIURL, caCrt, clientCert, clientKey, kubeCluster.Certificates, nil
+}
+
+func clusterRotateFromCli(ctx *cli.Context) error {
+	clusterFile, filePath, err := resolveClusterFile(ctx)
+	if err != nil {
+		return fmt.Errorf("Failed to resolve cluster file: %v", err)
+	}
+	clusterFilePath = filePath
```
alena1108: how exactly the clusterFile parameter being used? Is it being passed as a reference somewhere? As in this particular func, I don't see filePath or clusterFilePath being used.

cmd/cert.go
```diff
@@ -0,0 +1,122 @@
...
+		log.Infof(ctx, "This is a newely built cluster, can't rotate certificates")
+		return APIURL, caCrt, clientCert, clientKey, nil, nil
+	}
+
+	err = cluster.RotateRKECertificates(ctx, kubeCluster, currentCluster)
+	if err != nil {
+		return APIURL, caCrt, clientCert, clientKey, nil, err
+	}
+
+	err = kubeCluster.SetUpHosts(ctx, true)
```
alena1108: Nit: you might use `if err := <>; err != nil` construct on this line, and lines 51, 66, 85

cmd/cert.go
galal-hussein: clusterFilePath is a global variable we defined in cmd/up.go, and is being used to resolve cli file path, we refactored that in 0.2

PR #1296, opened 2019-04-17
Add prefix path validation

galal-hussein:
https://github.com/rancher/rancher/issues/17965


alena1108:
I'm fine with merging this PR as a temporary fix to prevent user of modifying the path as currently it breaks the cluster.

But in a long term, instead of disabling a way to modify prefix_path, we should add a proper support for it. I understand that it might not be easy (etcd, cni, other components' directories content needs to be moved once the prefix is modified).

alena1108:
Closing for now

PR #1301, opened 2019-04-22
Added onetime container and active running checks

superseb:
https://github.com/rancher/rancher/issues/19438

* Differentiates running containers and onetime containers
* Remove the native Docker wait and implement active polling for actively checking container via Docker and to keep the tunnel active (if there is no reply within 50 seconds, the tunnel closes). This also solves etcd snaphots that take longer than 50 seconds to complete (https://github.com/rancher/rancher/issues/19496). The current hardcoded timeout is 300 retries with 1 second sleep each. (5 minutes)
* The new onetime container failed on `rke remove` due to `rm -rf` ing the bind mounted folder. This is replaced with using find like for the rke logs.

### DIFF
--- a/cluster/file-deployer.go
+++ b/cluster/file-deployer.go
@@ -49,7 +49,7 @@ func doDeployFile(ctx context.Context, host *hosts.Host, fileName, fileContents,
 			fmt.Sprintf("%s:/etc/kubernetes:z", path.Join(host.PrefixPath, "/etc/kubernetes")),
 		},
 	}
-	if err := docker.DoRunContainer(ctx, host.DClient, imageCfg, hostCfg, ContainerName, host.Address, ServiceName, prsMap); err != nil {
+	if err := docker.DoRunOnetimeContainer(ctx, host.DClient, imageCfg, hostCfg, ContainerName, host.Address, ServiceName, prsMap); err != nil {
 		return err
 	}
 	if err := docker.DoRemoveContainer(ctx, host.DClient, ContainerName, host.Address); err != nil {
--- a/docker/docker.go
+++ b/docker/docker.go
@@ -101,6 +101,45 @@ func DoRunContainer(ctx context.Context, dClient *client.Client, imageCfg *conta
 	return nil
 }
 
+func DoRunOnetimeContainer(ctx context.Context, dClient *client.Client, imageCfg *container.Config, hostCfg *container.HostConfig, containerName string, hostname string, plane string, prsMap map[string]v3.PrivateRegistry) error {
+	if dClient == nil {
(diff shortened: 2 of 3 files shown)


docker/docker.go
```diff
@@ -345,19 +374,31 @@ func WaitForContainer(ctx context.Context, dClient *client.Client, hostname stri
 	if dClient == nil {
 		return 1, fmt.Errorf("Failed waiting for container: docker client is nil for container [%s] on host [%s]", containerName, hostname)
 	}
-	// We capture the status exit code of the container
-	statusCh, errCh := dClient.ContainerWait(ctx, containerName, container.WaitConditionNotRunning)
-	select {
-	case err := <-errCh:
+	// 5 minutes timeout, especially for transferring snapshots
+	for retries := 0; retries < 300; retries++ {
```
galal-hussein: I am not sure we should rewrite this part of docker client, I can see example of setting timeout to ContainerWait function https://godoc.org/github.com/docker/docker/client#example-Client-ContainerWait-WithTimeout

docker/docker.go
```diff
@@ -101,6 +101,35 @@ func DoRunContainer(ctx context.Context, dClient *client.Client, imageCfg *conta
...
+		}
+		if err := dClient.ContainerStart(ctx, resp.ID, types.ContainerStartOptions{}); err != nil {
+			return fmt.Errorf("Failed to start [%s] container on host [%s]: %v", containerName, hostname, err)
+		}
+		log.Infof(ctx, "Successfully started [%s] container on host [%s]", containerName, hostname)
+		log.Infof(ctx, "Waiting for [%s] container to exit on host [%s]", containerName, hostname)
+		if _, err := WaitForContainer(ctx, dClient, hostname, containerName); err != nil {
+			return fmt.Errorf("Container [%s] did not complete in time on host [%s]", containerName, hostname)
+		}
+		return nil
```
galal-hussein: If we are going to differentiate this, we should probably remove the container here as well if it served its purpose

cluster/file-deployer.go
```diff
@@ -49,7 +49,7 @@ func doDeployFile(ctx context.Context, host *hosts.Host, fileName, fileContents,
 			fmt.Sprintf("%s:/etc/kubernetes:z", path.Join(host.PrefixPath, "/etc/kubernetes")),
 		},
 	}
-	if err := docker.DoRunContainer(ctx, host.DClient, imageCfg, hostCfg, ContainerName, host.Address, ServiceName, prsMap); err != nil {
+	if err := docker.DoRunOnetimeContainer(ctx, host.DClient, imageCfg, hostCfg, ContainerName, host.Address, ServiceName, prsMap); err != nil {
```
galal-hussein: we should also change other container including:

- etcd onetime snapshot
- network check containers
- certificate deployer

Issue #1244, opened 2019-04-01
Failed to get /health for host - remote error: tls: bad certificate


danielbjornadal:
Getting `Failed to get /health for host - remote error: tls: bad certificate` when trying to `upgrade` an existing cluster. No modification to certificates have been done.

**RKE version:**
`rke version v0.2.1`

**Docker version:**
```
Client:
 Version:           18.06.3-ce
 API version:       1.38
 Go version:        go1.10.3
 Git commit:        d7080c1
 Built:             Wed Feb 20 02:27:18 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server:
 Engine:
  Version:          18.06.3-ce
  API version:      1.38 (minimum version 1.12)
  Go version:       go1.10.3
  Git commit:       d7080c1
  Built:            Wed Feb 20 02:26:20 2019
  OS/Arch:          linux/amd64
  Experimental:     false
```

**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)**
`16.04.4 LTS (Xenial Xerus) 4.4.0-116-generic`

**Type/provider of hosts: (VirtualBox/Bare-metal/AWS/GCE/DO)**
ESXi Virtual Machine

**cluster.yml file:**
```
nodes:
  - address: 10.10.7.121
    user: daniel
    role: [controlplane,worker,etcd]
  - address: 10.10.7.122
    user: daniel
    role: [controlplane,worker,etcd]
  - address: 10.10.7.123
    user: daniel
    role: [controlplane,worker,etcd]

services:
  etcd:
    snapshot: true
    creation: 6h
    retention: 24h
```

**Steps to Reproduce:**
`./rke -d up`

**Results:**
````
...
DEBU[0028] [remove/rke-log-linker] Container doesn't exist on host [10.10.7.123] 
DEBU[0028] [etcd] Checking image [rancher/rke-tools:v0.1.27] on host [10.10.7.123] 
DEBU[0028] Checking if image [rancher/rke-tools:v0.1.27] exists on host [10.10.7.123] 
DEBU[0028] Image [rancher/rke-tools:v0.1.27] exists on host [10.10.7.123] 
DEBU[0028] [etcd] No pull necessary, image [rancher/rke-tools:v0.1.27] exists on host [10.10.7.123] 
INFO[0029] [etcd] Successfully started [rke-log-linker] container on host [10.10.7.123] 
[...]

PR #1302, opened 2019-04-22
Search for legacy state and certs when trying to upgrade

galal-hussein:
https://github.com/rancher/rke/issues/1244

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -6,10 +6,12 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/rancher/rke/cluster"
 	"github.com/rancher/rke/hosts"
 	"github.com/rancher/rke/log"
+	"github.com/rancher/rke/util"
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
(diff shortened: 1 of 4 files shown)


cmd/common.go
```diff
@@ -128,17 +133,34 @@ func setS3OptionsFromCLI(c *cli.Context) *v3.S3BackupConfig {
...
-		// this cluster has a previous state, I don't need to upgrade!
+	if stateFileExists {
 		logrus.Debug("[state] previous state found, this is not a legacy cluster")
 		return nil
 	}
+	logrus.Debug("[state] previous state not found, possible legacy cluster")
+	return upgradePossibleLegacyCluster(ctx, kubeCluster, fullState, flags)
+}
+
+func upgradePossibleLegacyCluster(ctx context.Context, kubeCluster *cluster.Cluster, fullState *cluster.FullState, flags cluster.ExternalFlags) error {
```
alena1108: can we change the function name to reflect it's purpose better? as it's not going to do the upgrade

cmd/common.go
```diff
@@ -128,17 +133,34 @@ func setS3OptionsFromCLI(c *cli.Context) *v3.S3BackupConfig {
...
+	if err != nil {
+		return err
+	}
+	if !kubeConfigExists {
+		// if kubeconfig doesn't exist and its a legacy cluster then error out
+		if err := kubeCluster.TunnelHosts(ctx, flags); err != nil {
+			return err
+		}
+		if recoveredCluster := cluster.GetStateFromNodes(ctx, kubeCluster); recoveredCluster != nil {
+			return fmt.Errorf("This is a legacy cluster with no kube config, aborting upgrade")
```
alena1108: Can we put one more line to the error. Suggesting the way to fix it, something like - "Please re-run rke up using legacy rke for the current cluster to fetch and store the kubeconfig"

alena1108:
LGTM

Issue #1312, opened 2019-04-26
etcdv3.3 reports ETCDCTL_ENDPOINT as unknown environment variable


superseb:
**RKE version:**
v0.2.2

**Steps to Reproduce:**
* Create cluster with one node and Kubernetes 1.14
* Login to the node and run

```
docker exec etcd etcdctl member list
```

**Results:**
```
2019-04-26 15:11:21.688907 W | pkg/flags: unrecognized environment variable ETCDCTL_ENDPOINT=https://0.0.0.0:2379
```

alena1108:
Can be tested with rke `v0.3.0-rc2`

sowmyav27:
verified with `v0.3.0-rc4`

with with cluster - k8s version: 1.13. - on executing command - docker exec etcd etcdctl member list, message - `unrecognized environment variable ETCDCTL_ENDPOINT` is NOT seen.

with with cluster - k8s version: `v1.14.3-rancher1-1`. - on executing command - docker exec etcd etcdctl member list, message - `unrecognized environment variable ETCDCTL_ENDPOINT` is NOT seen.

PR #1318, opened 2019-04-30
Add correct env vars if etcd 3.3 is used

superseb:
https://github.com/rancher/rke/issues/1312

### DIFF
--- a/cluster/plan.go
+++ b/cluster/plan.go
@@ -11,7 +11,6 @@ import (
 
 	b64 "encoding/base64"
 
-	ref "github.com/docker/distribution/reference"
 	"github.com/docker/docker/api/types"
 	"github.com/rancher/rke/docker"
 	"github.com/rancher/rke/hosts"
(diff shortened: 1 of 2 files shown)


cluster/plan.go
```diff
@@ -770,11 +771,41 @@ func (c *Cluster) BuildEtcdProcess(host *hosts.Host, etcdHosts []*hosts.Host, pr
...
+		logrus.Debugf("Version [%s] is equal or higher than version [%s]", etcdSemVer, etcdChangedEnvSemVer)
+		Env = append(Env, "ETCDCTL_API=3")
+		// Point etcdctl to localhost in case we have listen all (0.0.0.0) configured
+		if listenAddress == "0.0.0.0" {
+			Env = append(Env, "ETCDCTL_ENDPOINTS=https://127.0.0.1:2379")
+			// If internal address is configured, set endpoint to that address as well
+		} else {
+			Env = append(Env, fmt.Sprintf("ETCDCTL_ENDPOINTS=https://%s:2379", listenAddress))
+		}
+		Env = append(Env, fmt.Sprintf("ETCDCTL_CACERT=%s", pki.GetCertPath(pki.CACertName)))
```
galal-hussein: Can we not repeat the common environments including:

ETCDCTL_API
ETCDCTL_CACERT
ETCDCTL_CERT
ETCDCTL_KEY

superseb: I thought about this and that would mean that to keep the correct order (to not change anything to trigger an update):

ETCDCTL_API first, then the if block, and then the rest, which makes it less readable in my opinion than this. Let me know

PR #1332, opened 2019-05-07
remove PersistentVolumeLabel controller

kinarashah:
It has been deprecated, adding it to baseEnabledAdmissionPlugins results into a warning, but the controller still gets loaded and affects the behavior for persistent volume creation. 

https://github.com/rancher/rancher/issues/19749


alena1108:
https://github.com/rancher/rancher/issues/19216
https://github.com/rancher/rancher/issues/19237

alena1108:
LGTM

Issue #705, opened 2018-06-20
Overide server URL in generated kube_config_cluster.yml


sebastien-prudhomme:
**RKE version:**

0.17

**Use case**

I have an external load balancer that setup a virtual IP for all my Kubernetes API endpoints.

When it configures kube_config_cluster.yml, RKE setup the cluster server URL with the first (or the last?) node with the "controlplane" role.

Having an option in cluster.yml config with the URL i want to use would be great.

Thanks

galal-hussein:
@sebastien-prudhomme RKE currently can't configure kube config with the ip of the load balancer, however rke can configure the certificates SAN to include the load balancer dns:
```
authentication:
  strategy: x509
  sans:
  - "10.18.160.10"
  - "my-loadbalancer-1234567890.us-west-2.elb.amazonaws.com"
```

sebastien-prudhomme:
I yet use the "sans" feature. But each time i do an "rke up", i need to manually correct the kube_config_cluster.yml.

Just need an option as the "cluster_name" one. Wish i know Go to implement it :-)

lmickh:
Using the SANs option to put the ELB name into the certs works, but if you don't have direct public access to the API server(s) the outputted kubeconfig "doesn't work".  The user has to edit the kubeconfig file with the externally accessible server address.

I get that RKE doesn't configure the kube config.  It is basically pulling it from the servers after the kube setup process generated.  It might make sense to give the user an edited version that "just works".  I would imagine the easy target would be to find and replace the server address with the `external-hostname` arg in the config when it exists.  When paired with the `sans` option already mentioned, it would give the user a working config file.

```
services:
  kube-api:
    extra_args:  
      external-hostname: my-api-elb.example.com
authentication:
  strategy: x509
  sans:
  - "10.10.10.10"
  - "my-api-elb.example.com"
```

Would result in a kubeconfig file with:
```
apiVersion: v1
kind: Config
clusters:
- cluster:
    api-version: v1
    server: "https://my-api-elb.example.com"
    certificate-authority-data: <snip>
  name: "rke-example"
```

The `external-hostname` option for the api-server is exactly for this use case.  It makes sense to just borrow that value for the kubeconfig file the user can use externally from the cluster in addition to the internal kubeconfig file resulting from the cluster setup.

willmao:
[...]

PR #1348, opened 2019-05-18
Use a provided `external-hostname` kube-apiserver arg in the kubeconfig and during RKE provisioning

jnicholls:
If a cluster config contains

```
services:
  kube-api:
    extra_args:
      external-hostname: my-load-balancer.example.com
```

this value will be used as the primary means to communicate with the Kube API for setup and port checks, as well as will go into the generated certs and kube config file.

If these semantics may not be what you're looking for, perhaps you would prefer a type change to the cluster config and have a top-level value such as `external_kube_api_host` or something that intrinsically passes down to the `external-hostname` arg to kube-apiserver.

#705

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -244,12 +244,13 @@ func rebuildLocalAdminConfig(ctx context.Context, kubeCluster *Cluster) error {
 	var workingConfig, newConfig string
 	currentKubeConfig := kubeCluster.Certificates[pki.KubeAdminCertName]
 	caCrt := kubeCluster.Certificates[pki.CACertName].Certificate
-	for _, cpHost := range kubeCluster.ControlPlaneHosts {
+	externalHostname := util.GetKubeAPIExternalHostname(&kubeCluster.RancherKubernetesEngineConfig)
+	if externalHostname != "" {
 		if (currentKubeConfig == pki.CertificatePKI{}) {
 			kubeCluster.Certificates = make(map[string]pki.CertificatePKI)
-			newConfig = getLocalAdminConfigWithNewAddress(kubeCluster.LocalKubeConfigPath, cpHost.Address, kubeCluster.ClusterName)
+			newConfig = getLocalAdminConfigWithNewAddress(kubeCluster.LocalKubeConfigPath, externalHostname, kubeCluster.ClusterName)
 		} else {
-			kubeURL := fmt.Sprintf("https://%s:6443", cpHost.Address)
+			kubeURL := "https://" + externalHostname
 			caData := string(cert.EncodeCertPEM(caCrt))
 			crtData := string(cert.EncodeCertPEM(currentKubeConfig.Certificate))
 			keyData := string(cert.EncodePrivateKeyPEM(currentKubeConfig.Key))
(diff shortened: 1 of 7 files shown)


billimek:
This would be a nice feature to have.

As it stands today if you deploy a multi-master cluster (using on `rke`), the configuration provided is hard-coded with just one of the hosts in the master.

As a kuberntes operator using rke,
I would like to denote a single host/ip for my cluster (e.g. a loadbalancer or VIP),
So that I don't need to manually alter the generated configuration to use the desired value

jnicholls:
> As a kuberntes operator using rke, I would like to denote a single host/ip for my cluster (e.g. a loadbalancer or VIP), so that I don't need to manually alter the generated configuration to use the desired value

Precisely. RKE generates the kubeconfig using this provided hostname, and uses it to test kube-api connectivity during provisioning. In general this is how most multi-master kube clusters are provisioned. The additional benefit to this is not needing to open up public access to port 6443 on the controlplane nodes. At my company, they have a strict whitelist of outbound connections where 443 is allowed but 6443 is not allowed, and I cannot convince them to whitelist all of 6443 (nor can I bear the churn of providing them specific IPs to whitelist for outbound 6443 access as I am spinning up and down clusters :-P). But that's a personal problem...not opening public 6443 on the nodes is very beneficial.

mjbright:
I've questions about this issue - feel free to point me elsewhere.
I'm deploying HA k8s on a customer site using rke.

If I understand correctly there's currently no way of configuring rke to use a front-end LB for the k8s api.
So rke cannot support HA directly without manually modifying the kubernetes config directly once rke had generated the cluster - is that correct?

Apart from specifying the sans address in the rke config, what would be needed to workaround this issue today?

Would it be sufficient to add some option to the relevant kubelet config on each node to point to the LB for the api?

Maybe I'm missing something else as I didn't understand the discussion about port 443/6443 use.

jnicholls:
@mjbright 

> If I understand correctly there's currently no way of configuring rke to use a front-end LB for the k8s api. So rke cannot support HA directly without manually modifying the kubernetes config directly once rke had generated the cluster - is that correct?

Yes that is correct. The support for the additional SANS in the ingress certificate is basically 33% of the support needed. The next third is supplying the `external-hostname` to the kubelet configuration, which answers your question about "add some option to the relevant kubelet config." The other third needed that RKE is missing is solved by this PR (if not the perfect way, in at least *a* way). This PR takes the supplied `external-hostname` if present, and ensures two things:

1. That the generated Kubeconfig file for the cluster uses this value as the endpoint to contact the kube-apiserver, rather than one of the IP addresses directly. With the kubelet's `external-hostname` and the SANS addition having the same value, everything fits together.
2. That while RKE is provisioning the cluster, it validates kube-apiserver connectivity and readiness by dialing this `external-hostname` endpoint rather than dialing each node's IP address directly. This is important if you want to keep the controlplane nodes private behind a loadbalancer for example.

I really wish this PR would receive some attention, as I find it to be a gaping hole in RKE's support for an HA setup. I wouldn't mind having to post-process the Kubeconfig file. The fact is that RKE provisioning completely fails if you either a) host controlplane nodes in a private network, or b) do not want to open port 6443 to your controlplane nodes or otherwise cannot dial-out to port 6443 from a corporate network that only allows common ports like 443. So #2 for me is the biggest issue at hand here that needs resolution.

Like I said earlier, I've been using this patch successfully for awhile now. It may not be how the Rancher folks want to solve it specifically, but it works. The only other optional I can think of is a new first-class config option available in the RKE cluster config, which will automatically fill in `external-hostname`, add a SANS entry, and be used in the generated Kubeconfig file without having to remember to set both `external-hostname` and the SANS entry to the same value.

jnicholls:
@galal-hussein Bump. Who can we task this to to provide feedback and/or approve for merge? Thanks.

superseb:
Having logic based on `extra_args` is not the way to go here, the parameter is to provide additional args to components, not have logic behind setting it which changes behavior other than the parameter. If this is based on existing logic, please point me to it. This should be a higher level setting, which then applies the needed logic/parameters to make it work.

Things that need to change based on the setting:

- Additional names in certificates so it accepts request using that name (`authentication.sans`)
- Generated kubeconfig for cluster
- Connectivity check to kube-api

Optionally:
- Generated kubeconfig for nodes (this will be `127.0.0.1:6443` and on non-controlplane nodes it will be `nginx-proxy` load balancing between controlplane nodes)

The latter depends on if we want to support only configuring external connectivity to kube-apiserver, because there is also an use-case for supporting internal connectivity (not having to open firewall rules from each worker node -> controlplane node. When we implement this, we also want to be able to change ports as there is probably a restricted network in play.

(key names to be decided)
```
loadbalancer:
  // Will apply changes described above
  kubeapi_external_fqdn: kubeapi.yourdomain.com
  // Optional
  kubeapi_external_port: 8443
  // Optional, disables nginx-proxy on the nodes and uses this in the node's kubecfg
  kubeapi_internal_fqdn: kubeapi-internal.yourdomain.com
  // Optional
  kubeapi_internal_port: 9443
```

While we are at it, we might as well make kube-apiserver port configurable so everything can be adjusted network wise.

jnicholls:
I agree it would be better to add an additional top-level config option, or as you have above, set of options. I didn't do that first-go-around because 1) I wanted to confirm that it was a preferred approach, and 2) it is far more involved to edit the config type, in an entirely different repo, and get it all tied together in this repo too for a release. I do not know the process that Rancher follows when changing their core type definitions.

Please advise very specifically what the process is for modifying core types and referencing them in downstream reports. Or, please make your suggested changes to this branch directly. I agree with your suggested option names, and on what they will affect.

jnicholls:
I rebased from v0.3.0-rc10 with no conflicts, and had a successful build. Apparently master created more conflicts on top of v0.3.0-rc10, so I will redo that soon.

superseb:
This won't go in as is, I will place a final design in here and describe the steps that are needed in detail. From there it can be picked up by anyone or I can pick it up myself to work on.

jnicholls:
Yes I know, we established that already. I’m looking for more detail on updating the `types` repository and coordinating that with a PR here. Or, you can prescribe and/or implement what it is you want. I had to rebase my branch in order to pick up the latest RKE release and close some CVEs. Thanks!

superseb:
The design is now in https://github.com/rancher/rke/issues/1682, we can work from there.

jnicholls:
@superseb There has been no movement, and now no assignment, for #1682. If I go ahead and implement your suggestions in #1682 will that receive a swift review and merge? Thanks.

PR #1370, opened 2019-05-29
 fix upgrade path for clusters created before 0.1.10

galal-hussein:
https://github.com/rancher/rancher/issues/20381


erikwilson:
LGTM

Issue #1373, opened 2019-05-30
Make SubnetLen field configurable in net-conf.json 


ukinau:
**RKE version:**
master

**feature requests**
Currently we can not specify SubnetLen for flannel. 
It will be better for end-user if we could specify SubnetLen for flannel
SubnetLen is the option to control the length of subnet to be allocated for each node
https://github.com/coreos/flannel/blob/master/Documentation/configuration.md

PR #1374, opened 2019-05-30
Support Configurable SubnetLen for flannel backend

ukinau:
Implemented this feature requests: https://github.com/rancher/rke/issues/1373


templates/flannel.go
```diff
@@ -74,6 +74,7 @@ data:
   net-conf.json: |
     {
       "Network": "{{.ClusterCIDR}}",
+      "SubnetLen": {{ .FlannelBackend.SubnetLen }},
```
moelsayed: There are 4 canal templates, and 2 flannel templates. I think this should go into all of them if it's supported by the addon versions.

moelsayed: https://github.com/rancher/rke/blob/master/templates/templates.go#L17-L25

ukinau: Thanks for your review. Yes I'm gonna add this to all of templates. 
Could you check PR again?

ukinau: @moelsayed

PR #1377, opened 2019-05-31
Add etcd snapshot compression

superseb:
https://github.com/rancher/rancher/issues/19496

### DIFF
--- a/services/etcd.go
+++ b/services/etcd.go
@@ -24,11 +24,12 @@ import (
 )
 
 const (
-	EtcdSnapshotPath     = "/opt/rke/etcd-snapshots/"
-	EtcdRestorePath      = "/opt/rke/etcd-snapshots-restore/"
-	EtcdDataDir          = "/var/lib/rancher/etcd/"
-	EtcdInitWaitTime     = 10
-	EtcdSnapshotWaitTime = 5
+	EtcdSnapshotPath                = "/opt/rke/etcd-snapshots/"
+	EtcdRestorePath                 = "/opt/rke/etcd-snapshots-restore/"
+	EtcdDataDir                     = "/var/lib/rancher/etcd/"
+	EtcdInitWaitTime                = 10
+	EtcdSnapshotWaitTime            = 5
+	EtcdSnapshotCompressedExtension = "gz"
 )
 
 func RunEtcdPlane(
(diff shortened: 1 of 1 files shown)


services/etcd.go
```diff
@@ -374,7 +375,7 @@ func DownloadEtcdSnapshotFromS3(ctx context.Context, etcdHost *hosts.Host, prsMa
 			fmt.Sprintf("%s:/backup:z", EtcdSnapshotPath),
 			fmt.Sprintf("%s:/etc/kubernetes:z", path.Join(etcdHost.PrefixPath, "/etc/kubernetes"))},
 		NetworkMode:   container.NetworkMode("host"),
-		RestartPolicy: container.RestartPolicy{Name: "always"},
+		RestartPolicy: container.RestartPolicy{Name: "on-failure"},
```
kinarashah: Just want to confirm, this change isn't related to compression snapshot logic?

services/etcd.go
```diff
@@ -526,10 +526,13 @@ func GetEtcdSnapshotChecksum(ctx context.Context, etcdHost *hosts.Host, prsMap m
 	if _, err := docker.WaitForContainer(ctx, etcdHost.DClient, etcdHost.Address, EtcdChecksumContainerName); err != nil {
 		return checksum, err
 	}
-	_, checksum, err = docker.GetContainerLogsStdoutStderr(ctx, etcdHost.DClient, EtcdChecksumContainerName, "1", false)
+	stderr, checksum, err = docker.GetContainerLogsStdoutStderr(ctx, etcdHost.DClient, EtcdChecksumContainerName, "1", false)
 	if err != nil {
 		return checksum, err
 	}
+	if stderr != "" {
+		logrus.Debugf("Error output not nil from snapshot checksum container [%s]: %v", EtcdChecksumContainerName, err)
```
kinarashah: should we return here as well similar to L529?

services/etcd.go
alena1108: We should be careful changing this default. I'd rather avoid doing it as a part of etcd compression pr, unless it's necessary.

superseb: We changed the behavior of waiting for a container to exit in https://github.com/rancher/rke/pull/1301/, which means that this will keep waiting for it to exit but it won't as it will restart. Also being a one-off container and having retry logic in the code in the container, there is no need. If we don't want this change, I can run other test to validate if it will work.

galal-hussein:
LGTM, only one small comment, if we are going to compress the snapshot I think we should add the cluster.rkestate into the compressed file, this will ensure that our snapshots has the right certificates all the time

PR #1387, opened 2019-06-10
WIP don't consider spec.SystemImages for rke-tools

kinarashah:


cluster/addons.go
```diff
@@ -231,7 +231,7 @@ func (c *Cluster) deployKubeDNS(ctx context.Context) error {
 	log.Infof(ctx, "[addons] Setting up %s", c.DNS.Provider)
 	KubeDNSConfig := KubeDNSOptions{
 		KubeDNSImage:           c.SystemImages.KubeDNS,
-		KubeDNSSidecarImage:    c.SystemImages.KubeDNSSidecar,
+		KubeDNSSidecarImage:    DefaultRKETools,
```
moelsayed: This actually `gcr.io/google_containers/k8s-dns-sidecar-amd64`, not `rke-tools`.

kinarashah: thank you, mistook it for Kubernetes sidecar

cluster/plan.go
```diff
@@ -27,9 +27,8 @@ const (
 	ContainerNameLabel = "io.rancher.rke.container.name"
 	CloudConfigSumEnv  = "RKE_CLOUD_CONFIG_CHECKSUM"
 
-	DefaultToolsEntrypoint        = "/opt/rke-tools/entrypoint.sh"
-	DefaultToolsEntrypointVersion = "0.1.13"
-	LegacyToolsEntrypoint         = "/opt/rke/entrypoint.sh"
+	DefaultToolsEntrypoint = "/opt/rke-tools/entrypoint.sh"
+	DefaultRKETools        = "rancher/rke-tools:v0.1.30"
```
moelsayed: I think it's better to move this to rancher/types and source it from there.

alena1108: @moelsayed we are working on a bigger task of moving all system images out of types and out of rke :)

moelsayed:
@kinarashah Did we test this change in an upgrade scenario? I think this would trigger an automatic cluster upgrade if applied.

alena1108:
@moelsayed we are not removing the rke-tools from system-images, so no auto upgrade should be triggered. The system-images.rke-tools image will just be ignored.

cluster/plan.go
```diff
@@ -879,20 +878,5 @@ func getUniqStringList(l []string) []string {
 }
 
 func (c *Cluster) getRKEToolsEntryPoint() string {
-	v := strings.Split(c.SystemImages.KubernetesServicesSidecar, ":")
```
moelsayed: This is required to support older 1.9.x and 1.10.x clusters from rancher 2.0 line.

kinarashah:
> @moelsayed we are not removing the rke-tools from system-images, so no auto upgrade should be triggered. The system-images.rke-tools image will just be ignored.

@alena1108, I don't respect system-images at all, so when setup upgrades, all older clusters will get triggered to get the new rke-tools I think

cluster/plan.go
kinarashah: oh. they can't work with newer rke-tools version?

moelsayed: TL;DR: No. 

We have two rke-tools "milestones":
- At one point, we had to change the entry point to fix a bug. For that we created `getRKEToolsEntryPoint()` which will check if the clusters rke-tools version is > or <= `0.1.13` and return an entry point based on that version. Most of the 2.0.x line clusters are `0.1.13` or less. Today I saw several clusters with `0.1.8`! This workaround was meant to be deprecated once we EOL those versions, but some users are still using them. So, it's safer to keep it around for a bit longer. 

- When we added rancher etcd backup, we maintained backward comparability with rke backup UI, but several internal changes had to be done to support s3 backup. This changed how `rke-tools` is used. So,  we created `util.IsRancherBackupSupported()` and the related logic to handle this. this way clusters with older `rke-tools` version that doesn't support s3 backup doesn't have to be updated to the latest k8s/rke-tools version, but still can utilize rancher s3 backup. Using this logic, rancher will pull the the default `rke-tools` and use it for backup operations without upgrading the cluster `system_images`. If the image is support the new backup api, it will be simply used instead of pulling a different image.

kinarashah: I was trying to avoid back porting and introducing K8s versions just because rke-tools changed, since they are not tied to kubernetes version. Based on these milestones, looks like we'll have to continue doing it for upgraded setups. 

Would it be helpful if we handled this logic this way? 

1) new clusters 
- rancher won't pass rke-tools version, so it'll only be dependent on rke version 

2) upgraded clusters 
- rancher uses the new driver metadata information to read rke-tools for k8s version and passes this in system images as before. I think this way we shouldn't need `util.IsRancherBackupSupported()` logic

kinarashah:
Looks like this will need more thought in order to handle upgrade scenarios. I am keeping this open for future reference. 
Issue: rke-tools is used as a sidecar for all k8s components and is a part of plan, so on all updates, clusters will undergo upgrade if rke-tools changes. 

But, we can remove dependency of etcd snapshot actions on this process and just use the default rke-tools. Putting up a separate pr for this, https://github.com/rancher/rke/pull/1389

alena1108:
Closing for now

Issue #997, opened 2018-11-02
Update cluster always duplicate nodes in Nodes page


myclau:
After add segment to cluster.yml:
```
- address: <host1>
      user: <user>
      role:
        - controlplane
        - etcd
        - worker
      hostname_override: <host1>
```
and run rke up again.

Always will show duplicate nodes in 
https://rancher.mydomain/c/clustername/nodes

other than the nodes create using rke first time will not duplicate.
For the new add nodes will duplicate once on this page.
But if you kubectl get all nodes it is normal

Example:

I use rke create cluster with 3 machine hostA,hostB,hostC

from 
https://rancher.mydomain/c/clustername/nodes

I will see 
hostA
hostB
hostC
total 3 nodes

After that I modify cluster to add hostD hostE, and run rke up again to update the cluster
from 
https://rancher.mydomain/c/clustername/nodes

I will see 
hostA
hostB
hostC
hostD
hostD
hostE
hostE

total 7 nodes 

but it should be 5 nodes.

for my cluster config template can found in:
https://github.com/myclau/Rancher2.0_HA_rke/blob/master/cluster.yml

alena1108:
@myclau could you provide more details: 

* what rke version you are using
* do you provision the cluster with rke using `rke up`, and then import it to rancher? If so, what rancher version are you using

myclau:
``` bash
$rke version
Server Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.3", GitCommit:"a4529464e4629c21224b3d52edfe0ea91b072862", GitTreeState:"clean", BuildDate:"2018-09-09T17:53:03Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"}
```
rancher version is 2.0.8.

To repo create 4vm:
1 as jumbox & LB
1 as etcd
2 as control & worker

using this cluster.yml
```yml
nodes:
  - address: ip2
    user: xxx
    role: [controlplane,etcd]
    hostname_override: ktest02
  - address: ip3
    user: xxx
    role: [worker]
    hostname_override: ktest03
  - address: ip4
    user: xxx
    role: [worker]
    hostname_override: ktest04

#If set to true, RKE will not fail when unsupported Docker version are found
ignore_docker_version: true

#Cluster level SSH private key
# Used if no ssh information is set for the node
ssh_key_path: ./keys/rancher_rsa

services:
  etcd:
    snapshot: true
    creation: 6h
    retention: 24h
[...]

PR #1398, opened 2019-06-13
duplicting nodes based on address and port

blcksrx:
Identifying duplicated nodes only based on the **addresses** is not really interesting because of some situation that I was dealt with that and others makes a problem (check this issue #997 and also some issue about that in **rancher** repository). In addition, I tried to do what people said in those issues I refer but it's not working at all and these solutions try to remove the question not responding it well! so the best solution that I think is to check both **address** and **port**

### DIFF
--- a/cluster/validation.go
+++ b/cluster/validation.go
@@ -183,7 +183,7 @@ func validateDuplicateNodes(c *Cluster) error {
 			if i == j {
 				continue
 			}
-			if c.Nodes[i].Address == c.Nodes[j].Address {
+			if c.Nodes[i].Address == c.Nodes[j].Address && c.Nodes[i].Port == c.Nodes[j].Port {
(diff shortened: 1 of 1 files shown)


moelsayed:
The code is assuming that it's possible to have two nodes with the same address and different ssh ports. I don't see this as a possible supported scenario.

blcksrx:
> The code is assuming that it's possible to have two nodes with the same address and different ssh ports. I don't see this as a possible supported scenario.

It's possible and i were in this situation as i am! and i did this PR for the problem that i were  dealing with.

moelsayed:
@blcksrx Could you please open an issue with clear description of your use case and link it to this PR? It will be useful to discuss the issue and validate it by QA later.

blcksrx:
> @blcksrx Could you please open an issue with clear description of your use case and link it to this PR? It will be useful to discuss the issue and validate it by QA later.

Okay.

PR #1399, opened 2019-06-13
CoreDNS default DNS provider for k8s 1.14 and up

superseb:
https://github.com/rancher/rancher/issues/19522

### DIFF
--- a/cluster/defaults.go
+++ b/cluster/defaults.go
@@ -11,7 +11,9 @@ import (
 	"github.com/rancher/rke/log"
 	"github.com/rancher/rke/services"
 	"github.com/rancher/rke/templates"
+	"github.com/rancher/rke/util"
 	"github.com/rancher/types/apis/management.cattle.io/v3"
+	"github.com/sirupsen/logrus"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


cluster/defaults.go
```diff
@@ -273,6 +276,32 @@ func (c *Cluster) setClusterImageDefaults() error {
 	return nil
 }
 
+func (c *Cluster) setClusterDNSDefaults() error {
+	if c.DNS == nil || len(c.DNS.Provider) == 0 {
```
moelsayed: any chance we can reverse this check and return nil as early as possible ?
``` 
if c.DNS != nil && len(c.DNS.Provider) != 0 {
   return nil
}
```

cluster/defaults.go
```diff
@@ -273,6 +276,32 @@ func (c *Cluster) setClusterImageDefaults() error {
...
+		}
+		logrus.Debugf("No DNS provider configured, setting default based on cluster version [%s]", clusterSemVer)
+		K8sVersionCoreDNSSemVer, err := util.StrToSemVer(K8sVersionCoreDNS)
+		if err != nil {
+			return err
+		}
+		// Default DNS provider for cluster version 1.14.0 and higher is coredns
+		ClusterDNSProvider := CoreDNSProvider
+		// If cluster version is less than 1.14.0 (K8sVersionCoreDNSSemVer), use kube-dns
+		if clusterSemVer.LessThan(*K8sVersionCoreDNSSemVer) {
```
moelsayed: also might be better reversed, use the Default and override it at the condition.

superseb: Due to a vendoring issue, we have an older `go-semver` lib with any `LessThan` available. I can reverse it by using `! if clusterSemVer.LessThan` but that seemed more ugly than this. If you still rather have that I can change it.

The vendoring issue will take more time to investigate and fix appropriately (and backport)

moelsayed: Not worth it. It was just a nit.

PR #1401, opened 2019-06-14
Fix always returning the default network template

garyhuang123:
templates/calico.go and templates/canal.go all have TemplateV112 and TemplateV113 definitions, but it always returns the default templateV112. On Rancher 2.2.4, this causes the Custom cluster calico-node always fails with readiness/liveness probe. Currently the workaround is that user can switch the readiness probe way from the command method /bin/calico-node -felix-ready -bird-ready to the http request method via :9099/readiness.


garyhuang123:
cancel this.

PR #1418, opened 2019-06-24
Add k8s 1.15

superseb:
https://github.com/rancher/rancher/issues/20986


cluster/plan.go
```diff
@@ -431,6 +432,19 @@ func (c *Cluster) BuildKubeletProcess(host *hosts.Host, prefixPath string) v3.Pr
 			c.Services.Kubelet.ExtraEnv,
 			fmt.Sprintf("%s=%s", KubeletDockerConfigFileEnv, path.Join(prefixPath, KubeletDockerConfigPath)))
 	}
+	// allow-privileged is removed in k8s 1.15
+	clusterSemVer, err := util.StrToSemVer(c.Version)
```
kinarashah: q: trying to understand, why can't we get major version from cluster version and use that?

superseb: We need SemVer to use `LessThan` on both versions. And `c.Version` is the cluster version right?

kinarashah: yea,  getting major from c.Version would give "v1.14" / "v1.15" and K8sVersion115 would be "v1.15" .. and we can directly do c.Version < K8sVersion115? not sure what I'm missing

superseb: I don't really trust using Go's comparison for versions (especially with semvers) but it seems to work, I updated the code.

kinarashah: oh okay, I just wanted to understand if I was missing something, since I have been doing similar comparisons for min/max versions for new metadata refactor

Issue #1437, opened 2019-06-30
NodeSelector value should be always string (Failed to deploy ingress-nginx, kube-dns, coredns when specify "XXX.com: true" in nodeSelector


ukinau:
**RKE version:**
master

**cluster.yml file:**

```
ingress:
    provider: nginx
    node_selector:
      ingress-node.test.com: "true"
    options:
      proxy-read-timeout: "3600"
    extra_args:
      v: 2

addon_job_timeout: 30
authentication: 
  strategy: "x509"
ignore_docker_version: true

ingress: 
  provider: none
kubernetes_version: "v1.13.5-rancher1-2"
monitoring: 
  provider: "metrics-server"

network: 
  options: 
    flannel_backend_type: "vxlan"
  plugin: "flannel"

services: 
  etcd: 
    backup_config: 
      enabled: true
      interval_hours: 12
      retention: 6
    creation: "12h"
    extra_args: 
      heartbeat-interval: 500
      election-timeout: 5000
    retention: "72h"
    snapshot: false
  kube-api: 
    always_pull_images: false
    pod_security_policy: false
    service_node_port_range: "30000-32767"
    service_cluster_ip_range: 172.30.0.0/15
  kube-controller:
    cluster_cidr: 172.16.0.0/14
    service_cluster_ip_range: 172.30.0.0/15
  kubelet:
    cluster_domain: cluster.local
    cluster_dns_server: 172.30.0.10

ssh_agent_auth: false
docker_root_dir: "/var/lib/docker"
enable_cluster_alerting: false
enable_cluster_monitoring: false
enable_network_policy: false
local_cluster_auth_endpoint: 
  enabled: false
```

**Steps to Reproduce:**
 - prepare node which have `ingress-node.test.com: "true"` label
 - prepare cluster.yml which configure ingress.nodeSelector to use `ingress-node.test.com: "true"`
 - Run `rke up`

**Results:**
Continuously failed to complete `rke-ingress-controller-deploy-job`. 

**Investigation:**
Kubernetes always expect nodeSelector to have `string` as a value, but if we specified `true` without quote as a value,  `true` will be treated as a boolean and then kube-apiserver reject that request because of type mis-match, that's why we need to quote value in nodeSelector in template. Related issue https://github.com/kubernetes/kubernetes/issues/23251

jiaqiluo:
The bug is reproduced in rke version v0.2.6

Steps:
- run rke up to start a single-node cluster
[...]

PR #1438, opened 2019-06-30
NodeSelector should specify string explicitly

ukinau:
#1437

Kubernetes doesn't accept any other value but string in nodeSelector,
but if we specified ambiguous value like true, it's treated as a
non-string like bool and then failed to create resource because of type
mis-match, that's why we should make value of nodeSelector enclosed by
double quotations to ensure value is always string


superseb:
LGTM

PR #1462, opened 2019-07-10
Check if certificates are present in state

superseb:
https://github.com/rancher/rancher/issues/21056

Problem: If certificates are empty in cluster state (or missing rkestate file), RKE and Rancher would throw NPE.

Solution: Check if certificates are present or error out (for now this situation needs manual intervention)


alena1108:
LGTM

Issue #1479, opened 2019-07-18
rke cert rotate does not work with --ssh-agent-auth: Can not update cluster


ajfriesen:
**RKE version:**

v0.2.5

**Docker version: (`docker version`,`docker info` preferred)**

```
docker version
Client:
 Version:           18.09.7
 API version:       1.39
 Go version:        go1.10.8
 Git commit:        2d0083d
 Built:             Thu Jun 27 17:56:17 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          18.09.2
  API version:      1.39 (minimum version 1.12)
  Go version:       go1.10.6
  Git commit:       6247962
  Built:            Sun Feb 10 03:42:13 2019
  OS/Arch:          linux/amd64
  Experimental:     false
```

docker info
```
docker info
Containers: 40
 Running: 37
 Paused: 0
 Stopped: 3
Images: 20
Server Version: 18.09.2
Storage Driver: aufs
 Root Dir: /var/lib/docker/aufs
 Backing Filesystem: extfs
 Dirs: 198
 Dirperm1 Supported: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge host macvlan null overlay
 Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: 894b81a4b802e4eb2a91d1ce216b8817763c29fb
runc version: 425e105d5a03fabd737a126ad93d62a9eeede87f
init version: fec3683
Security Options:
 apparmor
 seccomp
  Profile: default
Kernel Version: 4.4.0-1087-aws
Operating System: Ubuntu 16.04.6 LTS
OSType: linux
Architecture: x86_64
CPUs: 8
Total Memory: 31.42GiB
Name: ip-172-19-5-194
ID: 6AWQ:S5OP:LU3M:RSHU:D65B:P3QM:5UBJ:PRA2:7BAS:ZIXT:TDDN:BX6I
Docker Root Dir: /var/lib/docker
Debug Mode (client): false
Debug Mode (server): false
Registry: https://index.docker.io/v1/
Labels:
Experimental: false
Insecure Registries:
 127.0.0.0/8
Live Restore Enabled: false
Product License: Community Engine

WARNING: No swap limit support
```

[...]

PR #1522, opened 2019-07-30
Add per node kubelet server certificate

superseb:
This also includes fixes for:

https://github.com/rancher/rancher/issues/15862
https://github.com/rancher/rke/issues/1479

### DIFF
--- a/hosts/hosts.go
+++ b/hosts/hosts.go
@@ -272,6 +272,16 @@ func buildCleanerConfig(host *Host, toCleanDirs []string, cleanerImage string) (
 
 func NodesToHosts(rkeNodes []v3.RKEConfigNode, nodeRole string) []*Host {
 	hostList := make([]*Host, 0)
+	// Return all nodes if there is no noderole passed to the function
+	if nodeRole == "" {
+		for _, node := range rkeNodes {
+			newHost := Host{
+				RKEConfigNode: node,
+			}
+			hostList = append(hostList, &newHost)
+		}
+		return hostList
+	}
 	for _, node := range rkeNodes {
 		for _, role := range node.Role {
 			if role == nodeRole {
--- a/pki/constants.go
+++ b/pki/constants.go
@@ -21,6 +21,7 @@ const (
 	KubeSchedulerCertName      = "kube-scheduler"
 	KubeProxyCertName          = "kube-proxy"
 	KubeNodeCertName           = "kube-node"
+	KubeletCertName            = "kube-kubelet"
 	EtcdCertName               = "kube-etcd"
 	EtcdClientCACertName       = "kube-etcd-client-ca"
 	EtcdClientCertName         = "kube-etcd-client"
--- a/pki/services.go
+++ b/pki/services.go
@@ -372,7 +372,7 @@ func GenerateEtcdCertificates(ctx context.Context, certs map[string]CertificateP
 	sort.Strings(ips)
 
 	for _, host := range etcdHosts {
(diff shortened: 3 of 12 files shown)


hosts/hosts.go
```diff
@@ -272,6 +272,16 @@ func buildCleanerConfig(host *Host, toCleanDirs []string, cleanerImage string) (
 
 func NodesToHosts(rkeNodes []v3.RKEConfigNode, nodeRole string) []*Host {
 	hostList := make([]*Host, 0)
+	// If nodeRole is not set, return all
```
galal-hussein: how exactly the role is not set, the validation fails if the role not set in the config

pki/constants.go
```diff
@@ -21,6 +21,7 @@ const (
 	KubeSchedulerCertName      = "kube-scheduler"
 	KubeProxyCertName          = "kube-proxy"
 	KubeNodeCertName           = "kube-node"
+	KubeletCertName            = "kube-kubelet"
```
galal-hussein: the kube-node is supposed to be the kubelet cert, you should replace it if we are going to go with the cert per node solution since it won't be needed anymore

pki/services.go
```diff
@@ -484,6 +484,63 @@ func GenerateRKERequestHeaderCACert(ctx context.Context, certs map[string]Certif
 	return nil
 }
 
+func GenerateKubeletCertificate(ctx context.Context, certs map[string]CertificatePKI, rkeConfig v3.RancherKubernetesEngineConfig, configPath, configDir string, rotate bool) error {
```
galal-hussein: we also need to remove the kubenode cert generation, since it won't be needed anymore

pki/util.go
```diff
@@ -136,6 +136,34 @@ func GenerateCACertAndKey(commonName string, privateKey *rsa.PrivateKey) (*x509.
 	return kubeCACert, rootKey, nil
 }
 
+func GetIPHostname(host *hosts.Host) *cert.AltNames {
```
galal-hussein: can we change the name to express that we are getting alt names for a specific hosts

pki/util.go
```diff
@@ -342,7 +374,17 @@ func getEtcdCertKeys(rkeNodes []v3.RKEConfigNode, etcdRole string) []string {
...
 	etcdHosts := hosts.NodesToHosts(rkeNodes, etcdRole)
 	for _, host := range etcdHosts {
-		certList = append(certList, GetEtcdCrtName(host.InternalAddress))
+		certList = append(certList, GetCrtNameForAddress(host.InternalAddress, EtcdCertName))
+	}
+	certList = append(certList, getKubeletCertKeys(rkeNodes)...)
+	return certList
+}
+
+func getKubeletCertKeys(rkeNodes []v3.RKEConfigNode) []string {
```
galal-hussein: can we make a generic method to get cert/keys for specific hosts, you can pass the nodes + the role and return the certlist

hosts/hosts.go
superseb: if its not passed to the function, we return all hosts

pki/constants.go
superseb: I don't think we should merge the certificates because the kube-node certificate (and kubeconfig) is used by the kubelet to connect to the kube-api server. The kubelet certificate is what used to serve port 10250 on the kubelet and kube-api connects to. Besides the different function (auth vs server cert), it has different requirements which should not be merged into a single certificate.

Issue #157, opened 2017-12-16
Specify taints to nodes


superseb:
Ability to specify taints to nodes in cluster.yml

moelsayed:
Add `extra_arge:` to the nodes to be passed to the kubelet maybe ?

superseb:
Could be, need to test if multiple `--node-labels` arguments get merged. But I think labels/taints param to node would be nice.

moelsayed:
They get merged, we do [this](https://github.com/rancher/rke/blob/f5de7ff42ebdc247909a54e439dfc1c613be6c91/services/kubelet.go#L45) to assign multiple labels to nodes with multiple roles. 

I was thinking of a generic solution, but explicit params for this do sound better!

galal-hussein:
@superseb @moelsayed node-lables is still an alpha feature in k8s 1.8, so i don't think we should add an explicit option for it in rke, and the user may still add it through extra_args, and you can assign it multiple times so it will have no problem as @moelsayed pointed out.

rossedman:
@superseb @moelsayed @galal-hussein Is there a way to do this that is recommended? Use case would be I want nodes labeled so my addons can utilize taints and tolerations when standing up clusters.

japaniel:
I would also like to taint the nodes during cluster creation.

mei-tech:
> I would also like to taint the nodes during cluster creation.

Looking forward to see this also!

tlvenn:
Hi, is there any update on this please ?

nicknezis:
I also want this feature. I think I have a solution that I'll try to submit a merge request for. But I first need to update `rancher/types` repo. I'm having some Go dependency problems though.

nicknezis:
I've updated `RKEConfigNode` and `RKEConfigNodePlan` to include Taints similar to how Labels exist. But I'm having issues with `go generate` or `make` failing. Would someone be able to show me the steps to update `rancher/types`?

nicknezis:
Ok, I might need help getting `rancher/types` to generate the code that's causing the CI build to fail, but I believe the following two Pull Requests would satisfy this feature request (unless I'm missing something). I'll do some testing to make sure. (https://github.com/rancher/types/pull/829 & https://github.com/rancher/rke/pull/1349)

mkjpryor-stfc:
I would also like to see this feature. 

[...]

PR #1527, opened 2019-08-01
Support node taints configuration

orangedeng:
**Problem:**
We can not set node taints in RKE node config.

**Solution:**
Sync taints from config in `SyncLabelsAndTaints` function

Related issue:
https://github.com/rancher/rancher/issues/13972

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -363,6 +363,9 @@ func (c *Cluster) SyncLabelsAndTaints(ctx context.Context, currentCluster *Clust
 		}
 	}
 
+	// sync node taints. Add or remove taints from hosts
+	syncTaints(ctx, currentCluster, c)
+
 	if len(c.ControlPlaneHosts) > 0 {
 		log.Infof(ctx, "[sync] Syncing nodes Labels and Taints")
 		k8sClient, err := k8s.NewClient(c.LocalKubeConfigPath, c.K8sWrapTransport)
--- a/cluster/reconcile.go
+++ b/cluster/reconcile.go
@@ -21,6 +21,9 @@ const (
 	unschedulableEtcdTaint    = "node-role.kubernetes.io/etcd=true:NoExecute"
 	unschedulableControlTaint = "node-role.kubernetes.io/controlplane=true:NoSchedule"
 
+	unschedulableEtcdTaintKey    = "node-role.kubernetes.io/etcd=:NoExecute"
+	unschedulableControlTaintKey = "node-role.kubernetes.io/controlplane=:NoSchedule"
+
 	EtcdPlaneNodesReplacedErr = "Etcd plane nodes are replaced. Stopping provisioning. Please restore your cluster from backup."
 )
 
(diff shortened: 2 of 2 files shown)


gitlawr:
Can we do it in a similar way as labels - Prepare `ToAdd/DelLabels` early before the `SyncLabelsAndTaints` function.

cluster/cluster.go
```diff
@@ -84,6 +84,9 @@ const (
 	WorkerThreads = util.WorkerThreads
 
 	serviceAccountTokenFileParam = "service-account-key-file"
+
+	unschedulableEtcdTaintKey    = "node-role.kubernetes.io/etcd=:NoExecute"
+	unschedulableControlTaintKey = "node-role.kubernetes.io/controlplane=:NoSchedule"
```
gitlawr: Can you show reference for these reserved taints

orangedeng: I put the similar taints together. I think we will understand better in this way.

orangedeng:
Move the logic to `ReconcileCluster`, put them in function `syncTaints`. Updated.

cluster/reconcile.go
```diff
@@ -377,3 +382,74 @@ func isEtcdPlaneReplaced(ctx context.Context, currentCluster, kubeCluster *Clust
...
+	}
+	return rtn
+}
+
+func getTaintKey(taint v3.RKETaint) string {
+	return fmt.Sprintf("%s=:%s", taint.Key, taint.Effect)
+}
+
+func getTaintValue(taint v3.RKETaint) string {
+	return fmt.Sprintf("%s=%s:%s", taint.Key, taint.Value, taint.Effect)
```
gitlawr: Can we use `taint.Value` directly?

orangedeng: I am adding this value to `toAddTaints` and `toDelTaints` array.

gitlawr:
LGTM

cjellick:
@galal-hussein can you review?

vendor/github.com/rancher/types/apis/management.cattle.io/v3/rke_types.go
```diff
@@ -174,7 +174,7 @@ type RKEConfigNode struct {
 	// Node Labels
 	Labels map[string]string `yaml:"labels" json:"labels,omitempty"`
 	// Node Taints
-	Taints []v1.Taint `yaml:"taints" json:"taints,omitempty"`
+	Taints []RKETaint `yaml:"taints" json:"taints,omitempty"`
```
moelsayed: Why are you using a custom type ?

moelsayed:
@orangedeng We are doing this(or something very similar) near the end of provisioning the cluster https://github.com/rancher/rke/blob/master/cmd/up.go#L176. It's not a good idea to do it at the reconcile stage since it will affect nodes the might be added/deleted/modified later. 

Also, there are several edge cases around this process, and it's kind of a bottleneck when you have many nodes. Would it be possible to incorporate your changes in the current logic or extract the taints related code to avoid doing the work twice ?

orangedeng:
Updated

cjellick:
@moelsayed please re-review

deniseschannon:
https://github.com/rancher/rke/issues/157

cjellick:
worth having another set of eyes on this. @superseb please review

cjellick:
@superseb please review

daxmc99:
Does this support node taints on `rke up` with a config file?

orangedeng:
Yes.

orangedeng:
Rebased on master

Issue #1474, opened 2019-07-16
Ingress pods fail to start when PSP is enabled for cluster


janeczku:
**RKE version:**
v0.25

**cluster.yml file:**
```
nodes:
  - address: 167.71.xx.xx
    user: rancher
    role: [controlplane,worker,etcd]
  - address: 167.71.xx.xx
    user: rancher
    role: [controlplane,worker,etcd]
  - address: 207.154.xx.xx
    user: rancher
    role: [controlplane,worker,etcd]

services:
  kube-api:
    pod_security_policy: true
```

**Steps to Reproduce:**
Create cluster with above configuration, enabling pod security policy.

**Expected Result:**

Enabling PSP for a cluster should not break any of the workloads natively deployed by RKE.

The service account in the `ingress-nginx` namespace is assigned a role that allows replica set controller to use a predefined pod security policy to create required pods.
See: https://github.com/rancher/rke/blob/74b81a8766b7316c70f1d6dadcf794263539e60e/templates/authz.go#L73

**Actual Results:**
Ingress controller and default-backend pods fail to start:

```
Warning   FailedCreate  replicaset/default-http-backend-5954bd5d8c
Error creating: pods "default-http-backend-5954bd5d8c-" is forbidden: unable to validate against any pod security policy: []
```

dehotot:
+1 for a fix on this; it's blocking deployment of RBAC'ed end-user clusters.

dehotot:
FYI I was seeing this problem when using `plugin: flannel`.

Now I've switched to `plugin: canal` I don't seem to be able to trigger the problem.

deniseschannon:
Available with v0.3.0-rc6

sowmyav27:
### Reproduced issue with rke v0.2.5
**Cluster.yml file:**
```
nodes:
  - address: x.x.x.x
    user: ubuntu
    role: [controlplane,worker,etcd]
  - address: y.y.y.y
    user: ubuntu
    role: [controlplane,worker,etcd]
  - address: z.z.z.z
    user: ubuntu
    role: [controlplane,worker,etcd]

services:
  kube-api:
    pod_security_policy: true
```
- Do an rke up
- Result: No pods are seen running in `ingress-nginx` namespace
```
Sowmyas-MBP:import_cluster sowmyaviswanathan$ kubectl get pods --all-namespaces
NAMESPACE     NAME                                      READY   STATUS      RESTARTS   AGE
[...]

PR #1528, opened 2019-08-01
Fix ingress deployment issue with PSP enabled

moelsayed:
https://github.com/rancher/rke/issues/1474

### DIFF
--- a/authz/psp.go
+++ b/authz/psp.go
@@ -21,16 +21,16 @@ func ApplyDefaultPodSecurityPolicy(ctx context.Context, kubeConfigPath string, k
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


cjellick:
Can you explain the nature of this change?

moelsayed:
@cjellick For all components we deploy in `kube-system` namespace, we deploy a defailt role/rolebinding to allow the pods to work when PSP is enabled. The ingress controller runs in it's own name space so it doesn't have the default role/rolebinding deployed. This PR will make rke deploy those defaults in the ingress namespace as well, if PSP is enabled.

cjellick:
that is a great descriptoin @moelsayed. In my opinion, that type of description should go into the body of your commit messages. So that we can have that level of context further down the road

cjellick:
LGTM

PR #1539, opened 2019-08-08
Use correct labels to delete calico pods

superseb:
Problem: Calico pods are not deleted when etcd snapshot restore is done.
Root cause: The wrong k8s-label was used to select calico pods for removal (`calico` vs `calico-node`)
Fix: Provide correct labels and loop over labels to delete calico pods.

https://github.com/rancher/rancher/issues/21490

Also fixes a typo: `WeaveNetowrkAppName` -> `WeaveNetworkAppName `


dramich:
lgtm

deniseschannon:
@superseb can you file an issue for this to be tested in RKE CLI use case as well? It's available in v0.3.0-rc6

PR #1540, opened 2019-08-08
Make commonflags work with rke cert commands

superseb:
Problem: The flag `--ssh-agent-auth` cannot be used in `rke cert` commands.
Root cause: Common flags are not added to the `cert` subcommand
Fix: Add common flags to the `cert` subcommand

https://github.com/rancher/rke/issues/1479


dramich:
lgtm

Issue #1192, opened 2019-03-12
etcd member can not rejoin the cluster


cloudnautique:
**RKE version:**
v0.2.0-rc9

**Docker version: (`docker version`,`docker info` preferred)**
18.09
**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)**
```
NAME="Ubuntu"
VERSION="16.04.5 LTS (Xenial Xerus)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 16.04.5 LTS"
VERSION_ID="16.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
VERSION_CODENAME=xenial
UBUNTU_CODENAME=xenial
```
**Type/provider of hosts: (VirtualBox/Bare-metal/AWS/GCE/DO)**
GCE
**cluster.yml file:**
```
nodes:
- address: <pub_ip1>
  port: "22"
  internal_address: 10.240.0.4
  role:
  - controlplane
  - worker
  - etcd
  hostname_override: bill-rancher-rke-000
  user: wmaxwell
  docker_socket: /var/run/docker.sock
- address: <pub_ip2>
  port: "22"
  internal_address: 10.240.0.68
  role:
  - controlplane
  - worker
  - etcd
  hostname_override: bill-rancher-rke-001
  user: wmaxwell
  docker_socket: /var/run/docker.sock
- address: <pub_ip3>
  port: "22"
  internal_address: 10.240.0.70
  role:
  - controlplane
  - worker
  - etcd
  hostname_override: bill-rancher-rke-002
  user: wmaxwell
  docker_socket: /var/run/docker.sock
network:
  plugin: flannel
```
**Steps to Reproduce:**
1. In a 3 node cluster
1. Stop one of the nodes. 
1. Start back up. (Node public IP is reconfigured. Private stays consistent)
1. Edit cluster.yml and add nodes new public ip
1. Run `rke up`
1. run `rke cert rotate`

**Results:**
Etcd node doesn't rejoin.

sebastien-prudhomme:
Same "problem" here : my use case is reinstalling 1 node with the etcd role in a etcd cluster made of 3 nodes.

[...]

PR #1541, opened 2019-08-08
Handle changing public ip for etcd member delete

galal-hussein:
https://github.com/rancher/rke/issues/1192


hosts/tunnel.go
```diff
@@ -34,7 +36,11 @@ func (h *Host) TunnelUp(ctx context.Context, dialerFactory DialerFactory, cluste
 	}
 	// set Docker client
 	logrus.Debugf("Connecting to Docker API for host [%s]", h.Address)
-	h.DClient, err = client.NewClient("unix:///var/run/docker.sock", DockerAPIVersion, httpClient, nil)
+	h.DClient, err = client.NewClientWithOpts(
```
moelsayed: What happened here ?

galal-hussein: NewCLient is going to be deprecated and basically now it calls NewClientWithOpts, NewClientWithOpts supports configuring the dialer timeout which was causing a very big delay ~ 2 min every time rke tries to connect to a failed node

PR #1545, opened 2019-08-09
don't override tag of custom system images for defaultRKETools

kinarashah:
Issue: We extract tag from the passed system image and replace it from Default k8's system image. This causes problem when the passed system image is custom, and gets overridden 

https://github.com/rancher/rke/issues/1551

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -2,11 +2,12 @@ package util
 
 import (
 	"fmt"
-	"github.com/rancher/rke/metadata"
 	"net/url"
 	"os"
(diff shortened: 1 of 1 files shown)


prachidamle:
LGTM

superseb:
I don't think we can configure anything else besides having `rancher/rke-tools`. The only thing I can think of is `rancher/rke-tools:` to isolate to `rke-tools`.

dramich:
lgtm

Issue #1531, opened 2019-08-02
RKE should allow setting the DNS Policy for nginx-ingress controller to ClusterFirstWithHostNet (amongst others)


Oats87:
Currently, the nginx ingress controller gets deployed with the default DNS policy of `ClusterFirst` even though it operates on the host network.

It can be desirable to have the nginx-ingress-controller use the DNS policy `ClusterFirstWithHostNet` which will configure DNS queries to go to the cluster's DNS provider, rather than to the DNS servers of the node.

Note that this should not be a blind default, but rather an option to enable if necessary.

Oats87:
Information on the DNS policies here: https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy

sangeethah:
Tested with rancher server build from master - `d1718af7d` 

nginx-ingress-controller deployed in cluster any of the K8s versions - v1.13.10 , v1.14.6 and v1.15.3 have `dnsPolicy: ClusterFirst` set.

sangeethah:
Tested with latest rancher server build from master `24e4bdfe4`

nginx-ingress-controller deployed in cluster with K8s versions - v1.15.3 still has  `dnsPolicy: ClusterFirst` set.

mrajashree:
@sangeethah I do see dnsPolicy being set to `ClusterFirstWithHostNet` when I pass the following to ingress spec
`dns_policy:  ClusterFirstWithHostNet`

sangeethah:
Tested with rancher build from master - `` 

By default `dnsPolicy: ClusterFirst` gets set for nginx-ingress-controller d. 
We have the ability now to set `dnsPolicy` cluster yml duiring cluster creation (using Edit as YAML option UI)
```
  ingress:
    dns_policy: ClusterFirstWithHostNet
    provider: nginx
```

nginx-ingress-controller  gets deployed with "ClusterFirstWithHostNet" as expected.
Existing automated ingress scenarios work as expected .

PR #1550, opened 2019-08-13
Adding DNS Policy support for nginx ingress controller

Oats87:
Works on: https://github.com/rancher/rke/issues/1531

Relies on: https://github.com/rancher/kontainer-driver-metadata/pull/23 and https://github.com/rancher/types/pull/954


superseb:
With the values being static, we can add validation on the values.

superseb:
The only possible issue I see here is that its only done for ingress controller, and not other addons, which means this needs to be either copy pasted to the other addons later or refactored into a generic approach to adjust this.

Oats87:
OK, I'll work on building validation logic in and get this PR updated.

Oats87:
@superseb added some simple smoke-test validation

superseb:
Cool, thanks. If you can move it to `cluster/validation.go`, this is good to go.

Issue #1520, opened 2019-07-30
selinux-enabled nodes have hyperkube components receive permission denied for files from service-sidekick


Oats87:
**RKE version:** 0.2.6

**Docker version: (`docker version`,`docker info` preferred)** 18.09.8

**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)** CentOS Linux release 7.6.1810 (Core)

**Type/provider of hosts: (VirtualBox/Bare-metal/AWS/GCE/DO)** N/A, doesn't matter

**cluster.yml file:** 
```
nodes:
- address: 172.16.134.48
  user: centos
  role: [etcd, controlplane, worker]
```

**Steps to Reproduce:**
Start CentOS 7 or RHEL 7 with 
```
[root@ck-selinux-test _data]# cat /etc/docker/daemon.json
{
"selinux-enabled": true
}
[root@ck-selinux-test ~]# sestatus
SELinux status:                 enabled
SELinuxfs mount:                /sys/fs/selinux
SELinux root directory:         /etc/selinux
Loaded policy name:             targeted
Current mode:                   enforcing
Mode from config file:          enforcing
Policy MLS status:              enabled
Policy deny_unknown status:     allowed
Max kernel policy version:      31
[root@ck-selinux-test ~]#
```

and try to run `rke up` from another machine that has the capability to ssh

**Results:**
```
Endeavor:ck-selinux-test chriskim$ rke up
INFO[0000] Initiating Kubernetes cluster
INFO[0000] [dialer] Setup tunnel for host [172.16.134.48]
INFO[0003] [state] Pulling image [rancher/rke-tools:v0.1.34] on host [172.16.134.48]
INFO[0015] [state] Successfully pulled image [rancher/rke-tools:v0.1.34] on host [172.16.134.48]
INFO[0016] [state] Successfully started [cluster-state-deployer] container on host [172.16.134.48]
INFO[0017] [certificates] Generating CA kubernetes certificates
INFO[0017] [certificates] Generating Kubernetes API server aggregation layer requestheader client CA certificates
INFO[0017] [certificates] Generating Kubernetes API server proxy client certificates
INFO[0017] [certificates] Generating Kubernetes API server certificates
INFO[0017] [certificates] Generating Service account token key
INFO[0017] [certificates] Generating Kube Controller certificates
INFO[0017] [certificates] Generating Kube Scheduler certificates
INFO[0018] [certificates] Generating Kube Proxy certificates
[...]

PR #1556, opened 2019-08-15
Configure MCS labels if selinux is enabled

superseb:
https://github.com/rancher/rke/issues/1520

As we are editing `GetProcessConfig` and Rancher also uses this function, in Rancher we will pass an empty Host to the function.

### DIFF
--- a/docker/docker.go
+++ b/docker/docker.go
@@ -506,7 +506,8 @@ func IsContainerUpgradable(ctx context.Context, dClient *client.Client, imageCfg
 		!sliceEqualsIgnoreOrder(containerInspect.Config.Entrypoint, imageCfg.Entrypoint) ||
 		!sliceEqualsIgnoreOrder(containerInspect.Config.Cmd, imageCfg.Cmd) ||
 		!isContainerEnvChanged(containerInspect.Config.Env, imageCfg.Env, imageInspect.Config.Env) ||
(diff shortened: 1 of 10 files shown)


docker/docker.go
```diff
@@ -515,6 +516,8 @@ func IsContainerUpgradable(ctx context.Context, dClient *client.Client, imageCfg
 }
 
 func sliceEqualsIgnoreOrder(left, right []string) bool {
+	logrus.Debugf("Comparing, left: %s", left)
```
galal-hussein: This will generate tons of debug messages and will not really be useful to see which lists are we comparing, can we omit this

alena1108: It would make sense to log in debug when the result of comparison is false, so it's easier to figure out the discrepancy.

nheinemans:
Any chance this will make it into 2.3?

alena1108:
@nheinemans that's a plan

docker/docker.go
superseb: This was scoped to checking if it changed, if so, it will log the diff not both the values.

superseb:
Updates on the PR:
- Labels are now scoped to host with selinux enabled and to containers that share the service-sidekick container
- Changed the `sliceEqualsIgnoreOrder` function to only show the diff when its not equal.

Issue #1560, opened 2019-08-16
Print version when RKE is run


superseb:
Just like in Rancher, we should print the version in the log so we know what version is being used as this is sometimes missed when logs are provided or just to make sure you are looking at the correct log and the expected version was used.

jiaqiluo:
The enhancement is validated in rke version v0.3.0-rc9

Now it will print the rke cmd's version in the output 

```
➜  rke up
WARN[0000] This is not an officially supported version (v0.3.0-rc9) of RKE. Please download the latest official release at https://github.com/rancher/rke/releases/latest
INFO[0000] Running RKE version: v0.3.0-rc9

➜  rke version
WARN[0000] This is not an officially supported version (v0.3.0-rc9) of RKE. Please download the latest official release at https://github.com/rancher/rke/releases/latest
INFO[0000] Running RKE version: v0.3.0-rc9
```

PR #1561, opened 2019-08-16
Print version on start

superseb:
https://github.com/rancher/rke/issues/1560

This also moves the released match to under the part where proxy environment variables are shown, as they were only shown on an unreleased version and the code was not reached when it was determined to be a released version (`return nil`)


superseb:
Let me close this as it now also shows when you just run the binary without arguments which is weird.

Issue #1473, opened 2019-07-16
Log proxy environment variables when used


superseb:
When a proxy is used by RKE (for instance, for ServiceAccount creation) and it will not allow the request, you get an error like:

```
Failed to apply the ServiceAccount needed for job execution: Forbidden
```

It would help debug/diagnose if the log contains a line where it will log the configured environment variables regarding proxy.

sowmyav27:
### verified on rke `v0.3.0-rc6`
**Scenario#1**
3 environment variables set
```
export HTTP_PROXY = test
export HTTPS_PROXY=test1
export NO_PROXY=test2
```
- Do an rke up. The three env variables are see in the logs during an rke up
```
./rke up --config rancher-cluster.yml 
WARN[0000] This is not an officially supported version (v0.3.0-rc6) of RKE. Please download the latest official release at https://github.com/rancher/rke/releases/latest 
INFO[0000] Using proxy environment variable HTTP_PROXY with value [test] 
INFO[0000] Using proxy environment variable HTTPS_PROXY with value [test1] 
INFO[0000] Using proxy environment variable NO_PROXY with value [test2]
.....
```

**Scenario#2**
username and password available in the environment variables
```
export HTTP_PROXY=http://user1:password1@proxyurl.invalid
export HTTPS_PROXY=http://user2:password2@proxyurl.invalid
export NO_PROXY="*.aventail.com,home.com,.seanet.com"
```
- Do an rke up. The password is striped from the env variables
```
./rke up --config rancher-cluster.yml 
WARN[0000] This is not an officially supported version (v0.3.0-rc6) of RKE. Please download the latest official release at https://github.com/rancher/rke/releases/latest 
INFO[0000] Using proxy environment variable HTTP_PROXY with value [http://user1:***@proxyurl.invalid] 
INFO[0000] Using proxy environment variable HTTPS_PROXY with value [http://user2:***@proxyurl.invalid] 
INFO[0000] Using proxy environment variable NO_PROXY with value [*.aventail.com,home.com,.seanet.com] 
```

**Scenario#3**
Verified env variable uppercase vs lowercase
```
export HTTP_PROXY=http://user1:password1@proxyurl.invalid
export HTTPS_PROXY=http://user2:password2@proxyurl.invalid
export NO_PROXY="*.aventail.com,home.com,.seanet.com"
export http_proxy=test123
export https_proxy=test234
export no_proxy=test345
....
```
- on doing an rke up, the uppercase env variables are being preferred
```
[...]

PR #1562, opened 2019-08-16
Print proxy env vars when applying authz resources

superseb:
https://github.com/rancher/rke/issues/1473

Issue: Proxy environment variables were printed in main
Root cause: This means it will be printed every time and while it is only used in the rke -> kube-api check, so if we print it every time, it might confuse users that proxy is being used in other commands while they dont pass local proxy env vars to other containers executing actions.
Solution: add the log to ClusterUp so it will be used when using `rke up` and when Rancher uses it.

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -310,6 +310,8 @@ func ApplyAuthzResources(ctx context.Context, rkeConfig v3.RancherKubernetesEngi
 	if len(kubeCluster.ControlPlaneHosts) == 0 {
 		return nil
 	}
+	// Print proxy environment variables as we are directly contacting the cluster
+	util.PrintProxyEnvVars()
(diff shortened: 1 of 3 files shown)


cmd/up.go
```diff
@@ -76,6 +80,23 @@ func UpCommand() cli.Command {
 }
 
 func ClusterUp(ctx context.Context, dialersOptions hosts.DialersOptions, flags cluster.ExternalFlags, data map[string]interface{}) (string, string, string, string, map[string]pki.CertificatePKI, error) {
+	// Print proxy related environment variables
```
galal-hussein: we should move this part to an isolated function and move it outside ClusterUP as its not really related to the clusterup process

galal-hussein: I suggest to be part of validation.go code

superseb: I moved it to util, as its not really validating anything, its more that if the authz resources get applied, the proxy in the env vars is used and should be printed. The function is called just before authz resources are being created.

PR #1566, opened 2019-08-17
Remove uncompressed snapshot after restore

moelsayed:
https://github.com/rancher/rancher/issues/21710


services/etcd.go
```diff
@@ -479,28 +479,41 @@ func RestoreEtcdSnapshot(ctx context.Context, etcdHost *hosts.Host, prsMap map[s
...
 	}
 
 	imageCfg := &container.Config{
-		Cmd: []string{
-			"sh", "-c", fmt.Sprintf("rm -f %s", fullPath),
-		},
 		Image: etcdSnapshotImage,
 		Env:   es.ExtraEnv,
 	}
+	if cleanup {
```
kinarashah: can we rename cleanup to something that indicates its logic? Currently if cleanup is false, we remove everything

PR #1590, opened 2019-08-23
reverted Debugf + added --list-versions and small linting refactor

luthermonson:


cmd/config.go
```diff
@@ -405,6 +423,26 @@ func getAddonManifests(reader *bufio.Reader) ([]string, error) {
 	return addonSlice, nil
 }
 
+func generteK8sVersionList(all bool) error {
```
galal-hussein: typo

cmd/config.go
```diff
@@ -405,6 +423,26 @@ func getAddonManifests(reader *bufio.Reader) ([]string, error) {
...
+func generateK8sVersionList(all bool) error {
+	if !all {
+		fmt.Println(metadata.DefaultK8sVersion)
+		return nil
+	}
+
+	allVersions := []string{}
+	for _, version := range metadata.K8sVersionsCurrent {
+		if _, ok := metadata.K8sBadVersions[version]; !ok {
+			allVersions = append(allVersions, version)
```
rmweir: why not just print the version instead of build a slice then iterating it?

luthermonson: yup good point.

luthermonson:
@galal-hussein and @moelsayed this is pretty much ready to go if you two could sign off on it and confirm you like the --quiet fix.

docs PR: https://github.com/rancher/docs/pull/1692

cmd/config.go
```diff
@@ -98,28 +102,42 @@ func writeConfig(cluster *v3.RancherKubernetesEngineConfig, configFile string, p
 func clusterConfig(ctx *cli.Context) error {
 	if ctx.Bool("system-images") {
 		if metadata.K8sVersionToRKESystemImages == nil {
-			metadata.InitMetadata(context.Background())
+			err := metadata.InitMetadata(context.Background())
```
rmweir: nit: could be if err :=

cmd/config.go
```diff
@@ -98,28 +102,42 @@ func writeConfig(cluster *v3.RancherKubernetesEngineConfig, configFile string, p
...
+				return err
+			}
+
 		}
 		return generateSystemImagesList(ctx.String("version"), ctx.Bool("all"))
 	}
+
+	if ctx.Bool("list-versions") {
+		if metadata.K8sVersionToRKESystemImages == nil {
+			err := metadata.InitMetadata(context.Background())
```
rmweir: nit: could be if err :=

cmd/config.go
```diff
@@ -405,6 +423,21 @@ func getAddonManifests(reader *bufio.Reader) ([]string, error) {
 	return addonSlice, nil
 }
 
+func generateK8sVersionList(all bool) error {
```
rmweir: nit: maybe call it show or print list? generate makes me think something is coming back or get saved somewhere

PR #1597, opened 2019-08-28
RKE deploy job should tolerate all taints

orangedeng:
**Problem:**
Cluster fails to come up when the nodes have taints with effect
`NoExecute`.

**Solution:**
RKE deploy job should have tolerate all toleration to execute.
And this job should be only scheduled in linux node.

Related issue:
https://github.com/rancher/rancher/issues/22456


superseb:
LGTM

PR #1600, opened 2019-08-28
print version

leodotcloud:
https://github.com/rancher/rke/issues/1560

### DIFF
--- a/cmd/cert.go
+++ b/cmd/cert.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/sirupsen/logrus"
+
 	"github.com/rancher/rke/cluster"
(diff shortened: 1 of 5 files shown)


cmd/config.go
```diff
@@ -100,6 +101,7 @@ func writeConfig(cluster *v3.RancherKubernetesEngineConfig, configFile string, p
 }
 
 func clusterConfig(ctx *cli.Context) error {
+	logrus.Infof("Running RKE version: %v", ctx.App.Version)
```
superseb: I think we should stick to printing version when we run a sequence that involves checking/sharing logs, I don't think running config is part of those.

leodotcloud: Agree with you, will change.

superseb:
LGTM

PR #1608, opened 2019-08-29
Handle missing request header ca in rotate certificate

galal-hussein:
https://github.com/rancher/rancher/issues/20381

### DIFF
--- a/pki/deploy.go
+++ b/pki/deploy.go
@@ -198,25 +198,37 @@ func FetchCertificatesFromHost(ctx context.Context, extraHosts []*hosts.Host, ho
 	for certName, config := range crtList {
 		certificate := CertificatePKI{}
 		crt, err := FetchFileFromHost(ctx, GetCertTempPath(certName), image, host, prsMap, CertFetcherContainer, "certificates")
-		// I will only exit with an error if it's not a not-found-error and this is not an etcd certificate
-		if err != nil && !strings.HasPrefix(certName, "kube-etcd") {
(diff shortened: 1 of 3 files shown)


pki/deploy.go
```diff
@@ -199,24 +199,36 @@ func FetchCertificatesFromHost(ctx context.Context, extraHosts []*hosts.Host, ho
...
 		crt, err := FetchFileFromHost(ctx, GetCertTempPath(certName), image, host, prsMap, CertFetcherContainer, "certificates")
 		// I will only exit with an error if it's not a not-found-error and this is not an etcd certificate
-		if err != nil && !strings.HasPrefix(certName, "kube-etcd") {
+		if err != nil && !strings.HasPrefix(certName, "kube-etcd") &&
+			certName != RequestHeaderCACertName &&
+			certName != APIProxyClientCertName {
 			// IsErrNotFound doesn't catch this because it's a custom error
 			if isFileNotFoundErr(err) {
-				return nil, nil
+				return nil, fmt.Errorf("Certificate %s is not found", GetCertTempPath(certName))
```
moelsayed: From the  comment on 201, it looks like we actually intended to return nil here. Did you account for that in this change ?

galal-hussein: we intended to return nil in the earlier version because we were always pulling the certs from nodes, unlike now we only pull those certs if we are doing upgrades/rotation on older clusters, so we should return error if one of the certs are missing

moelsayed: Probably should amend the comment and explain that for future us.

PR #1641, opened 2019-09-17
Move snapshot remove logic to rke-tools

moelsayed:
Requires: https://github.com/rancher/rke-tools/pull/83

Support local local and s3 snapshot removal and restore clean up through rke-tools. 
https://github.com/rancher/rancher/issues/22900


alena1108:
LGTM

PR #1654, opened 2019-09-20
Add Secret Encryption Provider Support

moelsayed:
Requires: https://github.com/rancher/types/pull/977

Adds support for:
- Enable/disable managed secret encryption provider configuration.
- Automated key rotation for managed configuration.
- Support for passing custom provider configuration.


cluster/cluster.go
```diff
@@ -58,6 +58,13 @@ type Cluster struct {
 	UseKubectlDeploy                 bool
 	v3.RancherKubernetesEngineConfig `yaml:",inline"`
 	WorkerHosts                      []*hosts.Host
+	EncryptionConfig                 encryptionConfig
+}
+
+type encryptionConfig struct {
+	RewriteSecrets         bool
+	RotateKey              bool
+	EncryptionProviderFile string
```
leodotcloud: Instead of just taking the path to the encryption file, can we introduce the object `EncryptionConfiguration` in RKE config? That way the config can be passed directly.

moelsayed: This object is not exposed externally and it's used internally only. The `EncryptionConfiguration` type is used in `RancherKubernetesEngineConfig` where it can be added in the cluster yaml file or passed directly.

cluster/cluster.go
```diff
@@ -150,11 +160,49 @@ func ParseConfig(clusterFile string) (*v3.RancherKubernetesEngineConfig, error)
...
 
-func InitClusterObject(ctx context.Context, rkeConfig *v3.RancherKubernetesEngineConfig, flags ExternalFlags) (*Cluster, error) {
+func parseCustomEncryptionConfig(clusterFile string, rkeConfig *v3.RancherKubernetesEngineConfig) error {
+	if !isEncryptionCustomConfig(rkeConfig) {
+		return nil
+	}
+	var r map[string]interface{}
+	err := sigsyaml.Unmarshal([]byte(clusterFile), &r)
+	if err != nil {
+		return fmt.Errorf("error unmarshalling: %v", err)
```
alena1108: lets mention that we fail to unmarshall custom config

cmd/encryption.go
```diff
@@ -0,0 +1,101 @@
...
+	// We generate the first encryption config in ClusterInit, to store it ASAP. It's written
+	// to the DesiredState
+	stateEncryptionConfig := rkeFullState.DesiredState.EncryptionConfig
+
+	// if CurrentState has EncryptionConfig, it means this is NOT the first time we enable encryption, we should use the _latest_ applied value from the current cluster
+	if rkeFullState.CurrentState.EncryptionConfig != "" {
+		stateEncryptionConfig = rkeFullState.CurrentState.EncryptionConfig
+	}
+
+	kubeCluster, err := cluster.InitClusterObject(ctx, rkeConfig, flags, stateEncryptionConfig)
```
alena1108: @moelsayed why stateEncryption config is not a part of rke config?

services/kubeapi.go
```diff
@@ -28,3 +28,17 @@ func removeKubeAPI(ctx context.Context, host *hosts.Host) error {
 func RestartKubeAPI(ctx context.Context, host *hosts.Host) error {
 	return docker.DoRestartContainer(ctx, host.DClient, KubeAPIContainerName, host.Address)
 }
+
+func RestartKubeAPIWithHealthcheck(ctx context.Context, hostList []*hosts.Host, df hosts.DialerFactory, certMap map[string]pki.CertificatePKI) error {
+	for _, runHost := range hostList {
```
alena1108: we should add some logging at the beginning and in the end of the method

k8s/secret.go
```diff
@@ -2,31 +2,24 @@ package k8s
...
-			return err
-		}
-		// update secret if its already exist
-		if _, err := k8sClient.CoreV1().Secrets(metav1.NamespaceSystem).Update(secret); err != nil {
-			return err
-		}
-	}
-	return nil
+func GetSecret(k8sClient *kubernetes.Clientset, secretName, namespace string) (*v1.Secret, error) {
+	return k8sClient.CoreV1().Secrets(namespace).Get(secretName, metav1.GetOptions{})
```
alena1108: why do we need to wrap it in the method? It's just a one liner for GetSecret and GetSecretsList.

cluster/encryption.go
```diff
@@ -0,0 +1,293 @@
...
+			kubeCluster.EncryptionConfig.EncryptionProviderFile = ""
+			return kubeCluster.DeployEncryptionProviderFile(ctx)
+		}
+		return kubeCluster.DisableSecretsEncryption(ctx, currentCluster)
+	}
+	return nil
+}
+
+func (c *Cluster) DisableSecretsEncryption(ctx context.Context, currentCluster *Cluster) error {
+	if len(c.ControlPlaneHosts) == 0 {
```
alena1108: We should add more logging :)

cluster/cluster.go
```diff
@@ -150,11 +160,49 @@ func ParseConfig(clusterFile string) (*v3.RancherKubernetesEngineConfig, error)
...
 	}
+	// the customConfig is mapped to a k8s type, which doesn't unmarshal well because it has a
+	// nested struct and no yaml tags. Therefor, we have to re-parse it again and assign it correctly.
+	// this only affects rke cli. Since rkeConfig is passed from rancher directly in the rancher use case.
+	parseCustomEncryptionConfig(clusterFile, &rkeConfig)
 	return &rkeConfig, nil
 }
 
-func InitClusterObject(ctx context.Context, rkeConfig *v3.RancherKubernetesEngineConfig, flags ExternalFlags) (*Cluster, error) {
+func parseCustomEncryptionConfig(clusterFile string, rkeConfig *v3.RancherKubernetesEngineConfig) error {
```
alena1108: This method belongs to utils. Lets put it to the go file where you do yaml manipulations

cluster/encryption.go
```diff
@@ -0,0 +1,293 @@
...
+	"github.com/rancher/rke/templates"
+	"github.com/rancher/types/apis/management.cattle.io/v3"
+	"k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	apiserverconfig "k8s.io/apiserver/pkg/apis/config/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	EncryptionProviderFilePath = "/etc/kubernetes/ssl/encryption.yaml"
```
alena1108: will this file be always static? the location, the name?

cluster/reconcile.go
```diff
@@ -131,6 +130,10 @@ func reconcileControl(ctx context.Context, currentCluster, kubeCluster *Cluster,
 	if err := rebuildLocalAdminConfig(ctx, kubeCluster); err != nil {
 		return err
 	}
+
+	// if err = ReconcileEncryptionProviderConfig(ctx, kubeCluster, currentCluster); err != nil {
```
alena1108: do we need these commented lines? or can they be cleaned up

cluster/state.go
```diff
@@ -292,3 +284,43 @@ func GetStateFromNodes(ctx context.Context, kubeCluster *Cluster) *Cluster {
 	log.Infof(ctx, "[state] Successfully fetched cluster state from Nodes")
 	return &currentCluster
 }
+
+func buildFreshState(ctx context.Context, kubeCluster *Cluster, newState *FullState) error {
+	rkeConfig := &kubeCluster.RancherKubernetesEngineConfig
```
alena1108: @superseb can you check this logic

leodotcloud:
One more comment, do we need to store the EncryptionConfig outside of RKEConfig? If we figure out the needed info directly from RKEConfig, probably we can avoid all the state logic?

cmd/encryption.go
moelsayed: We try to keep everything in the state, not the config. This way we can persist it faster and avoid syncing issues across kontainer-engine when working with rancher. Additional, in rancher case, the configuration will be stored in the cluster store in a secret, which should be more secure than the cluster CRD.

k8s/secret.go
moelsayed: It's used in few other places. Less typing and easier to read.

cluster/encryption.go
moelsayed: Yes. This is the rke "managed" file location. And it's safer to control the file name.

services/kubeapi.go
```diff
@@ -28,3 +28,17 @@ func removeKubeAPI(ctx context.Context, host *hosts.Host) error {
...
+		if err := RestartKubeAPI(ctx, runHost); err != nil {
+			return err
```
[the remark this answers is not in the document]
moelsayed: That's tricky. If we continue, we might end up with a split brain situation, where api servers are using different configurations. In that case, it's better to fail fast.

cluster/encryption.go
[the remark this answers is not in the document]
moelsayed: We have a lot of logic around managing this directory and securing it's permissions. It made more sense to use it.

cluster/encryption.go
```diff
@@ -0,0 +1,293 @@
...
+	key, err := c.extractActiveKey(c.EncryptionConfig.EncryptionProviderFile)
+	if err != nil {
+		return "", err
+	}
+	return disabledProviderFileFromKey(key)
+}
+
+func rewriteSecret(k8sClient *kubernetes.Clientset, secret *v1.Secret) error {
+	var err error
+	if err = k8s.UpdateSecret(k8sClient, secret); err == nil {
```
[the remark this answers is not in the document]
moelsayed: We don't, since we don't make any updates to the object in this case, we are simply rewriting it.

moelsayed:
@leodotcloud It's actually much easier to use rkeConfig instead of the state. But we use the state because it's more guarded and more resilient to failure when used from rancher.

cluster/cluster.go
moelsayed: It's encryption specific, so I moved it to encryption.go

cmd/up.go
```diff
@@ -84,7 +84,16 @@ func ClusterUp(ctx context.Context, dialersOptions hosts.DialersOptions, flags c
...
 		return APIURL, caCrt, clientCert, clientKey, nil, err
 	}
-	kubeCluster, err := cluster.InitClusterObject(ctx, clusterState.DesiredState.RancherKubernetesEngineConfig.DeepCopy(), flags)
+	// We generate the first encryption config in ClusterInit, to store it ASAP. It's written
+	// to the DesiredState
+	stateEncryptionConfig := clusterState.DesiredState.EncryptionConfig
+
+	// if CurrentState has EncryptionConfig, it means this is NOT the first time we enable encryption, we should use the _latest_ applied value from the current cluster
+	if clusterState.CurrentState.EncryptionConfig != "" {
+		stateEncryptionConfig = clusterState.CurrentState.EncryptionConfig
```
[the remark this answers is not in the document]
moelsayed: Yes, if there were failures in previous runs.

cluster/encryption.go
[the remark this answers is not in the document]
leodotcloud: May be, we can introduce another directory for all the configs, including audit and eventratelimit and replicate the secure settings. But, overloading ssl directory for secrets doesn't make sense.

cluster/cluster.go
```diff
@@ -150,11 +160,49 @@ func ParseConfig(clusterFile string) (*v3.RancherKubernetesEngineConfig, error)
...
+	if !ok {
+		return fmt.Errorf("secrets encryption custom config parsing error")
+	}
+	jsonConfig, err := json.Marshal(sec["custom_config"])
+	if err != nil {
+		return fmt.Errorf("error marshalling: %v", err)
+	}
+	customConfig := &v3.CustomEncryptionConfiguration{}
+	if err = json.Unmarshal(jsonConfig, customConfig); err != nil {
+		return err
```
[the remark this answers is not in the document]
moelsayed: Sorry, I pushed updates before committing this.

cluster/encryption.go
```diff
@@ -0,0 +1,293 @@
...
+
+// ReconcileDesiredStateEncryptionConfig We do the rotation outside of the cluster reconcile logic. When we are done,
+// DesiredState needs to be updated to reflect the "new" configuration
+func (c *Cluster) ReconcileDesiredStateEncryptionConfig(ctx context.Context, fullState *FullState) error {
+	fullState.DesiredState.EncryptionConfig = c.EncryptionConfig.EncryptionProviderFile
+	return fullState.WriteStateFile(ctx, c.StateFilePath)
+}
+
+func (c *Cluster) IsEncryptionEnabled() bool {
+	if c == nil {
```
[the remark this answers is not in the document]
moelsayed: I thought so, but it does break, when called from rancher!

cluster/cluster.go
```diff
@@ -150,11 +160,49 @@ func ParseConfig(clusterFile string) (*v3.RancherKubernetesEngineConfig, error)
...
+	}
+	rkeConfig.Services.KubeAPI.SecretsEncryptionConfig.CustomConfig = customConfig
+	return nil
+}
+
+func InitClusterObject(ctx context.Context, rkeConfig *v3.RancherKubernetesEngineConfig, flags ExternalFlags, encryptConfig string) (*Cluster, error) {
```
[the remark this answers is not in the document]
moelsayed: As discussed offline. It's doesn't look great, but it's necessary.

cluster/encryption.go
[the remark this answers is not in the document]
moelsayed: That's a good idea, but it should be done on it's own. This will probably touch a lot of areas including the worker code.

cluster/state.go
```diff
@@ -292,3 +284,43 @@ func GetStateFromNodes(ctx context.Context, kubeCluster *Cluster) *Cluster {
...
+	if pkiCertBundle[pki.RequestHeaderCACertName].Certificate == nil {
+		if err := pki.GenerateRKERequestHeaderCACert(ctx, pkiCertBundle, flags.ClusterFilePath, flags.ConfigDir); err != nil {
+			return err
+		}
+	}
+	if err := pki.GenerateRKEServicesCerts(ctx, pkiCertBundle, *rkeConfig, flags.ClusterFilePath, flags.ConfigDir, false); err != nil {
+		return err
+	}
+	newState.DesiredState.CertificatesBundle = pkiCertBundle
+	if oldState.DesiredState.EncryptionConfig != "" && isEncryptionEnabled(rkeConfig) {
```
superseb: Can we move this to a `isEncryptionEnabled(rkeConfig)` and then `if oldState.DesiredState.EncryptionConfig != "" {} else {}` ?

alena1108:
The changes are incorporated in https://github.com/rancher/rke/pull/1727

PR #1665, opened 2019-09-25
WIP: feature/control-plane loadbalancer

rowanj:
Allow the configuration of a TCP load-balancer to the control-plane (kube-apiserver) hosts.

depends on:
rancher/types#987

to-do:
- [ ] remove go.mod override of rancher/types


rowanj:
We've adapted our environment to allow the more conventional direct config for the k8s API addresses.

PR #1666, opened 2019-09-26
Changes for CIS - EventRateLimit and AuditLog 

leodotcloud:


cluster/cluster.go
```diff
@@ -151,6 +156,82 @@ func (c *Cluster) DeployWorkerPlane(ctx context.Context, svcOptionData map[strin
...
 }
 
+func parseAuditLogConfig(clusterFile string, rkeConfig *v3.RancherKubernetesEngineConfig) error {
+	if rkeConfig.Services.KubeAPI.AuditLog != nil &&
+		rkeConfig.Services.KubeAPI.AuditLog.Enabled &&
+		rkeConfig.Services.KubeAPI.AuditLog.Configuration.Policy == nil {
+		return nil
+	}
+	logrus.Debugf("audit log policy found in cluster.yml")
+	var err error
```
alena1108: no need to define the error here. Can just assign starting line 168

cluster/cluster.go
```diff
@@ -151,6 +156,82 @@ func (c *Cluster) DeployWorkerPlane(ctx context.Context, svcOptionData map[strin
...
+	var r map[string]interface{}
+	err = ghodssyaml.Unmarshal([]byte(clusterFile), &r)
+	if err != nil {
+		return fmt.Errorf("error unmarshalling: %v", err)
+	}
+	services := r["services"].(map[string]interface{})
```
alena1108: can move lines 172-174 till after line 175. So if there is a failure on line 178, we can save on conversion calls

cluster/cluster.go
```diff
@@ -151,6 +156,82 @@ func (c *Cluster) DeployWorkerPlane(ctx context.Context, svcOptionData map[strin
...
+	}
+	scheme := runtime.NewScheme()
+	err = auditv1.AddToScheme(scheme)
+	if err != nil {
+		return fmt.Errorf("error adding to scheme: %v", err)
+	}
+	codecs := serializer.NewCodecFactory(scheme)
+	p := auditv1.Policy{}
+	err = runtime.DecodeInto(codecs.UniversalDecoder(), policyBytes, &p)
+	if err != nil || p.Kind != "Policy" {
```
alena1108: nit: you can check for p.Kind policy before performing the check on line 187. This way you save on decode call if p.Kind != policy

cluster/cluster.go
```diff
@@ -151,6 +156,82 @@ func (c *Cluster) DeployWorkerPlane(ctx context.Context, svcOptionData map[strin
...
+	rkeConfig.Services.KubeAPI.AuditLog.Configuration.Policy = &p
+	return err
+}
+
+func parseAdmissionConfig(clusterFile string, rkeConfig *v3.RancherKubernetesEngineConfig) error {
+	if rkeConfig.Services.KubeAPI.AdmissionConfiguration == nil {
+		return nil
+	}
+	logrus.Debugf("admission configuration found in cluster.yml")
+	var err error
```
alena1108: nit: similar here, no need to define the error. Can just assign on line 202

cluster/cluster.go
```diff
@@ -151,6 +156,82 @@ func (c *Cluster) DeployWorkerPlane(ctx context.Context, svcOptionData map[strin
...
+	var r map[string]interface{}
+	err = ghodssyaml.Unmarshal([]byte(clusterFile), &r)
+	if err != nil {
+		return fmt.Errorf("error unmarshalling: %v", err)
+	}
+	services := r["services"].(map[string]interface{})
+	kubeapi := services["kube-api"].(map[string]interface{})
```
alena1108: assign services after doing the check on kubeapi :)

cluster/cluster.go
leodotcloud: 👍

cluster/cluster.go
leodotcloud: The `Kind` info is filled after decoding, hence can't move the check.

cluster/cluster.go
leodotcloud: 👍

cluster/cluster.go
leodotcloud: I was actually missing nil checks here, I don't think it's possible to move the lines without actually checking if the key exists.

cluster/cluster.go
leodotcloud: Same as above. Can't dereference without checking the previous key.

PR #1698, opened 2019-10-15
Add per node kubelet server certificate

superseb:
* Moved stripping CA key from `GenerateRKENodeCerts` to `DeployCertificatesOnPlaneHost` because we use `GenerateRKENodeCerts` in Rancher worker logic to generate the kubelet certificate and I can't do that without the CA key.
* Implemented `GetCrtNameForHost` to get certificate name as Rancher does not default nodes to have `InternalAddress` equal to `Address` when `InternalAddress` is empty (which RKE does).
* Made `DeepEqualIPsAltNames` public as it is needed in the nodeconfig comparing old vs new cert

Depends on https://github.com/rancher/types/pull/1013

### DIFF
--- a/pki/services.go
+++ b/pki/services.go
@@ -30,7 +30,7 @@ func GenerateKubeAPICertificate(ctx context.Context, certs map[string]Certificat
 	kubeAPICert := certs[KubeAPICertName].Certificate
 	if kubeAPICert != nil &&
 		reflect.DeepEqual(kubeAPIAltNames.DNSNames, kubeAPICert.DNSNames) &&
-		deepEqualIPsAltNames(kubeAPIAltNames.IPs, kubeAPICert.IPAddresses) && !rotate {
+		DeepEqualIPsAltNames(kubeAPIAltNames.IPs, kubeAPICert.IPAddresses) && !rotate {
 		return nil
 	}
 	log.Infof(ctx, "[certificates] Generating Kubernetes API server certificates")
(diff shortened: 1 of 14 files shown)


pki/services.go
```diff
@@ -484,6 +484,63 @@ func GenerateRKERequestHeaderCACert(ctx context.Context, certs map[string]Certif
...
+
+func GenerateKubeletCSR(ctx context.Context, certs map[string]CertificatePKI, rkeConfig v3.RancherKubernetesEngineConfig) error {
+	allHosts := hosts.NodesToHosts(rkeConfig.Nodes, "")
+	for _, host := range allHosts {
+		kubeletName := GetCrtNameForHost(host, KubeletCertName)
+		kubeletCert := certs[kubeletName].Certificate
+		oldKubeletCSR := certs[kubeletName].CSR
+		kubeletAltNames := GetIPHostAltnamesForHost(host)
+		if oldKubeletCSR != nil &&
+			reflect.DeepEqual(kubeletAltNames.DNSNames, oldKubeletCSR.DNSNames) &&
```
alena1108: Question similar to one from corresponding rancher pr - are dnsnames and ips sorted?

superseb: Added sorting

pki/util.go
```diff
@@ -135,6 +135,34 @@ func GenerateCACertAndKey(commonName string, privateKey *rsa.PrivateKey) (*x509.
 	return kubeCACert, rootKey, nil
 }
 
+func GetIPHostAltnamesForHost(host *hosts.Host) *cert.AltNames {
+	ips := []net.IP{}
```
alena1108: nit: to avoid an unnecessary allocation, it's better to declare the slice in a form of `var ips []net.IP`. As a nil slice is functionally equivalent to a zero-length slice

PR #1707, opened 2019-10-18
Provide IP for kube-proxy if cloudprovider is set

superseb:
https://github.com/rancher/rancher/issues/22814

If cloudprovider is set (not empty), set the bind address because the node will not be able to retrieve it's IP address because the nodename could be set by the cloud provider (e.g. AWS and Openstack)

After running into other issues that had similar logic, we changed from checking for AWS only to if any cloud provider is set. We also added the check for `InternalAddress` being empty.


alena1108:
LGTM

PR #1724, opened 2019-10-28
Bind mount added for ipvs proxy mode

prachidamle:
https://github.com/rancher/rancher/issues/13769

### DIFF
--- a/cluster/plan.go
+++ b/cluster/plan.go
@@ -19,7 +19,7 @@ import (
 	"github.com/rancher/rke/pki"
 	"github.com/rancher/rke/services"
 	"github.com/rancher/rke/util"
-	"github.com/rancher/types/apis/management.cattle.io/v3"
+	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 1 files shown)


alena1108:
LGTM

cluster/plan.go
```diff
@@ -610,13 +610,15 @@ func (c *Cluster) BuildKubeProxyProcess(host *hosts.Host, prefixPath string, svc
 	Binds := []string{
 		fmt.Sprintf("%s:/etc/kubernetes:z", path.Join(prefixPath, "/etc/kubernetes")),
 		"/run:/run",
+		fmt.Sprintf("%s:/lib/modules:z", path.Join(prefixPath, "/lib/modules")),
```
kinarashah: Does this need to be read-only? @superseb @leodotcloud 

Asking cuz user mentioned 
```
First approach to solve this would be to add the missing volume of /lib/modules as read-only mount in that function.
```

superseb: Should only be needed to access (read), not anything else. Readonly should work fine.

cluster/plan.go
```diff
@@ -610,13 +610,15 @@ func (c *Cluster) BuildKubeProxyProcess(host *hosts.Host, prefixPath string, svc
...
 	}
 	if host.DockerInfo.OSType == "windows" { // compatible with Windows
 		Binds = []string{
 			// put the execution binaries to the host
 			fmt.Sprintf("%s:c:/host/etc/kubernetes", path.Join(prefixPath, "/etc/kubernetes")),
 			// exchange resources with other components
 			fmt.Sprintf("%s:c:/host/run", path.Join(prefixPath, "/run")),
+			fmt.Sprintf("%s:c:/host/lib/modules", path.Join(prefixPath, "/lib/modules")),
```
superseb: Is this needed for Windows?

prachidamle: I will keep it for both linux as well as windows, since we have the volume bind-mounted for both.

loganhz: Windows agent cannot start up due to this change. @thxCode should we remove this ?

prachidamle: I will be removing this change for Windows to unblock this failure.

Issue #1684, opened 2019-10-06
Error message when pulling new images should be removed


dnoland1:
**RKE version:**
```
$ rke -v
rke version v0.3.0
```
**Docker version: (`docker version`,`docker info` preferred)**
```
$ docker version
Client: Docker Engine - Community
 Version:           18.09.8
 API version:       1.39
 Go version:        go1.10.8
 Git commit:        0dd43dd87f
 Built:             Wed Jul 17 17:38:58 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          18.09.8
  API version:      1.39 (minimum version 1.12)
  Go version:       go1.10.8
  Git commit:       0dd43dd87f
  Built:            Wed Jul 17 17:48:49 2019
  OS/Arch:          linux/amd64
  Experimental:     false
```
**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)**
MacOS 10.14.6 (rke cli)
RancherOS 1.5.4 (nodes)

**Type/provider of hosts: (VirtualBox/Bare-metal/AWS/GCE/DO)**
AWS

**cluster.yml file:**
```yaml
authentication:
    sans:
      - "52.26.2.20"
      - "52.12.153.84"
      - "54.202.231.16"
nodes:
  - address: 52.26.2.20
    internal_address: 10.0.1.10
    user: rancher
    role:
      - worker
      - controlplane
      - etcd

  - address: 52.12.153.84
    internal_address: 10.0.1.11
    user: rancher
    role:
      - worker
      - controlplane
      - etcd

  - address: 54.202.231.16
    internal_address: 10.0.1.12
    user: rancher
    role:
      - worker
      - controlplane
      - etcd

ssh_agent_auth: true

services:
  etcd:
    snapshot: true
    creation: 1h
    retention: 240h
```
**Steps to Reproduce:**
run rke up

**Results:**
Appears in 0.3, first time images are pulled, it shows an "Error":
```
INFO[0000] Checking if image [rancher/rke-tools:v0.1.50] exists on host [52.26.2.20], try #1



INFO[0000] Image [rancher/rke-tools:v0.1.50] does not exist on host [52.26.2.20]: Error: No such image: rancher/rke-tools:v0.1.50
INFO[0000] Pulling image [rancher/rke-tools:v0.1.50] on host [52.26.2.20], try rancher/rke#1
```
[...]

PR #1736, opened 2019-10-30
Reduced log level to debug for image pull

alena1108:
https://github.com/rancher/rke/issues/1684

### DIFF
--- a/docker/docker.go
+++ b/docker/docker.go
@@ -217,23 +217,23 @@ func IsContainerRunning(ctx context.Context, dClient *client.Client, hostname st
 	return false, nil
 }
 
-func localImageExists(ctx context.Context, dClient *client.Client, hostname string, containerImage string) (bool, error) {
+func localImageExists(ctx context.Context, dClient *client.Client, hostname string, containerImage string) error {
 	var err error
 	for i := 1; i <= RetryCount; i++ {
-		logrus.Infof("Checking if image [%s] exists on host [%s], try #%d", containerImage, hostname, i)
+		logrus.Debugf("Checking if image [%s] exists on host [%s], try #%d", containerImage, hostname, i)
 		_, _, err = dClient.ImageInspectWithRaw(ctx, containerImage)
 		if err != nil {
 			if client.IsErrNotFound(err) {
-				logrus.Infof("Image [%s] does not exist on host [%s]: %v", containerImage, hostname, err)
-				return false, nil
+				logrus.Debugf("Image [%s] does not exist on host [%s]: %v", containerImage, hostname, err)
+				return err
 			}
-			logrus.Warnf("Error checking if image [%s] exists on host [%s]: %v", containerImage, hostname, err)
+			logrus.Debugf("Error checking if image [%s] exists on host [%s]: %v", containerImage, hostname, err)
 			continue
 		}
 		logrus.Infof("Image [%s] exists on host [%s]", containerImage, hostname)
-		return true, nil
+		return nil
 	}
-	return false, fmt.Errorf("Error checking if image [%s] exists on host [%s]: %v", containerImage, hostname, err)
+	return fmt.Errorf("Error checking if image [%s] exists on host [%s]: %v", containerImage, hostname, err)
 }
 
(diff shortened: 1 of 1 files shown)


docker/docker.go
```diff
@@ -217,23 +217,23 @@ func IsContainerRunning(ctx context.Context, dClient *client.Client, hostname st
...
 	var err error
 	for i := 1; i <= RetryCount; i++ {
 		logrus.Infof("Checking if image [%s] exists on host [%s], try #%d", containerImage, hostname, i)
 		_, _, err = dClient.ImageInspectWithRaw(ctx, containerImage)
 		if err != nil {
 			if client.IsErrNotFound(err) {
-				logrus.Infof("Image [%s] does not exist on host [%s]: %v", containerImage, hostname, err)
-				return false, nil
+				logrus.Debugf("Image [%s] does not exist on host [%s]: %v", containerImage, hostname, err)
+				return err
```
superseb: This is not an error situation, returning false, nil is correct as the function is `localImageExists` and the error from the Docker daemon is not found, meaning it does not exist and should return false.

docker/docker.go
```diff
@@ -217,23 +217,23 @@ func IsContainerRunning(ctx context.Context, dClient *client.Client, hostname st
...
 			}
-			logrus.Warnf("Error checking if image [%s] exists on host [%s]: %v", containerImage, hostname, err)
+			logrus.Debugf("Error checking if image [%s] exists on host [%s]: %v", containerImage, hostname, err)
 			continue
 		}
-		logrus.Infof("Image [%s] exists on host [%s]", containerImage, hostname)
-		return true, nil
+		logrus.Debugf("Image [%s] exists on host [%s]", containerImage, hostname)
```
superseb: This removes the feedback for `logrus.Infof("Checking if image [%s] exists on host [%s], try #%d", containerImage, hostname, i)`

alena1108:
@superseb 

> What is the reason to move from a function that was scoped to checking image locally to splitting it out to just returning error and doing another check (errnotfound in pullimage) in another function?

I think that notFound error is descriptive enough, and introducing one extra param - bool - to indicate whether the image exists, is redundant. Think of k8s getObject call - it returns an error if the object can't be found, and then it's up to the client to check for the error kind. It's also up to the client to decide on how to deal with NotFound type of error. I'm open for the argument :)

docker/docker.go
alena1108: I can add the info back for this one, no problem

docker/docker.go
```diff
@@ -217,23 +217,23 @@ func IsContainerRunning(ctx context.Context, dClient *client.Client, hostname st
...
-			logrus.Warnf("Error checking if image [%s] exists on host [%s]: %v", containerImage, hostname, err)
+			logrus.Debugf("Error checking if image [%s] exists on host [%s]: %v", containerImage, hostname, err)
 			continue
 		}
 		logrus.Infof("Image [%s] exists on host [%s]", containerImage, hostname)
-		return true, nil
+		return nil
 	}
-	return false, fmt.Errorf("Error checking if image [%s] exists on host [%s]: %v", containerImage, hostname, err)
+	return fmt.Errorf("error checking if image [%s] exists on host [%s]: %v", containerImage, hostname, err)
```
superseb: Why is this `error` and the rest is `Error`?

superseb:
The current changes don't line up with this comment: https://github.com/rancher/rke/issues/1684#issuecomment-539428464

I think we should either move them all to debug, or have feedback on the actions and just strip the error as the issue was that Docker reported it as error while we logged it as info

docker/docker.go
alena1108: @superseb we should fix the rest. It should align with the go best practices where the error should be in lower case

alena1108:
@superseb 

>I think we should either move them all to debug, or have feedback on the actions and just strip the error as the issue was that Docker reported it as error while we logged it as info

reduced level to debug in localImageExists, https://github.com/rancher/rke/pull/1736/files#diff-b715a530d535012e418a564a3eafedb9R223

superseb:
LGTM

Issue #1123, opened 2019-02-07
Explore options to check date/time between nodes


superseb:
**Steps to Reproduce:**
* Run `rke up` against nodes with date/time not in sync
* Run into various issues

**Results:**

Logging observed:

```
FATA[0305] [controlPlane] Failed to bring up Control Plane: Failed to verify healthcheck: Failed to check https://localhost:6443/healthz for service [kube-apiserver] on host [10.0.0.2]: Get https://localhost:6443/healthz: EOF, log: I0207 10:28:17.110391       1 plugins.go:161] Loaded 6 validating admission controller(s) successfully in the following order: LimitRanger,ServiceAccount,Priority,PersistentVolumeClaimResize,ValidatingAdmissionWebhook,ResourceQuota.
```

```
F0207 11:17:33.848659       1 storage_decorator.go:57] Unable to create storage backend: config (&{etcd3 /registry [https://X:2379 https://X:2379 https://X:2379] /etc/kubernetes/ssl/kube-node-key.pem /etc/kubernetes/ssl/kube-node.pem /etc/kubernetes/ssl/kube-ca.pem true false 1000 0xc4202b7180 <nil> 5m0s 1m0s}), err (context deadline exceeded)
```

```
2019-02-07 13:05:55.848860 I | etcdmain: rejected connection from "x:x" (error "tls: failed to verify client's certificate: x509: certificate has expired or is not yet valid", ServerName "")
```

acaiblog:
问下这个问题该怎么解决呢？我也碰到了

acaiblog:
ntp时间不同步会导致这个问题

PR #1752, opened 2019-11-05
Add time checking on cluster provisioning

superseb:
https://github.com/rancher/rke/issues/1123

For Rancher, this needs a vendor of https://github.com/rancher/kontainer-engine/pull/196

How this works:

 Time on provisioning end (host where RKE up is ran or where the Rancher container/pod) is taken, and compared to the time on the nodes in the cluster. We error out as soon as the time on the nodes is earlier than the provisioning end as the certificate will not be valid yet. If it's later on the nodes, we log a warning as it might still work and we don't want to break directly.

### DIFF
--- a/cluster/defaults.go
+++ b/cluster/defaults.go
@@ -86,6 +86,7 @@ type ExternalFlags struct {
 	ConfigDir        string
 	CustomCerts      bool
 	DisablePortCheck bool
+	DisableTimeCheck bool
 	GenerateCSR      bool
 	Local            bool
 	UpdateOnly       bool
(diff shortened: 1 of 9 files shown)


cluster/defaults.go
```diff
@@ -86,6 +86,7 @@ type ExternalFlags struct {
 	ConfigDir        string
 	CustomCerts      bool
 	DisablePortCheck bool
+	DisableTimeCheck bool
```
alena1108: do you think we should enable it in types?

superseb: This is identically setup as the port check, if it needs to be changed, both need to be moved.

cluster/timecheck.go
```diff
@@ -0,0 +1,94 @@
...
+const (
+	TimeCheckContainer = "rke-time-checker"
+)
+
+func (c *Cluster) CheckClusterTime(ctx context.Context, currentCluster *Cluster) error {
+	allHosts := hosts.GetUniqueHostList(c.EtcdHosts, c.ControlPlaneHosts, c.WorkerHosts)
+	var errgrp errgroup.Group
+	hostsQueue := util.GetObjectQueue(allHosts)
+	for w := 0; w < WorkerThreads; w++ {
+		errgrp.Go(func() error {
```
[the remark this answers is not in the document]
superseb: I guess this needs to change for every occurrence of this logic, which is in a lot of places based on a quick search.

Issue #1104, opened 2019-01-28
RKE should allow setting extra_env/etc. on the ingress provider


Oats87:
Currently, it is not possible to set extra_env/extra_binds on the ingress provider. We should allow this functionality, in order to allow users to perform customizations to the nginx ingress controller. This can include things like adding proxies, adding custom file-based certs, etc.

superseb:
https://github.com/rancher/rancher/issues/16868

cloudnautique:
This should be under similar keys to the upstream helm chart: https://github.com/helm/charts/tree/master/stable/nginx-ingress

Specifically the functionality introduced by:
`controller.extraVolumeMounts`
`controller.extraEnvs`

mrajashree:
The ingress config now accepts three more fields: ExtraEnvs, ExtraVolumes and ExtraVolumeMounts.
They refer the k8s native types EnvVar, Volume and VolumeMount respectively.

For example:
```
  ingress:
    provider: nginx
    extra_envs:
      - name: MY_NODE_NAME
        valueFrom:
          fieldRef:
            fieldPath: spec.nodeName
      - name: testenvvar
        value: testvalue
    extra_volume_mounts:
      - mountPath: /testmountpath
        name: testvol
    extra_volumes:
      - emptyDir: {}
        name: testvol
```

This should result in creating 2 env vars in the ingress nginx pod:
1. $MY_NODE_NAME with the node name on which it's running as the value
2. $testenvvar with `testvalue` as value.
3. A dir named testmountpath will be created within the pod at `/testmountpath`

soumyalj:
Tested with rancher version v2.3-head from Nov 11
Created a DO cluster and specified the extra_envs, extra_volume_mounts, extra_volumes in cluster.yml 
```
ingress:
    provider: nginx
    extra_envs:
      - name: soumyanode1
        valueFrom:
          fieldRef:
            fieldPath: spec.nodeName
      - name: testenvvar
        value: testvalue
    extra_volume_mounts:
      - mountPath: /testmountpath
        name: testvol
    extra_volumes:
      - emptyDir: {}
        name: testvol
```
The cluster got created successfully. In the pod , `nginx-ingress-controller-g8jqp`, the volume mount `/testmountpath` was created as expected. Read/Write to the volume was successful. 
The env variables were also verified as below in the pod by typing `env` command
```
KUBERNETES_SERVICE_PORT_HTTPS=443
KUBERNETES_SERVICE_PORT=443
[...]

PR #1757, opened 2019-11-06
Accept extraEnv, volumes and volumeMounts for ingress addon

mrajashree:
https://github.com/rancher/rke/issues/1104
Accept extraEnv, volumes and volumeMounts for ingress addon

The fields for ExtraEnv, extraVolumes and extraVolumeMounts for ingress
addon refer the k8s native types EnvVar, Volume and VolumeMounts.
The k8s native types have json tags, so this commit adds a template func to
first marshal and get json encoding and then convert to yaml.

Requires: https://github.com/rancher/types/pull/1024/files


templates/templates.go
```diff
@@ -34,6 +40,20 @@ func GetKubednsStubDomains(stubDomains map[string][]string) string {
 	return string(json)
 }
 
+func ToYAML(v interface{}) string {
+	data, err := json.Marshal(v)
+	if err != nil {
+		// Swallow errors inside of a template so it doesn't affect remaining template lines
```
alena1108: should we at least log the error here?

templates/templates.go
```diff
@@ -34,6 +40,20 @@ func GetKubednsStubDomains(stubDomains map[string][]string) string {
...
+		return ""
+	}
+	yamlData, err := yaml.JSONToYAML(data)
+	if err != nil {
+		// Swallow errors inside of a template so it doesn't affect remaining template lines
```
alena1108: same here

Issue #1232, opened 2019-03-27
Feature Request: Ability to set base directory for private registry


patrick0057:
**RKE version:**
0.1.17

**Docker version: (`docker version`,`docker info` preferred)**
[root@rkalvrsiel9 ~]# docker version
Client:
 Version:           18.06.2-ce
 API version:       1.38
 Go version:        go1.10.3
 Git commit:        6d37f41
 Built:             Sun Feb 10 03:46:03 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server:
 Engine:
  Version:          18.06.2-ce
  API version:      1.38 (minimum version 1.12)
  Go version:       go1.10.3
  Git commit:       6d37f41
  Built:            Sun Feb 10 03:48:29 2019
  OS/Arch:          linux/amd64
  Experimental:     false

[root@rkalvrsiel9 ~]# docker info
Containers: 27
 Running: 17
 Paused: 0
 Stopped: 10
Images: 10
Server Version: 18.06.2-ce
Storage Driver: overlay2
 Backing Filesystem: xfs
 Supports d_type: true
 Native Overlay Diff: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge host macvlan null overlay
 Log: awslogs fluentd gcplogs gelf journald json-file logentries splunk syslog
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: 468a545b9edcd5932818eb9de8e72413e616e86e
runc version: 69663f0bd4b60df09991c08812a60108003fa340
init version: fec3683
Security Options:
 seccomp
  Profile: default
Kernel Version: 3.10.0-957.1.3.el7.x86_64
Operating System: CentOS Linux 7 (Core)
OSType: linux
Architecture: x86_64
CPUs: 4
Total Memory: 3.701GiB
Name: rkalvrsiel9.kau.roche.com
ID: 7SCW:CJOX:Q346:PB56:CZMP:XGAB:KLDL:7JF6:SFVA:NH3T:P37U:JSHZ
Docker Root Dir: /var/lib/docker
Debug Mode (client): false
Debug Mode (server): false
Registry: https://index.docker.io/v1/
Labels:
Experimental: false
Insecure Registries:
 127.0.0.0/8
Live Restore Enabled: false

**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)**

[...]

PR #1759, opened 2019-11-06
Nested private registry lookup support

alena1108:
https://github.com/rancher/rke/issues/1232


docker/docker.go
```diff
@@ -634,16 +638,53 @@ func getRegistryAuth(pr v3.PrivateRegistry) (string, error) {
 }
 
 func GetImageRegistryConfig(image string, prsMap map[string]v3.PrivateRegistry) (string, string, error) {
+	/*
+		Image can be passed as
+		- repo.com/foo/bar/rancher/rke-tools:v0.1.51
```
leodotcloud: Does adding a test make sense to test this function?

docker/docker.go
```diff
@@ -634,16 +638,53 @@ func getRegistryAuth(pr v3.PrivateRegistry) (string, error) {
...
+	if len(prsMap) == 0 {
+		return "", "", nil
+	}
 	regURL := ref.Domain(namedImage)
-	if pr, ok := prsMap[regURL]; ok {
-		// We do this if we have some docker.io login information
-		regAuth, err := getRegistryAuth(pr)
-		return regAuth, pr.URL, err
+	regPath := ref.Path(namedImage)
+	splitPath := strings.Split(regPath, "/")
```
leodotcloud: If the user has provided `repo.com/foo/bar` as the registry URL, why do we have to check repo.com/foo and repo.com? That seems like an unintended side/extra behavior and we would have to document this behavior if we want to provide this functionality. If the user intends to use these different search paths, it's better to have the same explicitly defined in the config.

alena1108: @leodotcloud the logic works this way:

* system images can be coming from diff repos (although rarely used, we've always had support for it in rke)
* For every image, we do the repo lookup. If we are to add search path explicitly, you'd have to do it for each and every image, and that's more cumbersome in my opinion.

docker/docker.go
alena1108: @leodotcloud I can try adding a unittest. Don't think adding integration test would be possible with our current framework as we don't have long standing (and maintained) private repo

PR #1769, opened 2019-11-12
Handle unmarshal of ingressConfig's k8s native fields separately

mrajashree:
IngressConfig fields ExtraEnvs, ExtraVolumes and ExtraVolumeMounts are k8s types.
The yaml unmarshal done in ParseConfig can't unmarshal these properly because they have
nested fields without yaml tags. This commit adds logic to unmarshal
these fields separately so all nested fields get unmarshaled too.

https://github.com/rancher/rke/issues/1104

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/docker/docker/api/types"
 	ghodssyaml "github.com/ghodss/yaml"
+	"github.com/rancher/norman/types/convert"
 	"github.com/rancher/rke/authz"
 	"github.com/rancher/rke/docker"
 	"github.com/rancher/rke/hosts"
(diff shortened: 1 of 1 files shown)


cluster/cluster.go
```diff
@@ -250,6 +251,70 @@ func parseAdmissionConfig(clusterFile string, rkeConfig *v3.RancherKubernetesEng
...
+	err = json.Unmarshal(ingressEnvBytes, &envs)
+	if err != nil {
+		return fmt.Errorf("[parseIngressExtraEnv] error unmarshaling ingress config extraEnvs: %v", err)
+	}
+	rkeConfig.Ingress.ExtraEnvs = envs
+	return nil
+}
+
+func parseIngressExtraVolumes(ingressMap map[string]interface{}, rkeConfig *v3.RancherKubernetesEngineConfig) error {
+	ingressVolBytes, err := json.Marshal(ingressMap["extra_volumes"])
```
kinarashah: do we need to check if key is present in ingressMap before marshal?

alena1108:
LGTM

PR #1771, opened 2019-11-13
Enable option to exclude hosts from RKE up

levigross:


alena1108:
@levigross just curious - what's the use case you are targeting?

levigross:
At the moment, rke will apply an update too all machines in a cluster. This option helps when a person would like to selectively update (or not update) certain nodes to test the upgrade.

PR #1782, opened 2019-11-14
respect metadata's service options for etcd

kinarashah:


cluster/plan.go
```diff
@@ -953,12 +951,32 @@ func (c *Cluster) BuildEtcdProcess(host *hosts.Host, etcdHosts []*hosts.Host, pr
...
+
 	for arg, value := range c.Services.Etcd.ExtraArgs {
 		if _, ok := c.Services.Etcd.ExtraArgs[arg]; ok {
 			CommandArgs[arg] = value
 		}
 	}
 
+	// adding the old default value from L922 if not present in metadata options or passed by user
+	if _, ok := CommandArgs["client-cert-auth"]; !ok {
+		args = append(args, "--client-cert-auth")
```
prachidamle: Shoud this be `--client-cert-auth=true`?

cluster/plan.go
```diff
@@ -953,12 +951,32 @@ func (c *Cluster) BuildEtcdProcess(host *hosts.Host, etcdHosts []*hosts.Host, pr
...
+	}
+	if _, ok := CommandArgs["peer-client-cert-auth"]; !ok {
+		args = append(args, "--peer-client-cert-auth")
```
prachidamle: Same - `--peer-client-cert-auth=true`?

cluster/plan.go
kinarashah: this is to preserve the old behavior for existing clusters, otherwise they'll get auto upgraded and we don't want that

cluster/plan.go
kinarashah: this is to preserve the old behavior for existing clusters, otherwise they'll get auto upgraded and we don't want that

PR #1783, opened 2019-11-15
v0.3: add admission control config file arg

leodotcloud:


cluster/plan.go
```diff
@@ -269,6 +269,11 @@ func (c *Cluster) BuildKubeAPIProcess(host *hosts.Host, prefixPath string, svcOp
 		CommandArgs[admissionControlOptionName] = CommandArgs[admissionControlOptionName] + ",AlwaysPullImages"
 	}
 
+	if c.Services.KubeAPI.EventRateLimit != nil && c.Services.KubeAPI.EventRateLimit.Enabled {
+		CommandArgs[KubeAPIArgAdmissionControlConfigFile] = DefaultKubeAPIArgAdmissionControlConfigFileValue
```
prachidamle: @leodotcloud I don't understand why we need to add this only for EventRateLimit, I dont see we specify this for other admission control options.

prachidamle:
LGTM

PR #1784, opened 2019-11-15
add admission control config file arg

leodotcloud:


cluster/plan.go
```diff
@@ -267,6 +267,11 @@ func (c *Cluster) BuildKubeAPIProcess(host *hosts.Host, prefixPath string, svcOp
 		CommandArgs[admissionControlOptionName] = CommandArgs[admissionControlOptionName] + ",AlwaysPullImages"
 	}
 
+	if c.Services.KubeAPI.EventRateLimit != nil && c.Services.KubeAPI.EventRateLimit.Enabled {
+		CommandArgs[KubeAPIArgAdmissionControlConfigFile] = DefaultKubeAPIArgAdmissionControlConfigFileValue
```
prachidamle: Why do we add this only for EventRateLimit? Is this file passed otherwise when EventRateLimit is off?

leodotcloud: Multiple admission controllers configs can be passed in this single file. For now, when EventRateLimit is enabled, we need to pass this flag.

prachidamle:
LGTM

PR #1785, opened 2019-11-15
Added changes to support ACI CNI installation

Vikash082:


leodotcloud:
The CNI drivers currently available in the code are tested thoroughly for every release. It would be hard for us to test all possible network plugins available. But, we do not limit our users, they can bring in any network plugin they want. Here are the steps: https://rancher.com/docs/rke/latest/en/config-options/add-ons/network-plugins/custom-network-plugin-example/

superseb:
This will be covered in https://github.com/rancher/rke/pull/2159

PR #1792, opened 2019-11-20
Parse updateStrategy, replicas and autoscaler fields

mrajashree:
https://github.com/rancher/rancher/issues/23907

Requires:
Types changes: https://github.com/rancher/types/pull/1036
kontainer-driver-metadata changes: https://github.com/rancher/kontainer-driver-metadata/pull/83


cluster/addons.go
```diff
@@ -42,6 +44,9 @@ const (
 
 	CoreDNSProvider = "coredns"
 	KubeDNSProvider = "kube-dns"
+
+	// DNSAutoscalerLinearModeParams has previously hardcoded value of linear mode cluster-proportional-autoscaler in kontainer-driver-metadata
+	DNSAutoscalerLinearModeParams = `{"coresPerReplica":128,"nodesPerReplica":4,"min":1,"preventSinglePointFailure":true}`
```
alena1108: any reason why we are hardcoding it in rke? Shouldn't it be defined in k-d-m?

cluster/addons.go
```diff
@@ -313,6 +339,20 @@ func (c *Cluster) deployCoreDNS(ctx context.Context, data map[string]interface{}
 	return nil
 }
 
+func getDNSAutoscalerLinearModeParams(LinearAutoscalerParams *v3.LinearAutoscalerParams) (string, error) {
+	// The dns autoscaler uses linear mode of cluster-proportional-autoscaler to determine number of replicas
```
alena1108: would it be possible to define this logic in k-d-m template? Something like: `if !params then respect linearmode`

cluster/addons.go
```diff
@@ -341,6 +381,13 @@ func (c *Cluster) deployMetricServer(ctx context.Context, data map[string]interf
 		Options:            c.Monitoring.Options,
 		NodeSelector:       c.Monitoring.NodeSelector,
 		Version:            util.GetTagMajorVersion(versionTag),
+		UpdateStrategy:     c.Monitoring.UpdateStrategy,
+		Replicas:           c.Monitoring.Replicas,
+	}
+	if MetricsServerConfig.Replicas == 0 {
+		// if user didn't provide replica count, it will default to 0 and metrics-server pods won't get created.
+		// So as a default set replica number to 2 (as opposed to regular default of 1 in order to increase scale)
+		MetricsServerConfig.Replicas = 2
```
alena1108: same here. Can we move this logic to k-d-m?

PR #1797, opened 2019-11-21
change enc provider arg, update defaults for audit log config

leodotcloud:
https://github.com/rancherlabs/rancher-security/issues/360


leodotcloud:
https://github.com/rancherlabs/rancher-security/issues/362

Issue #1772, opened 2019-11-13
Change upgrade logic for worker nodes


mrajashree:
Worker nodes should be upgraded in user configurable batches. Users can specify maximum number of nodes that can be unavailable during the upgrade. The upgrade will follow a sliding window approach based on maxUnavailable value users specify.
The nodes will be cordoned before upgrade. Users can also optionally drain the node.
Nodes will be uncordoned after upgrade.
Users can specify maxUnavailable and drain input through cluster.yml
```
upgrade_strategy:
  max_unavailable: 1
  drain: true
  node_drain_input:
    ignore_daemonsets: true
    force:
    delete_local_data:
    grace_period:
    timeout:
```

soumyalj:
Tested with rke cli - v1.1.0-rc5
Some of the tests executed :
1. Verified that worker nodes are upgraded in batches . maxUnavailable can be configured which defines the number of nodes that get upgraded at a time. 
2. Verified worker node upgrade by specifying the drain=true
```
 upgrade_strategy:
  max_unavailable: 3
  drain: true
  node_drain_input:
    force: False
    delete_local_data: false
    grace_period: 30
    timeout:120
```
3. Update cluster and specify drain=true option and change kubelet flag or change any parameter other than K8s version.
4. Drain failure during cluster update Drain failure during cluster update
5. Add a worker node in rke file and then upgrade specifying maxUnavailable field
6. Remove a worker node in rke file and then upgrade specifying maxUnavailable field

Worker node upgrade succeeds in all the cases as expected. The nodes have updated images after upgrade (when there is a change in image)

Issue #1734, opened 2019-10-30
change upgrade flow for control plane nodes


kinarashah:
Upgrades of the Kubernetes control plane should be handled such that, 

- Control plane nodes should be upgraded 1 node at a time.
- The control plane node should come completely back up before moving to the next. 

https://github.com/rancher/rancher/issues/23038

Michaellaneous:
As a quick question, as I am not entirely sure:
Will this include rancher launched cluster on other infrastructure providers? Will you be able to to the same for worker nodes as well?

soumyalj:
Tested  using RKE version v1.1.0-rc6. 

Some of the test cases executed:

1. Create a RKE cluster with K8s version v1.16.6 using 3 control nodes, 1 etcd and 3 workers. Upgrade k8s version to v1.17.2 in cluster.yml file and perform `./rke up --config cluster.yml` 
Verified that control plane nodes are upgraded one at a time and the node is completely up before moving to the next.
2. Log  in to the control plane nodes and verify that the control node components get upgraded to the correct version. The following components were verified for new version:
a. kube-controller-manager (rancher/hyperkube)
b. kube-apiserver(rancher/hyperkube)
c. kube-scheduler(rancher/hyperkube)

2. Add a new control node in the cluster.yml file and then initiate an upgrade. Upgrade succeeds
3. Remove a control node in the cluster.yml file and then initiate an upgrade. Upgrade succeeds

PR #1800, opened 2019-11-25
Change RKE upgrade logic for zero downtime

mrajashree:
https://github.com/rancher/rke/issues/1772
Change worker plane components upgrade strategy for zero downtime upgrades
- [x] Accept maxUnavailable from user, default to 10%, round down
- [x] Calculate powered down/unreachable hosts first (already done by `TunnelHosts`). If number of unreachable hosts = maxUnavailable, stop upgrade
- [x] Adjust maxUnavailable according to number of unreachable nodes. If maxUnavailable is 10 and 3 nodes are unreachable, maxUnavailable for actual upgrade will be 7
- [x] Upgrade worker components on nodes with etcd/controlplane roles first and one at a time
- [x] Upgrade nodes in sliding window of size maxUnavailable. So if maxUnavailable=10, 10 nodes can start upgrading in parallel, as soon as `x` out of 10 are done, start upgrading the next `x` nodes
- [x] Cordon a node before upgrade, give user option to drain it too. Uncordon node after upgrade.
- [x] Node is considered upgraded when it can be listed using kube client and its status is Ready 
- [x] Keep saving names of nodes that get an error during upgrade.
- [x] If maxUnavailable nodes get error during upgrade, stop the upgrade process

For clusters with a large number of nodes, upgrading a percentage of them based on maxUnavailable will lead to multiple goroutines and errors due to that. [This](https://github.com/rancher/rke/issues/953) issue has details about it and why RKE switched to worker pool. 
So maxUnavailable will be respected as long as it's not too big and capped at 50 which is the current worker threads RKE uses

https://github.com/rancher/rke/issues/1734
Upgrade controlplane components one by one for zero downtime upgrades

Types PR for drain input: https://github.com/rancher/types/pull/1069


riaan53:
Does this include the ability to drain nodes before an upgrade to do a graceful rolling cluster upgrade?

cmd/up.go
```diff
@@ -65,6 +65,14 @@ func UpCommand() cli.Command {
 			Name:  "custom-certs",
 			Usage: "Use custom certificates from a cert dir",
 		},
+		cli.StringFlag{
```
alena1108: We should be *very* consistent with rancher upgrade strategy. From what I recall, we've agreed on having maxUnavailable parameter in leu of batch interval and size

mrajashree: yes, I'm yet to change it

cluster/cluster.go
```diff
@@ -120,35 +120,77 @@ func (c *Cluster) DeployControlPlane(ctx context.Context, svcOptionData map[stri
...
+		if err := services.RunControlPlane(ctx, c.ControlPlaneHosts,
+			c.LocalConnDialerFactory,
+			c.PrivateRegistriesMap,
+			cpNodePlanMap,
+			c.UpdateWorkersOnly,
+			c.SystemImages.Alpine,
+			c.Certificates); err != nil {
+			return fmt.Errorf("[controlPlane] Failed to bring up Control Plane: %v", err)
+		}
+	} else {
```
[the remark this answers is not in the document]
mrajashree: we only return from the error case of that if, but I'll add `return nil` to the non error case and still get rid of else

services/workerplane.go
```diff
@@ -42,6 +53,187 @@ func RunWorkerPlane(ctx context.Context, allHosts []*hosts.Host, localConnDialer
...
+							}
+							hostsFailedToUpgrade <- runHost.HostnameOverride
+						}
+					}
+					if maxUnavailableHit {
+						break
+					}
+				}
+				// when cluster is upgrading, for each node check if it's ready, cordon it and drain optionally
+				if err := checkNodeReady(kubeClient, runHost, WorkerRole); err != nil {
```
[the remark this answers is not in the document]
mrajashree: The above loop is for checking all nodes to find which ones are currently unavailable. I can include check for this node too in that.

cluster/defaults.go
```diff
@@ -188,10 +189,43 @@ func (c *Cluster) setClusterDefaults(ctx context.Context, flags ExternalFlags) e
...
+	c.setNodeUpgradeStrategy()
 
 	return nil
 }
 
+func (c *Cluster) setNodeUpgradeStrategy() {
+	if c.UpgradeStrategy == nil {
+		logrus.Info("No input provided for maxUnavailable, setting it to default value of 50%")
+		c.UpgradeStrategy = &v3.NodeUpgradeStrategy{
+			MaxUnavailable: "50%",
```
superseb: This needs to be changed now according to latest types changes

services/workerplane.go
```diff
@@ -42,6 +53,187 @@ func RunWorkerPlane(ctx context.Context, allHosts []*hosts.Host, localConnDialer
 	return nil
 }
 
+func UpgradeWorkerPlane(ctx context.Context, kubeClient *kubernetes.Clientset, multipleRolesHosts []*hosts.Host, workerOnlyHosts []*hosts.Host, inactiveHosts []*hosts.Host, localConnDialerFactory hosts.DialerFactory, prsMap map[string]v3.PrivateRegistry, workerNodePlanMap map[string]v3.RKEConfigNodePlan, certMap map[string]pki.CertificatePKI, updateWorkersOnly bool, alpineImage string, upgradeStrategy *v3.NodeUpgradeStrategy) error {
+	log.Infof(ctx, "[%s] Upgrading Worker Plane..", WorkerRole)
+	maxUnavailable, err := strconv.Atoi(upgradeStrategy.MaxUnavailable)
+	if err != nil {
+		return err
+	}
+	logrus.Infof("Parsed value of maxUnavailable: %v", maxUnavailable)
```
superseb: We can remove this now its in `CalculateMaxUnavailable`

cmd/up.go
```diff
@@ -118,6 +119,10 @@ func ClusterUp(ctx context.Context, dialersOptions hosts.DialersOptions, flags c
 		return APIURL, caCrt, clientCert, clientKey, nil, err
 	}
 
+	if currentCluster != nil {
+		isUpgrade = true
```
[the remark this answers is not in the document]
superseb: I actually meant to indicate to the user that we mark this run as an upgrade (`isUpgrade = true`)

services/workerplane.go
```diff
@@ -42,6 +53,187 @@ func RunWorkerPlane(ctx context.Context, allHosts []*hosts.Host, localConnDialer
...
+				if err := checkNodeReady(kubeClient, runHost, WorkerRole); err != nil {
+					errList = append(errList, err)
+					if len(hostsFailedToUpgrade) == maxUnavailable {
+						break
+					}
+					hostsFailedToUpgrade <- runHost.HostnameOverride
+					continue
+				}
+				// cordon and drain
+				if err := cordonAndDrainNode(kubeClient, runHost, upgradeStrategy.Drain, drainHelper, WorkerRole); err != nil {
```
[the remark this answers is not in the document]
mrajashree: not needed here

cluster/defaults.go
```diff
@@ -188,10 +189,42 @@ func (c *Cluster) setClusterDefaults(ctx context.Context, flags ExternalFlags) e
...
+	}
+	setDefaultIfEmpty(&c.UpgradeStrategy.MaxUnavailable, "10%")
+	if !c.UpgradeStrategy.Drain {
+		return
+	}
+	if c.UpgradeStrategy.DrainInput == nil {
+		c.UpgradeStrategy.DrainInput = &v3.NodeDrainInput{
+			IgnoreDaemonSets: true,
+			// default to 120 seems to work better for controlplane nodes
+			Timeout: 120,
```
kinarashah: this could possibly be a nit, but we have all default consts stored at the start of file, could we put these values there as well?

cluster/validation.go
```diff
@@ -196,6 +198,40 @@ func ValidateHostCount(c *Cluster) error {
...
+	}
+
+	maxUnavailable, err := services.CalculateMaxUnavailable(k8sutil.Parse(c.UpgradeStrategy.MaxUnavailable), workerOnlyHosts)
+	if err != nil {
+		return err
+	}
+	if len(inactiveWorkerOnlyHosts) >= maxUnavailable {
+		return fmt.Errorf("cannot proceed with upgrade of worker components since %v (>=maxUnavailable) hosts are inactive", strings.Join(inactiveWorkerOnlyHosts, ","))
+	}
+	c.UpgradeStrategy.MaxUnavailable = strconv.Itoa(maxUnavailable)
```
kinarashah: I don't think we should be updating values on cluster in validation file

services/controlplane.go
```diff
@@ -39,6 +44,68 @@ func RunControlPlane(ctx context.Context, controlHosts []*hosts.Host, localConnD
 	return nil
 }
 
+func UpgradeControlPlane(ctx context.Context, kubeClient *kubernetes.Clientset, controlHosts []*hosts.Host, localConnDialerFactory hosts.DialerFactory, prsMap map[string]v3.PrivateRegistry, cpNodePlanMap map[string]v3.RKEConfigNodePlan, updateWorkersOnly bool, alpineImage string, certMap map[string]pki.CertificatePKI, upgradeStrategy *v3.NodeUpgradeStrategy, newHosts map[string]bool) error {
+	if updateWorkersOnly {
+		return nil
+	}
+	var drainHelper drain.Helper
+
+	log.Infof(ctx, "[%s] Processing control plane components for upgrade one at a time", ControlRole)
```
kinarashah: skip if newHost

services/controlplane.go
```diff
@@ -39,6 +44,68 @@ func RunControlPlane(ctx context.Context, controlHosts []*hosts.Host, localConnD
...
+				if !k8s.IsNodeReady(node) {
+					maxUnavailableHit = true
+					break
+				}
+			}
+			if maxUnavailableHit {
+				return err
+			}
+		}
+		upgradable, err := isControlPlaneHostUpgradable(ctx, host, cpNodePlanMap[host.Address].Processes)
```
kinarashah: are we supposed to check this for newHosts as well? because otherwise, I think this would make it easier to read? 
```
if newHosts[hostName] {
    deployControl()
    continue
}
// entire flow
```

services/workerplane.go
```diff
@@ -42,6 +52,168 @@ func RunWorkerPlane(ctx context.Context, allHosts []*hosts.Host, localConnDialer
...
+	}
+	if maxUnavailable > WorkerThreads {
+		/* upgrading a large number of nodes in parallel leads to a large number of goroutines, which has led to errors regarding too many open sockets
+		Because of this RKE switched to using workerpools. 50 workerthreads has been sufficient to optimize rke up, upgrading at most 50 nodes in parallel.
+		So the user configurable maxUnavailable will be respected only as long as it's less than 50 and capped at 50 */
+		maxUnavailable = WorkerThreads
+		logrus.Info("Setting maxUnavailable to 50, to avoid issues related to upgrading large number of nodes in parallel")
+	}
+
+	log.Infof(ctx, fmt.Sprintf("Processing %v hosts for upgrade at a time", maxUnavailable))
```
kinarashah: this line should be logged after L70 if I'm getting it right?

services/workerplane.go
```diff
@@ -42,6 +52,168 @@ func RunWorkerPlane(ctx context.Context, allHosts []*hosts.Host, localConnDialer
...
+	maxUnavailable -= len(inactiveHosts)
+	log.Infof(ctx, "First checking and upgrading worker components on nodes with etcd/controlplane roles one at a time")
+	multipleRolesHostsFailedToUpgrade, err := startWorkerPlane(ctx, kubeClient, multipleRolesHosts, localConnDialerFactory, prsMap, workerNodePlanMap, certMap, updateWorkersOnly, alpineImage, 1, upgradeStrategy, newHosts)
+	if err != nil && len(multipleRolesHostsFailedToUpgrade) == maxUnavailable {
+		logrus.Errorf("Failed to upgrade hosts: %v", strings.Join(multipleRolesHostsFailedToUpgrade, ","))
+		return err
+	}
+
+	maxUnavailable -= len(multipleRolesHostsFailedToUpgrade)
+	log.Infof(ctx, "Now checking and upgrading worker components on nodes with only worker role")
```
kinarashah: `%v, maxUnavailable` worker components? that way we can see the changing count

services/controlplane.go
mrajashree: adding a separate log for it

services/workerplane.go
```diff
@@ -42,6 +53,187 @@ func RunWorkerPlane(ctx context.Context, allHosts []*hosts.Host, localConnDialer
...
+	var semHostsFailedOrCurrentlyProcessing = semaphore.NewWeighted(int64(maxUnavailable))
+
+	hostsQueue := util.GetObjectQueue(allHosts)
+	if upgradeStrategy.Drain {
+		drainHelper = getDrainHelper(kubeClient, *upgradeStrategy)
+	}
+	semCtx, semCancel := context.WithCancel(ctx)
+
+	/* Each worker thread starts a goroutine that reads the hostsQueue channel in a for loop
+	Using same number of worker threads as maxUnavailable ensures only maxUnavailable number of nodes are being processed at a time
```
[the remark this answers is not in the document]
mrajashree: I'm not using semaphore anymore. instead I'll quit a worker goroutine if it fails to process a host, that way maxUnavailable is respected. If n hosts fail, then only maxUnavailable-n goroutines will run

services/workerplane.go
```diff
@@ -42,6 +50,160 @@ func RunWorkerPlane(ctx context.Context, allHosts []*hosts.Host, localConnDialer
...
+					break
+				}
+				upgradable, err := isWorkerHostUpgradable(ctx, runHost, workerNodePlanMap[runHost.Address].Processes)
+				if err != nil {
+					errList = append(errList, err)
+					hostsFailedToUpgrade <- runHost.HostnameOverride
+					break
+				}
+				if !upgradable {
+					logrus.Debugf("[workerplane] Upgrade not required for worker components of host %v", runHost.HostnameOverride)
```
superseb: Can we change this to `Info`, similar to controlplane

services/workerplane.go
```diff
@@ -42,6 +50,160 @@ func RunWorkerPlane(ctx context.Context, allHosts []*hosts.Host, localConnDialer
...
+				nodes, err := k8s.GetNodeList(kubeClient)
+				if err != nil {
+					errList = append(errList, err)
+					hostsFailedToUpgrade <- runHost.HostnameOverride
+					break
+				}
+				var maxUnavailableHit bool
+				for _, node := range nodes.Items {
+					// in case any previously added nodes or till now unprocessed nodes become unreachable during upgrade
+					if !k8s.IsNodeReady(node) {
```
kinarashah: this could be true for previously failed nodes too, we'll need to keep track of this when it fails from

services/workerplane.go
```diff
@@ -42,6 +51,170 @@ func RunWorkerPlane(ctx context.Context, allHosts []*hosts.Host, localConnDialer
...
+	return nil
+}
+
+func CalculateMaxUnavailable(maxUnavailableVal string, numHosts int) (int, error) {
+	// if maxUnavailable is given in percent, round down
+	maxUnavailableParsed := k8sutil.Parse(maxUnavailableVal)
+	logrus.Debugf("Provided value for maxUnavailable: %v", maxUnavailableParsed)
+	maxUnavailable, err := k8sutil.GetValueFromIntOrPercent(&maxUnavailableParsed, numHosts, false)
+	if err != nil {
+		return 0, err
```
superseb: Can we add a helping text here, something along the lines of "Unable to parse max_unavailable, should be a number of percentage of nodes: %v"

services/controlplane.go
```diff
@@ -39,6 +46,76 @@ func RunControlPlane(ctx context.Context, controlHosts []*hosts.Host, localConnD
 	return nil
 }
 
+func UpgradeControlPlane(ctx context.Context, kubeClient *kubernetes.Clientset, controlHosts []*hosts.Host, localConnDialerFactory hosts.DialerFactory, prsMap map[string]v3.PrivateRegistry, cpNodePlanMap map[string]v3.RKEConfigNodePlan, updateWorkersOnly bool, alpineImage string, certMap map[string]pki.CertificatePKI, upgradeStrategy *v3.NodeUpgradeStrategy, newHosts map[string]bool) error {
+	if updateWorkersOnly {
```
superseb: Can we add a log line similar to https://github.com/rancher/rke/pull/1800/files#diff-ee57d1a642f3b4ee28fb4929afe13026R55

services/workerplane.go
```diff
@@ -42,6 +51,170 @@ func RunWorkerPlane(ctx context.Context, allHosts []*hosts.Host, localConnDialer
...
+	if maxUnavailable > WorkerThreads {
+		/* upgrading a large number of nodes in parallel leads to a large number of goroutines, which has led to errors regarding too many open sockets
+		Because of this RKE switched to using workerpools. 50 workerthreads has been sufficient to optimize rke up, upgrading at most 50 nodes in parallel.
+		So the user configurable maxUnavailable will be respected only as long as it's less than 50 and capped at 50 */
+		maxUnavailable = WorkerThreads
+		logrus.Info("Setting maxUnavailable to 50, to avoid issues related to upgrading large number of nodes in parallel")
+	}
+
+	maxUnavailable -= len(inactiveHosts)
+	log.Infof(ctx, "First checking and upgrading worker components on nodes with etcd/controlplane roles one at a time")
```
superseb: Can we bring this in line with https://github.com/rancher/rke/pull/1800/files#diff-a1440968342b4da00653d6c74b21f48eR55

services/workerplane.go
```diff
@@ -42,6 +51,170 @@ func RunWorkerPlane(ctx context.Context, allHosts []*hosts.Host, localConnDialer
...
+	if err != nil {
+		logrus.Errorf("Failed to upgrade hosts: %v", strings.Join(multipleRolesHostsFailedToUpgrade, ","))
+		return err
+	}
+
+	log.Infof(ctx, "Now checking and upgrading worker components on nodes with only worker role %v at a time", maxUnavailable)
+	workerOnlyHostsFailedToUpgrade, err := startWorkerPlane(ctx, kubeClient, workerOnlyHosts, localConnDialerFactory, prsMap, workerNodePlanMap, certMap, updateWorkersOnly, alpineImage, maxUnavailable, upgradeStrategy, newHosts)
+	if err != nil {
+		logrus.Errorf("Failed to upgrade hosts: %v", strings.Join(workerOnlyHostsFailedToUpgrade, ","))
+		if len(workerOnlyHostsFailedToUpgrade) == maxUnavailable {
```
superseb: This needs to be `>=` I assume.

services/controlplane.go
mrajashree: There is this line below it `log.Infof(ctx, "[%s] Processing control plane components for upgrade one at a time", ControlRole)`, that should be good?

services/workerplane.go
prachidamle: Maybe should be "... should be a number or percentage of nodes..."

services/workerplane.go
```diff
@@ -42,6 +53,180 @@ func RunWorkerPlane(ctx context.Context, allHosts []*hosts.Host, localConnDialer
...
+	for _, h := range allHosts {
+		_, err := k8s.GetNode(kubeClient, h.HostnameOverride)
+		if err != nil && apierrors.IsNotFound(err) {
+			// this host could have been added to cluster state upon successful controlplane upgrade but isn't a node yet.
+			newHosts[h.HostnameOverride] = true
+		}
+	}
+}
+
+func startWorkerPlane(ctx context.Context, kubeClient *kubernetes.Clientset, allHosts []*hosts.Host, localConnDialerFactory hosts.DialerFactory, prsMap map[string]v3.PrivateRegistry, workerNodePlanMap map[string]v3.RKEConfigNodePlan, certMap map[string]pki.CertificatePKI, updateWorkersOnly bool, alpineImage string,
```
prachidamle: nit: Maybe the method name  could be "startWorkerPlaneUpgrade" to make it clear that this is a part of the upgrade process?

services/controlplane.go
```diff
@@ -39,6 +47,80 @@ func RunControlPlane(ctx context.Context, controlHosts []*hosts.Host, localConnD
...
+		}
+		nodes, err := getNodeListForUpgrade(kubeClient, &sync.Map{}, newHosts, false)
+		if err != nil {
+			return err
+		}
+		var maxUnavailableHit bool
+		for _, node := range nodes {
+			// in case any previously added nodes or till now unprocessed nodes become unreachable during upgrade
+			if !k8s.IsNodeReady(node) {
+				maxUnavailableHit = true
```
prachidamle: Are we saying maxUnavailableHit = true because at any point in time only one control plane node can be down?

services/workerplane.go
```diff
@@ -42,6 +53,180 @@ func RunWorkerPlane(ctx context.Context, allHosts []*hosts.Host, localConnDialer
...
+				}
+				var maxUnavailableHit bool
+				for _, node := range nodes {
+					// in case any previously added nodes or till now unprocessed nodes become unreachable during upgrade
+					if !k8s.IsNodeReady(node) {
+						if len(hostsFailedToUpgrade) == maxUnavailable {
+							maxUnavailableHit = true
+							break
+						}
+						hostsFailed.Store(runHost.HostnameOverride, true)
```
prachidamle: We have not attempted upgrade on the runHost yet, so why are we adding the current runHost to "hostsFailed"?

services/workerplane.go
```diff
@@ -42,6 +53,180 @@ func RunWorkerPlane(ctx context.Context, allHosts []*hosts.Host, localConnDialer
...
+						hostsFailedToUpgrade <- node.Labels[k8s.HostnameLabel]
+						errList = append(errList, fmt.Errorf("host %v not ready", node.Labels[k8s.HostnameLabel]))
+					}
+				}
+				if maxUnavailableHit || len(hostsFailedToUpgrade) == maxUnavailable {
+					break
+				}
+				upgradable, err := isWorkerHostUpgradable(ctx, runHost, workerNodePlanMap[runHost.Address].Processes)
+				if err != nil {
+					errList = append(errList, err)
```
prachidamle: We should add information to the err here to indicate that upgrade on this host failed because of this err while checking `isWorkerHostUpgradable`

services/controlplane.go
mrajashree: yes

services/workerplane.go
mrajashree: The methods called within `isWorkerHostUpgradable` append hostname to the error before returning

cluster/validation.go
```diff
@@ -196,6 +196,39 @@ func ValidateHostCount(c *Cluster) error {
...
+		}
+		workerOnlyHosts++
+	}
+
+	maxUnavailable, err := services.CalculateMaxUnavailable(c.UpgradeStrategy.MaxUnavailable, workerOnlyHosts)
+	if err != nil {
+		return err
+	}
+	if len(inactiveWorkerOnlyHosts) >= maxUnavailable {
+		return fmt.Errorf("cannot proceed with upgrade of worker components since %v (>=maxUnavailable) hosts are inactive", strings.Join(inactiveWorkerOnlyHosts, ","))
```
superseb: Can we change this to

```
return fmt.Errorf("cannot proceed with upgrade of worker components since %d (>=maxUnavailable) hosts are inactive: %v", len(inactiveWorkerOnlyHosts), strings.Join(inactiveWorkerOnlyHosts, ","))
```

This is how this currently looks:

```
FATA[0006] cannot proceed with upgrade of worker components since 18.202.232.104 (>=maxUnavailable) hosts are inactive 
```

superseb:
The outcome of upgrading workers with one node that never gets to `Ready` state:

```
DEBU[0097] [worker] Now checking status of node 18.202.232.104                                                                            
DEBU[0097] [worker] Found node by name 18.202.232.104                                                                                     
DEBU[0102] [worker] Now checking status of node 18.202.232.104                                                                            
DEBU[0102] [worker] Found node by name 18.202.232.104                                                                                     
ERRO[0107] Failed to upgrade hosts: 18.202.232.104,18.202.232.104 with error [host 18.202.232.104 not ready] 
...
INFO[0130] Finished building Kubernetes cluster successfully 
```

- log contains node name twice
- End of the run doesnt log that there was an error with a node

cluster/defaults.go
```diff
@@ -98,6 +103,7 @@ func setDefaultIfEmptyMapValue(configMap map[string]string, key string, value st
 
 func setDefaultIfEmpty(varName *string, defaultValue string) {
 	if len(*varName) == 0 {
+		logrus.Debugf("No value provided for %v, setting it to default value of %v", varName, defaultValue)
```
superseb: This will print the memory address of varName

services/workerplane.go
```diff
@@ -42,6 +53,182 @@ func RunWorkerPlane(ctx context.Context, allHosts []*hosts.Host, localConnDialer
...
+						hostsFailedToUpgrade <- runHost.HostnameOverride
+						hostsFailed.Store(runHost.HostnameOverride, true)
+						break
+					}
+					continue
+				}
+				nodes, err := getNodeListForUpgrade(kubeClient, &hostsFailed, newHosts, true)
+				if err != nil {
+					errList = append(errList, err)
+					hostsFailed.Store(runHost.HostnameOverride, true)
```
prachidamle: Even here I think we should not be adding the runHost to hostsFailed since upgrade is not yet attempted

services/workerplane.go
```diff
@@ -42,6 +53,182 @@ func RunWorkerPlane(ctx context.Context, allHosts []*hosts.Host, localConnDialer
...
+
+func upgradeWorkerHost(ctx context.Context, kubeClient *kubernetes.Clientset, runHost *hosts.Host, drainFlag bool, drainHelper drain.Helper,
+	localConnDialerFactory hosts.DialerFactory, prsMap map[string]v3.PrivateRegistry, workerNodePlanMap map[string]v3.RKEConfigNodePlan, certMap map[string]pki.CertificatePKI, updateWorkersOnly bool,
+	alpineImage string) error {
+	if err := checkNodeReady(kubeClient, runHost, WorkerRole); err != nil {
+		return err
+	}
+	// cordon and drain
+	if err := cordonAndDrainNode(kubeClient, runHost, drainFlag, drainHelper, WorkerRole); err != nil {
+		return err
```
prachidamle: To this error returned here by cordonAndDrainNode, can we add a prefix saying "upgradeWorkerHost failed to cordonAndDrainNode with error:"
It would be good to have such prefixes in this method that can help debug at what step did the upgrade fail

services/workerplane.go
```diff
@@ -42,6 +53,182 @@ func RunWorkerPlane(ctx context.Context, allHosts []*hosts.Host, localConnDialer
...
+	if err := doDeployWorkerPlaneHost(ctx, runHost, localConnDialerFactory, prsMap, workerNodePlanMap[runHost.Address].Processes, certMap, updateWorkersOnly, alpineImage); err != nil {
+		return err
+	}
+	// consider upgrade done when kubeclient lists node as ready
+	if err := checkNodeReady(kubeClient, runHost, WorkerRole); err != nil {
+		return err
+	}
+	// uncordon node
+	if err := k8s.CordonUncordon(kubeClient, runHost.HostnameOverride, false); err != nil {
+		return err
```
prachidamle: Same comment as above about adding a prefix to the err to specify where upgrade failed for this worker host

services/workerplane.go
prachidamle: If the methods called already prefix enough information, thats fine. Otherwise we can add it here.

services/workerplane.go
mrajashree: CordonUncordon appends the hostname to error

PR #1824, opened 2019-12-12
Fix always show certificates are about to expire issue

XianglongLuo:
Problem: 

The ETCD certs of removed hosts will not be removed or rotated. 
API/UI will always show there are certs expired in this cluster even all needed certs are rotated.
In the screenshot, the highlight etcd host has been removed from the cluster.

So users will always see the certificates expire warning in Rancher

image

image

Solution: 

Delete the unused ETCD certs when rotating certs

Related Rancher PR:

https://github.com/rancher/rancher/pull/24548

Issue: 

https://github.com/rancher/rancher/issues/24333

### DIFF
--- a/pki/services.go
+++ b/pki/services.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"reflect"
 	"sort"
+	"strings"
 
 	"github.com/rancher/rke/hosts"
 	"github.com/rancher/rke/log"
@@ -370,7 +371,6 @@ func GenerateEtcdCertificates(ctx context.Context, certs map[string]CertificateP
 		ips = append(ips, ip.String())
 	}
 	sort.Strings(ips)
-
 	for _, host := range etcdHosts {
 		etcdName := GetCrtNameForHost(host, EtcdCertName)
 		if _, ok := certs[etcdName]; ok && certs[etcdName].CertificatePEM != "" && !rotate {
@@ -403,6 +403,8 @@ func GenerateEtcdCertificates(ctx context.Context, certs map[string]CertificateP
 		}
 		certs[etcdName] = ToCertObject(etcdName, "", "", etcdCrt, etcdKey, nil)
 	}
+	log.Debugf(ctx, "Checking and deleting unused etcd certificates, current etcd nodes are: %v", etcdHosts)
+	deleteUnusedCerts(ctx, certs, EtcdCertName, etcdHosts)
 	return nil
 }
 
@@ -516,6 +518,8 @@ func GenerateKubeletCertificate(ctx context.Context, certs map[string]Certificat
 		}
 		certs[kubeletName] = ToCertObject(kubeletName, "", "", kubeletCrt, kubeletKey, nil)
 	}
+	log.Debugf(ctx, "Checking and deleting unused kubelet certificates, current nodes are : %v", allHosts)
+	deleteUnusedCerts(ctx, certs, KubeletCertName, allHosts)
 	return nil
 }
 
@@ -555,6 +559,15 @@ func GenerateRKEServicesCerts(ctx context.Context, certs map[string]CertificateP
 	}
 	if IsKubeletGenerateServingCertificateEnabledinConfig(&rkeConfig) {
 		RKECerts = append(RKECerts, GenerateKubeletCertificate)
+	} else {
+		//Clean up kubelet certs when GenerateServingCertificate is disabled
+		log.Infof(ctx, "[certificates] GenerateServingCertificate is disabled, checking if there are unused kubelet certificates")
+		for k := range certs {
+			if strings.HasPrefix(k, KubeletCertName) {
+				log.Infof(ctx, "[certificates] Deleting unused kubelet certificate: %s", k)
+				delete(certs, k)
+			}
+		}
 	}
 	for _, gen := range RKECerts {
 		if err := gen(ctx, certs, rkeConfig, configPath, configDir, rotate); err != nil {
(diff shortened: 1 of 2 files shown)


gitlawr:
Can you explain why only the ETCD certs are affected? What if the generate_serving_certificate 
 is enabled for kubelet?(https://rancher.com/docs/rke/latest/en/config-options/services/#kubelet)

XianglongLuo:
fixed @gitlawr

gitlawr:
The issue/pr description is talking about Rancher APIs. Can you justify why the fix needs to be on the RKE side?

pki/services.go
```diff
@@ -555,6 +577,13 @@ func GenerateRKEServicesCerts(ctx context.Context, certs map[string]CertificateP
 	}
 	if IsKubeletGenerateServingCertificateEnabledinConfig(&rkeConfig) {
 		RKECerts = append(RKECerts, GenerateKubeletCertificate)
+	} else {
+		//Clean up kubelet certs which generated when GenerateServingCertificate enabled
+		for k := range certs {
```
gitlawr: Why do we remove all kubelet certs here

pki/services.go
```diff
@@ -492,10 +504,17 @@ func GenerateKubeletCertificate(ctx context.Context, certs map[string]Certificat
 		return fmt.Errorf("CA Certificate or Key is empty")
 	}
 	log.Infof(ctx, "[certificates] Generating Kubernetes Kubelet certificates")
+	unusedKubeletCerts := make(map[string]bool)
+	for k := range certs {
```
gitlawr: These changes look identical to the etcd ones. Can we extract a common function to be invoked in these two places

pki/services.go
XianglongLuo: When we rke up, the generate serving certificate is disabled in default,and there haven’t any kubelet cert in cluster.rkestate:
￼
image

And if we   enable  generate serving certificate and rke up again,it will generate kubelet certs,we can see them in cluster.rkestate:
image

￼
But if we disable generate serving certificate  and rke up again, the kubelet certs will be unused and left in cluster.rkestate:
image

￼
So we need to delete all unused kubelet certs when  generate serving certificate is disabled ~

pki/services.go
XianglongLuo: > The issue/pr description is talking about Rancher APIs. Can you justify why the fix needs to be on the RKE side?
  justified in:https://github.com/rancher/rancher/issues/24333

XianglongLuo:
> The issue/pr description is talking about Rancher APIs. Can you justify why the fix needs to be on the RKE side?


justified in:https://github.com/rancher/rancher/issues/24333

pki/services.go
```diff
@@ -588,3 +597,19 @@ func GenerateRKEServicesCSRs(ctx context.Context, certs map[string]CertificatePK
 	}
 	return nil
 }
+
+func DeleteUnusedCerts(certs map[string]CertificatePKI, certName string, hosts []*hosts.Host) {
```
gitlawr: Use lower case given that we don't need to export this function

pki/services.go
```diff
@@ -555,6 +557,13 @@ func GenerateRKEServicesCerts(ctx context.Context, certs map[string]CertificateP
 	}
 	if IsKubeletGenerateServingCertificateEnabledinConfig(&rkeConfig) {
 		RKECerts = append(RKECerts, GenerateKubeletCertificate)
+	} else {
+		//Clean up kubelet certs which generated when GenerateServingCertificate enabled
```
gitlawr: Fix grammar

XianglongLuo:
updated @gitlawr

pki/services.go
```diff
@@ -588,3 +597,19 @@ func GenerateRKEServicesCSRs(ctx context.Context, certs map[string]CertificatePK
 	}
 	return nil
 }
+
+func deleteUnusedCerts(certs map[string]CertificatePKI, certName string, hosts []*hosts.Host) {
```
superseb: Can we add some logging here to indicate what we are doing

superseb:
Was it considered to include https://github.com/rancher/rancher/issues/23869 and https://github.com/rancher/rancher/issues/23870? Else I just linked them so when they are being worked on, there is a reference.

loganhz:
> Was it considered to include [rancher/rancher#23869](https://github.com/rancher/rancher/issues/23869) and [rancher/rancher#23870](https://github.com/rancher/rancher/issues/23870)? Else I just linked them so when they are being worked on, there is a reference.

Hi @superseb Yes, considered.

pki/services.go
loganhz: Done

loganhz:
@superseb we removed the certs in secret of cattle-system, but the certs still exist on nodes (`/etc/kubernetes/ssl`)
Do you mean we should remove them on node as well?

superseb:
Yep, the API part is the most important as its blocking functionality. Just wanted to make sure if it was explored or not as we either close the other issue for removing certificates or make sure it was intentionally not included as it will come up and then we can refer to the open issue for this.

loganhz:
@superseb Thanks! Please help to re-review it.

loganhz:
updated. Thanks

pki/services.go
```diff
@@ -516,6 +518,8 @@ func GenerateKubeletCertificate(ctx context.Context, certs map[string]Certificat
 		}
 		certs[kubeletName] = ToCertObject(kubeletName, "", "", kubeletCrt, kubeletKey, nil)
 	}
+	log.Debugf(ctx, "Checking and deleting unused kubelet certificates, current worker nodes are : %v", allHosts)
```
superseb: Can we make this equal to Rancher's:


```
logrus.Debugf("Checking and deleting unused kubelet certificates, current nodes are: %v", hosts.NodesToHosts(cluster.Status.AppliedSpec.RancherKubernetesEngineConfig.Nodes, ""))
```

superseb:
One logging note, then I can merge

XianglongLuo:
Updated.

PR #1829, opened 2019-12-12
Vendor update

brendarearden:


brendarearden:
Changes are getting vendored in under different PR

Issue #1823, opened 2019-12-11
/lib/modules bind mount for kube-proxy incorrectly points to /opt/rke/lib/modules on RancherOS


bensallen:
**RKE version:**
Testing with v1.0.0, but relevant code doesn't appear changed for v1.1.0 or master.

**Docker version: (`docker version`,`docker info` preferred)**
```
$ docker version
Client: Docker Engine - Community
 Version:           19.03.4
 API version:       1.40
 Go version:        go1.12.10
 Git commit:        9013bf583a
 Built:             Fri Oct 18 15:49:05 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.4
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.12.10
  Git commit:       9013bf583a
  Built:            Fri Oct 18 15:55:51 2019
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.2.10
  GitCommit:        b34a5c8af56e510852c35414db4c1f4fa6172339
 runc:
  Version:          1.0.0-rc8+dev
  GitCommit:        3e425f80a8c931f88e6d94a8c831b9d5aa481657
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)**

```
$ cat /etc/os-release
NAME="RancherOS"
VERSION=v1.5.4
ID=rancheros
ID_LIKE=
VERSION_ID=v1.5.4
PRETTY_NAME="RancherOS v1.5.4"
HOME_URL="http://rancher.com/rancher-os/"
SUPPORT_URL="https://forums.rancher.com/c/rancher-os"
BUG_REPORT_URL="https://github.com/rancher/os/issues"
BUILD_ID=
```

```
$ uname -r
4.14.138-rancher
```

**Type/provider of hosts: (VirtualBox/Bare-metal/AWS/GCE/DO)**
Bare-metal

**cluster.yml file:**
```
...
services:
    kubeproxy:
      extra_args:
        proxy-mode: "ipvs"
...
```

**Steps to Reproduce:**

Provision a cluster on RancherOS with the cluster.yml enabling ipvs proxy-mode via extra_args.

**Results:**

[...]

PR #1837, opened 2019-12-17
[v1.0] Remove the prefixPath from the /lib/modules bind mount for kube-proxy…

deniseschannon:
… IPVS support


deniseschannon:
https://github.com/rancher/rke/issues/1823

PR #1856, opened 2019-12-23
[release/v1.0] Change MTU from string to int

superseb:
https://github.com/rancher/rancher/issues/24096

- vendor update of types and kontainer-driver-metadata
- change comparing string because its an int now


superseb:
@brendarearden @luthermonson Think this includes changes from you, let me know if this is breaking.

luthermonson:
mine are for a rancher controller feature, doesn't affect me any if this gets into RKE

PR #1872, opened 2020-01-09
Fix log output of full-cluster-state configmap name

Lucaber:
This PR corrects the log output during `rke up`.
The output still contains the legacy configmap name `cluster-state`. The new cluster state is saved in `full-cluster-state`.


superseb:
Thanks @Lucaber

PR #1875, opened 2020-01-12
Draining option added to the rke ujp

WoodProgrammer:
Hello Rancher/RKE team,

I'm using RKE to manage the clusters. Rke up command is updating cluster fastly but there is something wrong on  upgrade and update progress I think .

RKE do not drain  the k8s nodes according to that this can cause down time .

Regarding to that I have added pod eviction option for rke . I want to test and improve these PR maybe we can improve all together. 

I will expect your feedbacks :D 

Thanks

### DIFF
--- a/hosts/hosts.go
+++ b/hosts/hosts.go
@@ -161,9 +161,18 @@ func DeleteNode(ctx context.Context, toDeleteHost *Host, kubeClient *kubernetes.
 		return err
 
 	}
+
+
 	if err := k8s.CordonUncordon(kubeClient, toDeleteHost.HostnameOverride, true); err != nil {
 		return err
 	}
+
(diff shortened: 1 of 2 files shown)


mrajashree:
Hi @WoodProgrammer thank you for the PR! However I think we can use the drain pkg from kubernetes for this https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kubectl/pkg/drain/default.go#L33

[This](https://github.com/rancher/rke/pull/1800/files) PR will add the drain option while upgrading cluster

WoodProgrammer:
thanks for your advice, I just want to implement eviction progress basically on my nodes rather than directly drain nodes :)) I think with `RunNodeDrain ` function , will be code block more clean maybe. But your PR is looks  more comprehensive than this one .

superseb:
Let us know if the newly introduced functionality does not cover your use case.

PR #1884, opened 2020-01-16
Drop the log level to Debug since rancher can call this frequently

prachidamle:
When the parameter `generate_serving_certificate: true` is set on the cluster, rancher will regenerate kubelet certs every time worker agents connect back every ~2 mins. This is flooding the rancher logs in this case. Lowering the level to debug.

https://github.com/rancher/rancher/issues/24942

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -13,6 +13,7 @@ const (
 )
 
 type logger interface {
+	Debugf(msg string, args ...interface{})
 	Infof(msg string, args ...interface{})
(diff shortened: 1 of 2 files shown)


prachidamle:
https://github.com/rancher/rancher/issues/24942

prachidamle:
@superseb yes exactly my thought process.
I was thinking of logging it at INFO if rotate = true in this method, and debug otherwise.
But then changing to debug will always have this available, so dint go for the over kill.

PR #1888, opened 2020-01-20
Determine etcd s3 snapshots by s3 config presence

superseb:
https://github.com/rancher/rke/issues/1538

### DIFF
--- a/cluster/etcd.go
+++ b/cluster/etcd.go
@@ -3,7 +3,6 @@ package cluster
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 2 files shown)


superseb:
@galal-hussein Can this logic be simplified like this or what usecase is missed in this code?

galal-hussein:
> @galal-hussein Can this logic be simplified like this or what usecase is missed in this code?

I can't think of any missing usecase in that code.. I think its good fix, LGTM

PR #1895, opened 2020-01-24
cluster plan for worker nodes

kinarashah:
Need for cluster level plan: 
- on rancher side, we can compute `hash(clusterPlan.Processes)` and can use it to decide if worker nodes need to upgrade or not 
- have commented out changes from nodePlan just for easy comparison, will remove later 

Caveat: 
- clusterPlan process is kind of a mirror to node plan so on every code change in process, we'll also have to keep clusterProcess in sync, could introduce bugs
- also had to introduce a new `RKEProcess` and couldn't use the existing one, because we need map for a few fields (vs array) to easily check if there's a change irrespective of order. So far it hasn't been an issue, because we do container level checks with command args.


cluster/plan.go
```diff
@@ -416,6 +475,184 @@ func (c *Cluster) BuildKubeControllerProcess(host *hosts.Host, prefixPath string
...
+		CommandArgs["healthz-bind-address"] = "0.0.0.0"
+	}
+
+	//if host.IsControl && !host.IsWorker {
+	//	CommandArgs["register-with-taints"] = unschedulableControlTaint
+	//}
+	//if host.Address != host.InternalAddress {
+	//	CommandArgs["node-ip"] = host.InternalAddress
+	//}
+	if len(c.CloudProvider.Name) > 0 {
```
mrajashree: maybe we can club this and line 542 under a single if?

cluster/plan.go
```diff
@@ -416,6 +475,184 @@ func (c *Cluster) BuildKubeControllerProcess(host *hosts.Host, prefixPath string
...
+	addMap(bindsMap, "/var/log/containers:/var/log/containers:z")
+	addMap(bindsMap, "/var/log/pods:/var/log/pods:z")
+	addMap(bindsMap, "/usr:/host/usr:ro")
+	addMap(bindsMap, "/etc:/host/etc:ro")
+
+	// Special case to simplify using flex volumes
+	//if path.Join(prefixPath, "/var/lib/kubelet") != "/var/lib/kubelet" {
+	//	Binds = append(Binds, "/var/lib/kubelet/volumeplugins:/var/lib/kubelet/volumeplugins:shared,z")
+	//}
+	if isWindows {
```
mrajashree: this if can continue for updating envMap too on line 618?

cluster/plan.go
```diff
@@ -598,6 +835,119 @@ func (c *Cluster) BuildKubeletProcess(host *hosts.Host, prefixPath string, svcOp
 	}
 }
 
+func (c *Cluster) BuildClusterKubeProxyProcess(osType string, svcOptionData map[string]*v3.KubernetesServicesOptions) v3.RKEProcess {
+	isWindows := osType == "windows"
+
+	envMap := map[string]sets.Empty{}
+	bindsMap := map[string]sets.Empty{}
+
+	for _, env := range c.Services.Kubeproxy.ExtraEnv {
```
mrajashree: maybe we can extract this logic of populating envMap and bindsMap for all processes?

cluster/plan.go
```diff
@@ -598,6 +835,119 @@ func (c *Cluster) BuildKubeletProcess(host *hosts.Host, prefixPath string, svcOp
...
+	VolumesFrom := []string{
+		services.SidekickContainerName,
+	}
+	//TODO: we should reevaluate if any of the bind mounts here should be using read-only mode
+
+	addMap(bindsMap, fmt.Sprintf("%s:/etc/kubernetes:z", "/etc/kubernetes"))
+	addMap(bindsMap, "/run:/run")
+	addMap(bindsMap, fmt.Sprintf("%s:/lib/modules:z,ro", "/lib/modules"))
+
+	if isWindows { // compatible with Windows
```
mrajashree: continue the same if till line 913?

cluster/plan.go
```diff
@@ -167,6 +195,37 @@ func osLimitationFilter(osType string, processes map[string]v3.Process) map[stri
...
+
+	// windows limitations
+	for name, process := range processes {
+		// doesn't support host network on windows
+		if process.NetworkMode == "host" {
+			process.NetworkMode = ""
+		}
+
+		// doesn't support PID on windows
+		if process.PidMode != "" {
```
mrajashree: nit: you can directly set `process.PidMode` to `""` without checking and same for `process.Privileged`

superseb:
As discussed, initially the design was to reuse the data for the processes but creating a global function for cluster/node became unreadable. Still we have duplicated params in this code, so we decided to look into ways to use one source of params and post process them, and to add tests so we cant make a mistake in a upcoming PR where params are added in one place and not the other.

We also discussed that if the postprocessing/global function becomes unreadable or so complex that it won't really help us, we will keep the duplicated data but we will still add the test to guard us from making a mistake.

kinarashah:
@mrajashree @supseb
So, 
1) I'm going to leave these functions as is, mirror of existing node functions for now. 
2) After preview, work on adding tests so we either catch any out of sync errors between the two process builds or just combine them into one. 

Could I get a LGTM for now if just the logic for these cluster processes look ok to you?

PR #1910, opened 2020-02-12
Stop vendoring KDM and use data.json

StrongMonkey:
~Depends on rancher/types#1088.~ dependency updated.


StrongMonkey:
@superseb Updated and rebased.

superseb:
@StrongMonkey integration tests are not run because it cant list available k8s versions (and doesnt error out on it apparently):

```
Running validation
--
303 | Running: go vet
304 | Running: golint
305 | Running: go fmt
306 | INFO - Running ./integration
307 |  
308 | ---- TEST RESULTS ----
309 |  
310 | ---- Clean Up RKE ----
```

StrongMonkey:
@superseb CI was finally fixed and one of reason is that 1.16.7 images is not being pushed due to CI failure. Now all 5 versions should be tested and passed.

StrongMonkey:
Also this PR adds ability to specify an external `data.json` for rke to consume. This is exposed through `METADATA_URL`. For example:

```bash
METADATA_URL=${local_or_remote_file_url} rke up
```

We need to add docs for this new feature.

metadata/metadata.go
```diff
@@ -23,31 +31,66 @@ var (
...
+	}
+	initK8sRKESystemImages(data)
+	initAddonTemplates(data)
+	initServiceOptions(data)
+	initDockerOptions(data)
 	return nil
 }
 
+// this method loads metadata from three place, if METADATA_URL is provided then load data from specified location. Otherwise load data from bindata.
+func loadData() (kdm.Data, error) {
```
superseb: Can we add some logging and debug logging in this function to know:
- What source of data is being used
- When it is trying to retrieve data from URL or file
- That it is trying to retrieve something

metadata/metadata.go
```diff
@@ -23,31 +31,66 @@ var (
...
+	var b []byte
+	var err error
+	u := os.Getenv(metadataURLEnv)
+	if u != "" {
+		b, err = readFile(u)
+		if err != nil {
+			return kdm.Data{}, err
```
superseb: Can we describe this error better, its possible this error does not describe what is actually failing (just a HTTP response but not describing what it is trying to do)

metadata/metadata.go
```diff
@@ -23,31 +31,66 @@ var (
...
+		}
+	}
+	return kdm.FromData(b)
+}
+
+func readFile(file string) ([]byte, error) {
+	if strings.HasPrefix(file, "http") {
+		resp, err := http.Get(file)
```
superseb: Does this need timeouts specified?

README.md
```diff
@@ -23,6 +23,14 @@ Please use [High Availability (HA) Install](https://rancher.com/docs/rancher/v2.
 
 RKE can be built using the `make` command, and will use the scripts in the `scripts` directory as subcommands. The default subcommand is `ci` and will use `scripts/ci`. Cross compiling can be enabled by setting the environment variable `CROSS=1`. The compiled binaries can be found in the `build/bin` directory. Dependencies are managed by Go modules and can be found in [go.mod](https://github.com/rancher/rke/blob/master/go.mod).
 
+RKE now fetches `data.json` from https://github.com/rancher/kontainer-driver-metadata. To fetch data.json and compile it in rke, run
```
superseb: Can we describe both build and runtime parameters for using `data.json`?

metadata/metadata.go
StrongMonkey: Updated

metadata/metadata.go
StrongMonkey: Update with detailed error message for func `loadData`.

metadata/metadata.go
StrongMonkey: Added with 30 seconds timeout client.

README.md
StrongMonkey: Added. Also change both parameters to use the same env variable.

StrongMonkey:
https://github.com/rancher/rancher/issues/25162

Issue #1909, opened 2020-02-12
When one worker node(out of 5) is Powered OFF and then upgrade is initiated, upgrade fails. [FATA[0104] [workerPlane] Failed to upgrade Worker Plane: [host NW2 not ready] ]


soumyalj:
RKE version: v1.1.0-rc5

Docker version: (docker version,docker info preferred) 19.03

Operating system and kernel: (cat /etc/os-release, uname -r preferred) Ubuntu

**Type/provider of hosts: (VirtualBox/Bare-metal/AWS/GCE/DO)**EC2

Steps to Reproduce:

1. Create a RKE Cluster with 3 control planes, 1 etcd node and 5 worker nodes(NW1 to NW5)
2. Power OFF one worker node NW2
3. Upgrade to a new K8s version v1.17.2-rancher1-1 using the below upgrade strategy

```
kubernetes_version: v1.17.2-rancher1-1

upgrade_strategy:
  max_unavailable: 2
  drain: true
  node_drain_input:
    force: false
    ignore_daemonsets: true
    delete_local_data: false
    grace_period: -1
    timeout: 120

```

Upgrade exits though with the below error.  There are 4 nodes which are still Ready. 
```
INFO[0104] First checking and processing worker components for upgrades on nodes with etcd/controlplane roles one at a time 
INFO[0104] [workerplane] Processing host abcd (control plane host )
ERRO[0104] Failed to upgrade hosts: NW2(worker node) with error [host NW2 not ready] 
FATA[0104] [workerPlane] Failed to upgrade Worker Plane: [host NW2 not ready] 

```
Also, in the above case, it was observed that etcd plane was upgraded, but control plane nodes did not get upgraded. Only the state was saved. While processing the worker plane components on the control/etcd planes, the upgrade exits. RKE logs below
```
WARN[0000] This is not an officially supported version (v1.1.0-rc5) of RKE. Please download the latest official release at https://github.com/rancher/rke/releases/latest 
INFO[0000] Running RKE version: v1.1.0-rc5              
INFO[0000] Initiating Kubernetes cluster                
INFO[0000] [certificates] Generating admin certificates and kubeconfig 
INFO[0000] Successfully Deployed state file at [./clusterzeroworker.rkestate] 
INFO[0000] Building Kubernetes cluster                  
INFO[0000] [dialer] Setup tunnel for host [efgh] 
INFO[0000] [dialer] Setup tunnel for host [ijkl] 
[...]

Issue #1908, opened 2020-02-11
When one worker node is Powered OFF and then upgrade is initiated, upgrade exits (max_unavailable is provided as a percentage)


soumyalj:
RKE version: v1.1.0-rc5

Docker version: (docker version,docker info preferred) 19.03

Operating system and kernel: (cat /etc/os-release, uname -r preferred) Ubuntu

**Type/provider of hosts: (VirtualBox/Bare-metal/AWS/GCE/DO)**EC2

Steps to Reproduce:

1. Create a RKE Cluster with 1 control plane, 1 etcd node and 5 worker nodes(NW1 to NW5) using K8s version "v1.16.6-rancher1-1"
2. Power OFF one worker node NW2
3. Upgrade to a new K8s version v1.17.2-rancher1-1 using the below upgrade strategy

```
kubernetes_version: v1.17.2-rancher1-1

upgrade_strategy:
  max_unavailable: 40%
  drain: true
  node_drain_input:
    force: false
    ignore_daemonsets: true
    delete_local_data: false
    grace_period: -1
    timeout: 120

services:
   kubelet:
     extra_args:
       max-pods: 170

```
Upgrade fails with the below error:
```
DEBU[0018] FIXME: Got an status-code for which error does not match any expected type!!!: -1  module=api status_code=-1
WARN[0018] Failed to set up SSH tunneling for host [abcd]: Can't retrieve Docker Info: error during connect: Get http://%2Fvar%2Frun%2Fdocker.sock/v1.24/info: Unable to access node with address [abcd:22] using SSH. Please check if the node is up and is accepting SSH connections or check network policies and firewall rules. Error: dial tcp abcd:22: connect: operation timed out 
WARN[0018] Removing host [abcd] from node lists 
DEBU[0018] Provided value for maxUnavailable: {1 0 40%} 
INFO[0018] 1 worker nodes can be unavailable at a time  
FATA[0018] cannot proceed with upgrade of worker components since 1 (>=maxUnavailable) hosts are inactive; found inactive hosts: abcd 

```

In the above case with with one worker node OFF, 4 worker nodes are still ON, upgrade should still continue and complete for the other nodes. But upgrade exits.

soumyalj:
The same issue is observed when one node is powered OFF and upgrade is initiated, but only kubelet extra_args is changed and K8s version remains the same .
```
upgrade_strategy:
  max_unavailable: 40%
  drain: true
  node_drain_input:
    force: false
    ignore_daemonsets: true
[...]

Issue #1907, opened 2020-02-08
Powered OFF worker node should not be considered during upgrade


soumyalj:
**RKE version:** v1.1.0-rc5

**Docker version: (`docker version`,`docker info` preferred)** 19.03

**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)** Ubuntu

**Type/provider of hosts: (VirtualBox/Bare-metal/AWS/GCE/DO)**EC2

**cluster.yml file:**

**Steps to Reproduce:**
1. Create a RKE Cluster with 3 control planes, 1 etcd node and 5 worker nodes(NW1 to NW5)
```
ssh_key_path: /Users/soumya/.ssh/jenkins-rke-validation.pem
auth:
  strategy: x509
nodes:
  - address: abc
    user: ubuntu
    role: [etcd]
  - address: def
    user: ubuntu
    role: [controlplane]
  - address: ghk
    user: ubuntu
    role: [controlplane]
  - address: ijk
    user: ubuntu
    role: [controlplane]
  - address: nlp
    user: ubuntu
    role: [worker]
  - address: abcd
    user: ubuntu
    role: [worker]
  - address: efgh
    user: ubuntu
    role: [worker]
  - address: sdfds
    user: ubuntu
    role: [worker]
  - address: dsfsdfs
    user: ubuntu
    role: [worker]
#kubernetes_version: v1.17.2-rancher1-1
kubernetes_version: v1.16.6-rancher1-1
```
2. Create pods and daemonsets in the cluster
3. Power off one worker node ex: NW5
4. Run RKE up again by changing the kubelet extra args
```
ssh_key_path: /Users/soumya/.ssh/jenkins-rke-validation.pem
auth:
  strategy: x509
nodes:
  - address: abc
    user: ubuntu
    role: [etcd]
  - address: def
    user: ubuntu
    role: [controlplane]
  - address: ghk
    user: ubuntu
    role: [controlplane]
  - address: ijk
    user: ubuntu
    role: [controlplane]
  - address: nlp
    user: ubuntu
    role: [worker]
  - address: abcd
    user: ubuntu
    role: [worker]
  - address: efgh
    user: ubuntu
    role: [worker]
  - address: sdfds
    user: ubuntu
    role: [worker]
  - address: dsfsdfs
    user: ubuntu
    role: [worker]

#kubernetes_version: v1.17.2-rancher1-1
kubernetes_version: v1.16.6-rancher1-1

upgrade_strategy:
[...]

Issue #1906, opened 2020-02-08
Zero Down: Add label to ignore nodes


maggieliu:
RKE should allow users to label certain nodes so that they're not included in the upgrade process

mrajashree:
Available with v1.1.0-rc7

mrajashree:
RKE up should accept label on v1.Node so that it ignores that node during upgrade.
The nodes which should be left out of upgrade should have the following label:
```
rke.cattle.io/ignore-during-upgrade=true
```
Users can label the v1.Node using following command:
```
k label node  nodename "rke.cattle.io/ignore-during-upgrade=true"
```

mrajashree:
The updated label we'll use is `user.cattle.io/upgrade-policy=prevent`

mrajashree:
available with rc10

soumyalj:
Tested with rke version: v1.1.0-rc11. Feature to add  a label to ignore nodes is functional. 
Tested the below scenarios:
**Use case 1:**
**Setup:
3 control nodes(C1, C2, C3), 1 etcd, 2 workers             
max_Unavailable_controlplane = 1**  
1. Create a rke cluster using  “./rke up --config  cluster.yml”  command. Specify k8s version as v1.15.5 for example
2. Power off C3
3. Update the cluster by specifying a newer k8s version in the cluster.yml file . Cluster update fails as expected.
4. Label control node C3 with "user.cattle.io/upgrade-policy=prevent"
5. Run command “./rke up --config cluster.yml”  with new K8s version 
After labeling the control node C3, cluster upgrade proceeds. The labeled node will be ignored during upgrade. Cluster should upgraded successfully with new K8s version

**Use case 2:**
**Setup:
3 control nodes(C1, C2, C3), 1 etcd, 4 workers             
max_Unavailable_worker = 1** 
1. Create a rke cluster using  “./rke up --config  cluster.yml”  command. Specify k8s version as v1.15.5 for example
2. Power off W3
3. Update the cluster by specifying a newer k8s version in the cluster.yml file . Cluster update fails as expected.
4. Label control node W3 with "user.cattle.io/upgrade-policy=prevent"
5. Run command “./rke up --config cluster.yml”  with new K8s version 
After labeling the worker node C3, cluster upgrade proceeds. The labeled node will be ignored during upgrade. Cluster gets upgraded successfully with new K8s version

**Use case 3:**
**Setup:
3 control nodes(C1, C2, C3), 1 etcd, 4 workers             
[...]

PR #1912, opened 2020-02-14
Accept label to ignore nodes during upgrade

mrajashree:
RKE does a cluster scan to find the unreachable hosts, and if that number
is same as or exceeds maxUnavailable, upgrade won't proceed.
This commit introduces a label users can provide for their nodes so they
don't get counted as unavailable and are excluded from upgrade.
This commit also includes a couple of bug fixes:
1. When maxUnavailable was provided as a percentage, I was including only the active hosts in the calculation, whereas inactive hosts should be included too. So if from 5 worker hosts 1 is inactive and maxUnavailable is 40%, I was calculating maxUnavailable number as 40% of 4, instead of 40% of 5. Changed this to include all hosts  
2. When upgrading any node, I first check all nodes in the cluster to see if any previously upgraded hosts became inactive, I was listing them using kubectl get nodes. This returned the inactive hosts also but in NotReady state. We don’t need to check the status of Inactive hosts since they’re not considered part of cluster. And we already determine before upgrade starts if number of inactive hosts >= maxUnavailable  
3. The check mentioned above returns all nodes. So if I’m upgrading controlplane nodes, it will also return nodes with only worker role. And we don’t need to check their status when controlplane nodes are getting upgraded. So now I’m using only those nodes from kubectl get nodes output that belong to the current set of nodes being upgraded  
4. Upgrading controlplane and worker components in one go for control nodes

Issue tracking labels enhancement: https://github.com/rancher/rke/issues/1906
Bugs fixed: 
https://github.com/rancher/rke/issues/1909
https://github.com/rancher/rke/issues/1907
https://github.com/rancher/rke/issues/1908

https://github.com/rancher/rancher/issues/25480
Also adds fix to cordon/drain control nodes only once during upgrade by combining upgrade of controlplane and worker components

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -66,6 +66,8 @@ type Cluster struct {
 	WorkerHosts                      []*hosts.Host
 	EncryptionConfig                 encryptionConfig
 	NewHosts                         map[string]bool
+	MaxUnavailableForWorkerNodes     int
+	HostsLabeledToIgnoreUpgrade      map[string]bool
 }
 
 type encryptionConfig struct {
@@ -139,20 +141,32 @@ func (c *Cluster) DeployControlPlane(ctx context.Context, svcOptionData map[stri
 		}
 		return nil
 	}
-	if err := services.UpgradeControlPlane(ctx, kubeClient, c.ControlPlaneHosts,
+	inactiveHosts := make(map[string]bool)
+	var controlPlaneHosts []*hosts.Host
+	for _, host := range c.InactiveHosts {
+		if !c.HostsLabeledToIgnoreUpgrade[host.Address] {
+			inactiveHosts[host.HostnameOverride] = true
+		}
+	}
+	for _, host := range c.ControlPlaneHosts {
+		if !c.HostsLabeledToIgnoreUpgrade[host.Address] {
+			controlPlaneHosts = append(controlPlaneHosts, host)
+		}
+	}
+	if err := services.UpgradeControlPlaneNodes(ctx, kubeClient, controlPlaneHosts,
 		c.LocalConnDialerFactory,
 		c.PrivateRegistriesMap,
 		cpNodePlanMap,
 		c.UpdateWorkersOnly,
 		c.SystemImages.Alpine,
-		c.Certificates, c.UpgradeStrategy, c.NewHosts); err != nil {
+		c.Certificates, c.UpgradeStrategy, c.NewHosts, inactiveHosts); err != nil {
 		return fmt.Errorf("[controlPlane] Failed to upgrade Control Plane: %v", err)
 	}
 	return nil
 }
 
 func (c *Cluster) DeployWorkerPlane(ctx context.Context, svcOptionData map[string]*v3.KubernetesServicesOptions, reconcileCluster bool) (string, error) {
-	var workerOnlyHosts, multipleRolesHosts []*hosts.Host
+	var workerOnlyHosts, etcdAndWorkerHosts []*hosts.Host
 	kubeClient, err := k8s.NewClient(c.LocalKubeConfigPath, c.K8sWrapTransport)
 	if err != nil {
 		return "", fmt.Errorf("failed to initialize new kubernetes client: %v", err)
@@ -161,12 +175,18 @@ func (c *Cluster) DeployWorkerPlane(ctx context.Context, svcOptionData map[strin
 	workerNodePlanMap := make(map[string]v3.RKEConfigNodePlan)
 	// Build cp node plan map
 	allHosts := hosts.GetUniqueHostList(c.EtcdHosts, c.ControlPlaneHosts, c.WorkerHosts)
-	for _, workerHost := range allHosts {
-		workerNodePlanMap[workerHost.Address] = BuildRKEConfigNodePlan(ctx, c, workerHost, workerHost.DockerInfo, svcOptionData)
-		if !workerHost.IsControl && !workerHost.IsEtcd {
-			workerOnlyHosts = append(workerOnlyHosts, workerHost)
+	for _, host := range allHosts {
+		workerNodePlanMap[host.Address] = BuildRKEConfigNodePlan(ctx, c, host, host.DockerInfo, svcOptionData)
+		if host.IsControl || c.HostsLabeledToIgnoreUpgrade[host.Address] {
+			continue
+		}
+		if !host.IsEtcd {
+			// separating hosts with only worker role so they undergo upgrade in maxUnavailable batches
+			workerOnlyHosts = append(workerOnlyHosts, host)
 		} else {
-			multipleRolesHosts = append(multipleRolesHosts, workerHost)
+			// separating nodes with etcd role, since at this point worker components in controlplane nodes are already upgraded by `UpgradeControlPlaneNodes`
+			// and these nodes will undergo upgrade of worker components sequentially
+			etcdAndWorkerHosts = append(etcdAndWorkerHosts, host)
 		}
 	}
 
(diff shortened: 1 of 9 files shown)


cmd/up.go
```diff
@@ -140,6 +135,14 @@ func ClusterUp(ctx context.Context, dialersOptions hosts.DialersOptions, flags c
 		}
 		kubeCluster.NewHosts = newNodes
 		reconcileCluster = true
+
+		kubeCluster.RemoveHostsLabeledToIgnoreUpgrade(ctx)
+		maxUnavailable, err := kubeCluster.ValidateHostCountForUpgrade()
```
superseb: With `ValidateHostCountForUpgrade` returning an int, this function name should be different. Something along the line of `CalculateHostCountForUpgrade`

cluster/cluster.go
```diff
@@ -165,8 +170,8 @@ func (c *Cluster) DeployWorkerPlane(ctx context.Context, svcOptionData map[strin
 		workerNodePlanMap[workerHost.Address] = BuildRKEConfigNodePlan(ctx, c, workerHost, workerHost.DockerInfo, svcOptionData)
```
superseb: Can we change this loop from `for _, workerHost` to `for _, host`, as we define what it is later by checking what role(s) it has and it's confusing to call it `workerHost` from the beginning. (I know this is not related to this PR specifically)

cluster/cluster.go
```diff
@@ -165,8 +170,8 @@ func (c *Cluster) DeployWorkerPlane(ctx context.Context, svcOptionData map[strin
...
 		if !workerHost.IsControl && !workerHost.IsEtcd {
 			workerOnlyHosts = append(workerOnlyHosts, workerHost)
-		} else {
-			multipleRolesHosts = append(multipleRolesHosts, workerHost)
+		} else if workerHost.IsEtcd {
```
superseb: Please add comment here why we are making this differentiation for people that look at this later

cluster/hosts.go
```diff
@@ -54,14 +56,37 @@ func (c *Cluster) TunnelHosts(ctx context.Context, flags ExternalFlags) error {
...
-		c.WorkerHosts = removeFromHosts(host, c.WorkerHosts)
-		c.RancherKubernetesEngineConfig.Nodes = removeFromRKENodes(host.RKEConfigNode, c.RancherKubernetesEngineConfig.Nodes)
+		c.removeHostFromNodeList(ctx, host)
 	}
 	return ValidateHostCount(c)
+}
+
+func (c *Cluster) RemoveHostsLabeledToIgnoreUpgrade(ctx context.Context) {
+	kubeClient, err := k8s.NewClient(c.LocalKubeConfigPath, c.K8sWrapTransport)
+	nodes, err := kubeClient.CoreV1().Nodes().List(metav1.ListOptions{})
```
superseb: Should we use GetNode here with retries to make sure we did our best to get the nodes?

cluster/hosts.go
```diff
@@ -54,14 +56,37 @@ func (c *Cluster) TunnelHosts(ctx context.Context, flags ExternalFlags) error {
...
+	if err != nil {
+		// not returning error since upgrade can proceed
+		log.Infof(ctx, "[RemoveHostsLabeledToIgnoreUpgrade] Error listing nodes: %v", err)
```
superseb: Can this be changed to inform the user what we are doing based on the fact we can't get nodes? e.g. "Error listing nodes but continuing upgrade"

Can we also keep the logging in either `[plane] message` or just `message` and keep the function log lines for debug?

cluster/hosts.go
```diff
@@ -54,14 +56,37 @@ func (c *Cluster) TunnelHosts(ctx context.Context, flags ExternalFlags) error {
...
+			c.removeHostFromNodeList(ctx, &host)
+			c.InactiveHosts = removeFromHosts(&host, c.InactiveHosts)
+		}
+	}
+	return
+}
 
+func (c *Cluster) removeHostFromNodeList(ctx context.Context, host *hosts.Host) {
+	log.Warnf(ctx, "Removing host [%s] from node lists", host.Address)
+	c.EtcdHosts = removeFromHosts(host, c.EtcdHosts)
```
superseb: Can't this be consolidated into a function removeFromAllHosts and pass it once?

services/workerplane.go
```diff
@@ -105,7 +104,7 @@ func CalculateMaxUnavailable(maxUnavailableVal string, numHosts int) (int, error
 		// In case there is only one node and rounding down maxUnvailable percentage led to 0
 		maxUnavailable = 1
 	}
-	logrus.Infof("%v worker nodes can be unavailable at a time", maxUnavailable)
+	logrus.Infof("Parsed value of maxUnavailable: %v", maxUnavailable)
```
superseb: This seems like we switched this in the previous PR? The current active line is more a debug line, while the first one gives the user info on what the value means. `Parsed value of maxUnavailable` doesn't tell the user what it means.

mrajashree: I did this because maxUnavailable can get reset based on inactive hosts later on. Instead now I have separate info log lines when I explicitly set and if needed rest maxUnavailable
https://github.com/rancher/rke/pull/1912/files#diff-b8aac5efb9904be5c8ebe84055a5927fR144
https://github.com/rancher/rke/pull/1912/files#diff-ee57d1a642f3b4ee28fb4929afe13026R69

cluster/hosts.go
mrajashree: sure,  this function existed prior to my changes but will try

cluster/hosts.go
mrajashree: might be excessive to call GetNode on each host with retries, but I'll add retry logic around this List call

cluster/hosts.go
mrajashree: This function existed prior to my change, also looks like it's being called for each role separately since each of the lists etcdHosts, ControlHosts and workerHosts should get updated

cluster/cluster.go
```diff
@@ -161,12 +166,15 @@ func (c *Cluster) DeployWorkerPlane(ctx context.Context, svcOptionData map[strin
...
-		if !workerHost.IsControl && !workerHost.IsEtcd {
-			workerOnlyHosts = append(workerOnlyHosts, workerHost)
-		} else {
-			multipleRolesHosts = append(multipleRolesHosts, workerHost)
+	for _, host := range allHosts {
+		workerNodePlanMap[host.Address] = BuildRKEConfigNodePlan(ctx, c, host, host.DockerInfo, svcOptionData)
+		// separating hosts with only worker role so they undergo upgrade in maxUnavailable batches
+		if !host.IsControl && !host.IsEtcd {
+			workerOnlyHosts = append(workerOnlyHosts, host)
+		} else if host.IsEtcd {
```
superseb: This will still get "controlplane" nodes in case of all roles

services/workerplane.go
```diff
@@ -68,18 +64,21 @@ func UpgradeWorkerPlane(ctx context.Context, kubeClient *kubernetes.Clientset, m
...
+	if len(inactiveHosts) > 0 {
+		maxUnavailable -= len(inactiveHosts)
+		logrus.Infof("Resetting maxUnavailable to %v since %v host(s) are found to be inactive/unavailable prior to upgrade", maxUnavailable, len(inactiveHosts))
+	}
 
-	updateNewHostsList(kubeClient, append(multipleRolesHosts, workerOnlyHosts...), newHosts)
-	log.Infof(ctx, "First checking and processing worker components for upgrades on nodes with etcd/controlplane roles one at a time")
-	multipleRolesHostsFailedToUpgrade, err := processWorkerPlaneForUpgrade(ctx, kubeClient, multipleRolesHosts, localConnDialerFactory, prsMap, workerNodePlanMap, certMap, updateWorkersOnly, alpineImage, 1, upgradeStrategy, newHosts)
+	updateNewHostsList(kubeClient, append(mixedRolesHosts, workerOnlyHosts...), newHosts)
+	log.Infof(ctx, "First checking and processing worker components for upgrades on nodes with etcd role one at a time")
```
superseb: Can we log `First checking` and `Now checking` conditionally on length of nodes? This shows in all role nodes scenario and doesnt do anything, is there a fallback where this catches something?

mrajashree:
As discussed offline, the above changes will be in a separate PR

Issue #1744, opened 2019-11-03
Upgrade kubernetes cluster on CoreOS with Flexvolume


mohsenmottaghi:
**RKE version:** 
v0.3.2 

**Docker version: (`docker version`,`docker info` preferred)**
```
...
Server Version: 18.06.3-ce
Storage Driver: overlay2
 Backing Filesystem: extfs
 Supports d_type: true
 Native Overlay Diff: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge host macvlan null overlay
 Log: awslogs fluentd gcplogs gelf journald json-file logentries splunk syslog
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: 468a545b9edcd5932181eb9de8e72e92e616e86e
runc version: a592beb5bc4c4092b1c0cac971afed27687340c5
init version: fec3683b97ad9c3ef73f284f176e12c44b448662
Security Options:
 seccomp
  Profile: default
 selinux
Kernel Version: 4.19.66-coreos
Operating System: Container Linux by CoreOS 2191.4.1 (Rhyolite)
OSType: linux
Architecture: x86_64
CPUs: 8
Total Memory: 16.26GiB
Name: kw1.example.com
ID: KTTJ:Q3RN:ZLSU:WTLK:EWZ3:TB3T:DWPK:ONVB:EDQB:Z57U:4SQA:KR1L
Docker Root Dir: /var/lib/docker
Debug Mode (client): false
Debug Mode (server): false
HTTP Proxy: http://proxy.example.com
HTTPS Proxy: http://proxy.example.com
No Proxy: localhost, 127.0.0.0/8, repo.example.com, 172.16.0.0/16, proxy.example.com, proxy.example.com, proxy.example.com
Registry: https://index.docker.io/v1/
Labels:
Experimental: false
Insecure Registries:
 127.0.0.0/8
Live Restore Enabled: false
```
**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)** 4.19.66-coreos
```
$ uname -r
4.19.66-coreos
```
**Type/provider of hosts: (VirtualBox/Bare-metal/AWS/GCE/DO)**
Bar-metal
**cluster.yml file:**
```
# If you intened to deploy Kubernetes in an air-gapped environment,
# please consult the documentation on how to configure custom RKE images.
nodes:
- address: 172.16.101.190
  port: "22"
[...]

PR #1921, opened 2020-02-21
[v1.0] Add custom FlexVol plugin directory support

Oats87:
Adding ability to recognize network plugin specific entry for flexvolume directory

https://github.com/rancher/rke/issues/1744


Oats87:
> Is there a reason why this can't go in `defaults.go`?

No, I'll move it. Thanks

Oats87:
@superseb I moved the defaults to `defaults.go`

cluster/defaults.go
```diff
@@ -64,6 +64,10 @@ const (
 	DefaultFlannelBackendVxLanPort = "8472"
 	DefaultFlannelBackendVxLanVNI  = "1"
 
+	DefaultCalicoFlexVolPluginDirectory = "/usr/libexec/kubernetes/kubelet-plugins/volume/exec/nodeagent~uds"
```
superseb: Do we expect these to be different in the future as we are setting two of the same strings as default?

Oats87: These were kept separate for future-proofing in the event that we need to change it specifically for a plugin without having to change the rest of `defaults.go`

Oats87:
> What is considered regarding this change? This is dependent on kubelet `volume-plugin-dir` and kube-controller-manager `flex-volume-plugin-dir`. Was it considered to base it of the configuration or is it not strictly dependent on that? Does it not need to be validated in any case, is it going to be documented to change what is needed to configure it, or are we only providing a way to move it off the default location?

Generally, when someone is changing the `flex-volume-plugin-dir` they are doing some pretty advanced things on their cluster. I am thinking that we can simply document our way around this by specifying the multiple places you have to specify the `flex-volume-plugin-dir` for both the `kubelet` and `kube-controller-manager` similar to the way we handle service IP CIDR changes today.

Oats87:
> What is considered regarding this change? This is dependent on kubelet `volume-plugin-dir` and kube-controller-manager `flex-volume-plugin-dir`. Was it considered to base it of the configuration or is it not strictly dependent on that? Does it not need to be validated in any case, is it going to be documented to change what is needed to configure it, or are we only providing a way to move it off the default location?

Generally, when someone is changing the `flex-volume-plugin-dir` they are doing some pretty advanced things on their cluster. I am thinking that we can simply document our way around this by specifying the multiple places you have to specify the `flex-volume-plugin-dir` for both the `kubelet` and `kube-controller-manager` similar to the way we handle service IP CIDR changes today.

PR #1922, opened 2020-02-24
Attempt upgrade on NotReady hosts during cluster upgrade

mrajashree:


cluster/defaults.go
```diff
@@ -220,13 +221,15 @@ func (c *Cluster) setClusterDefaults(ctx context.Context, flags ExternalFlags) e
 
 func (c *Cluster) setNodeUpgradeStrategy() {
 	if c.UpgradeStrategy == nil {
-		logrus.Infof("No input provided for maxUnavailable, setting it to default value of %v", DefaultMaxUnavailable)
+		logrus.Debugf("No input provided for maxUnavailable, setting it to default values of %v", DefaultMaxUnavailableWorker, DefaultMaxUnavailableControlplane)
```
superseb: Please describe this as what value for controlplane and what value for workers, there is also only one value to be replaced in the string while 2 values are passed

cluster/cluster.go
```diff
@@ -161,16 +172,35 @@ func (c *Cluster) DeployControlPlane(ctx context.Context, svcOptionData map[stri
 			controlPlaneHosts = append(controlPlaneHosts, host)
 		}
 	}
-	if err := services.UpgradeControlPlaneNodes(ctx, kubeClient, controlPlaneHosts,
+
+	for _, host := range notReadyHosts {
+		notReadyHostNames = append(notReadyHostNames, host.HostnameOverride)
+	}
+	// attempt upgrade on NotReady hosts without respecting maxUnavailable
+	logrus.Infof("Attempting to redeploy controlplane components on following hosts in NotReady status: %v", strings.Join(notReadyHostNames, ","))
```
superseb: it's weird this comment says something different than what we log, redeploy is also different than upgrade.

mrajashree:
https://github.com/rancher/rancher/issues/25554
https://github.com/rancher/rancher/issues/25519

PR #1931, opened 2020-02-28
Maxunavailable reset should check number on inactive hosts

mrajashree:
https://github.com/rancher/rancher/issues/25648


services/node_util.go
```diff
@@ -104,18 +104,21 @@ func CalculateMaxUnavailable(maxUnavailableVal string, numHosts int) (int, error
...
 
-func resetMaxUnavailable(maxUnavailable, lenInactiveHosts int) int {
+func resetMaxUnavailable(maxUnavailable, lenInactiveHosts int, component string) (int, error) {
 	if maxUnavailable > WorkerThreads {
 		/* upgrading a large number of nodes in parallel leads to a large number of goroutines, which has led to errors regarding too many open sockets
 		Because of this RKE switched to using workerpools. 50 workerthreads has been sufficient to optimize rke up, upgrading at most 50 nodes in parallel.
 		So the user configurable maxUnavailable will be respected only as long as it's less than 50 and capped at 50 */
 		maxUnavailable = WorkerThreads
-		logrus.Info("Resetting maxUnavailable to 50, to avoid issues related to upgrading large number of nodes in parallel")
+		logrus.Info("Resetting %s to 50, to avoid issues related to upgrading large number of nodes in parallel", "max_unavailable"+component)
```
superseb: `logrus.Infof` and I assume the outcome of the string concat needs to be `max_unavailable_X`, which seems to be missing an underscore. Same down below

services/node_util.go
```diff
@@ -104,18 +104,21 @@ func CalculateMaxUnavailable(maxUnavailableVal string, numHosts int) (int, error
...
 	}
 
 	if lenInactiveHosts > 0 {
+		if maxUnavailable == lenInactiveHosts {
+			return 0, fmt.Errorf("cannot proceed with upgrade of %s since %v host(s) are found to be inactive/unavailable prior to upgrade", component, lenInactiveHosts)
```
superseb: Can we just use `inactive` to avoid confusion around what is inactive and what is unavailable?

Issue #1050, opened 2018-12-04
Add nodelocal DNS option


superseb:
Next to kubedns and CoreDNS itself, support setting nodelocal option as released in https://github.com/kubernetes/kubernetes/pull/70555 for v1.13

vikas027:
Hey @superseb ,

Is there a workaround by tweaking a file or configMap for the time being?

My use case is that we use `dnsmasq` on the host as well to resolve internal domains by pointing to specific nameservers. In order to do that, I need to have IP `172.17.0.1` (default) in `/etc/resolv.conf` of the docker container.

deniseschannon:
We should add this as an option in k8s 1.15-k8s 1.18 templates for when we release 2.4 but not turned on by default.

deniseschannon:
Available with v1.1.0-rc10

superseb:
Can be tested using:

```
dns:
  provider: coredns
  nodelocal:
    ip_address: "169.254.20.10"
```

Things to test:

* `--cluster-dns` on the `kubelet` container should be set to this address
* internal and external dns resolution should work

jiaqiluo:
the enhancement is validated with RKE version v1.1.0-rc10
```
>./rke -v
rke version v1.1.0-rc10
```

Use RKE to provision a cluster with the following yml file
```
nodes:
  - address: 
    internal_address: 
    user: ubuntu
    role: [etcd, controlplane, worker]
    ssh_key_path:  
  - address: 
    internal_address: 
    user: ubuntu
    role: [etcd, controlplane, worker]
    ssh_key_path:  
  - address:  
    internal_address: 
    user: ubuntu
    role: [etcd, controlplane, worker]
    ssh_key_path:  

dns:
  provider: coredns
  nodelocal:
    ip_address: "169.254.20.10"
```

Do the following checks on the cluster:
- confirm that a daemonset named `node-local-dns` is deployed in the cluster 
```
> k get daemonsets.apps -n kube-system
NAME             DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR   AGE
canal            3         3         3       3            3           <none>          20m
node-local-dns   3         3         3       3            3           <none>          20m
```

- confirm the `--cluster-dns` on the `kubelet` container should be set to designated address (need to ssh into nodes)
```
ubuntu@ip-172-31-25-27:~$ docker inspect kubelet | grep -e  'cluster-dns'
            "--cluster-dns=169.254.20.10",
[...]

PR #1932, opened 2020-02-28
Add support for nodelocal DNS, update Canal/Calico templates with privileged and add rke-tools:v0.1.53

superseb:
https://github.com/rancher/rke/issues/1050

Depends on:
https://github.com/rancher/kontainer-driver-metadata/pull/149 && https://github.com/rancher/kontainer-driver-metadata/pull/150
https://github.com/rancher/types/pull/1108 && https://github.com/rancher/types/pull/1114


cluster/validation.go
```diff
@@ -344,6 +344,11 @@ func validateDNSImages(c *Cluster) error {
 			return fmt.Errorf("coredns autoscaler image is not populated")
 		}
 	}
+	if c.DNS.Nodelocal != nil {
+		if len(c.SystemImages.Nodelocal) == 0 {
```
luthermonson: do this in one line?

cluster/validation.go
```diff
@@ -344,6 +344,11 @@ func validateDNSImages(c *Cluster) error {
...
+			return fmt.Errorf("nodelocal image is not populated")
```
luthermonson: error.New

superseb: Replaced all non formatted errors to `errors.New`

cluster/validation.go
superseb: This was to keep the style with the other systemimages checks, but as this is only one separate condition I merged them

PR #1939, opened 2020-03-03
Add flag to enable TRACE logging

superseb:
https://github.com/rancher/rancher/issues/24994

I've tried to think of a couple of ways to do this:

- Implement a loglevel flag and try to keep `--debug` and `--quiet`, and offer a new loglevel flag. But the new loglevel flag would only serve the new TRACE level, because we are not offering any per level logging. It's debug or quiet (or TRACE after this) so there is no real point in having a separate flag.
- Adding trace flag as below seems to be not bad, it's a global level flag and it is an exception to use it (TRACE will be logging everything, including sensitive info), so having a specific flag to enable it should be OK. I didn't include a shorthand for it (`-t` or `-T` because it probably gets confusing and it's not that common for people to know what `-t` is and `-T` is maybe more common but then uppercase so probably confusing. As its an exception to use it, people probably want to use it written out to make sure they know what they are enabling.

### DIFF
--- a/main.go
+++ b/main.go
@@ -32,11 +32,17 @@ func mainErr() error {
 	app.Version = VERSION
 	app.Usage = "Rancher Kubernetes Engine, an extremely simple, lightning fast Kubernetes installer that works everywhere"
 	app.Before = func(ctx *cli.Context) error {
-		if ctx.GlobalBool("debug") {
-			logrus.SetLevel(logrus.DebugLevel)
-		}
 		if ctx.GlobalBool("quiet") {
 			logrus.SetOutput(ioutil.Discard)
+		} else {
+			if ctx.GlobalBool("debug") {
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -34,6 +34,11 @@ func mainErr() error {
 	app.Before = func(ctx *cli.Context) error {
 		if ctx.GlobalBool("debug") {
 			logrus.SetLevel(logrus.DebugLevel)
+			logrus.Debugf("Loglevel set to [%v]", logrus.DebugLevel)
+		}
+		if ctx.GlobalBool("trace") {
```
luthermonson: not sure if you care but if you do --trace --debug --quiet you'll get two outputs for setting the log level and then it'll be quiet. maybe check quiet first and do an else for the other two?

PR #1942, opened 2020-03-04
Add support for zone and region in vSphere

jorioux:
vSphere Cloud Provider added Zones support since Kubernetes version 1.14, and this PR adds the support for those in RKE.

https://vmware.github.io/vsphere-storage-for-kubernetes/documentation/zones.html

### DIFF
--- a/templates/vsphere.go
+++ b/templates/vsphere.go
@@ -60,4 +60,12 @@ vm-name = "{{ .VsphereConfig.Global.VMName }}"
         {{- if ne .VsphereConfig.Network.PublicNetwork "" }}
         public-network = "{{ .VsphereConfig.Network.PublicNetwork }}"
         {{- end }}
+
+[Labels]
+        {{- if ne .VsphereConfig.Labels.Zone "" }}
+        zone = "{{ .VsphereConfig.Labels.Zone }}"
+        {{- end }}
+        {{- if ne .VsphereConfig.Labels.Region "" }}
+        region = "{{ .VsphereConfig.Labels.Region }}"
+        {{- end }}
 `


jorioux:
Can someone review this PR please? It would be really appreciated

finkr:
The label name should be configurable.

Kubernetes has some well [known names](https://v1-17.docs.kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/#topologykubernetesiozone) for labels.

The canonical name for an availability zone is `topology.kubernetes.io/zone` (some beta names are deprecated, like `failure-domain.beta.kubernetes.io/zone` used in [VMware documentation](https://vmware.github.io/vsphere-storage-for-kubernetes/documentation/zones.html).

Same applies for  `topology.kubernetes.io/region`.

fritz-gerneth:
We were running into the same issue of having non-shared datastores in Vsphere clusters. Right now we are pretty much in a dead-end as we cannot add these labels. What would be needed to move this PR forward?

finkr:
Quick note, mainly for those googling for this problem.

The pull request seems sensible. It's the configuration recommended by VMware [vSphere Storage for Kubernetes](https://vmware.github.io/vsphere-storage-for-kubernetes/documentation/zones.html#kubernetes-setup)

The pull request allow configuring the Tag in VMware that should be mapped to Kubrenetes.

VMware tag are typically named "k8s-region" and "k8s-zone" in VMware examples.

```
    [Labels]
    region = "k8s-region"
    zone = "k8s-zone"
```

Those VMware tags values are mapped to Kubernetes well known labels for availability regions and zones:
Those  labels were `failure-domain.beta.kubernetes.io/zone` and `failure-domain.beta.kubernetes.io/region` (to be replaced by  `topology.kubernetes.io/zone` and `topology.kubernetes.io/region` )

superseb:
@jorioux How did you test this? As the type doesnt have those fields so I dont think it will ever end up in the config

superseb (changes requested):
^^

jorioux:
> @jorioux How did you test this? As the type doesnt have those fields so I dont think it will ever end up in the config

I haven't tested it. I was deploying Rancher on multisite vSphere clusters with zoned datastores, and not being able to specify the zone and region labels was a deal breaker. I ended up using the out-of-tree vSphere storage provider instead (after upgrade to vSphere 7), which worked like a dream with Rancher! That was at Staples, now I don't work there anymore so I don't really care about that PR anymore, but it would be interesting for people stuck with vSphere version 6.5 or 6.0 and multisite clusters. Cheers 🥂

superseb:
As the out of tree provider is the future and we got that documented (https://rancher.com/docs/rancher/v2.5/en/cluster-provisioning/rke-clusters/cloud-providers/vsphere/out-of-tree/), I think users should go that route.

PR #1943, opened 2020-03-04
Rename ignore label and return error on controlplane failure

mrajashree:
https://github.com/rancher/rancher/issues/25606
https://github.com/rancher/rancher/issues/25599

### DIFF
--- a/services/node_util.go
+++ b/services/node_util.go
@@ -3,6 +3,7 @@ package services
 import (
 	"bytes"
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 
@@ -79,7 +80,7 @@ func getNodeListForUpgrade(kubeClient *kubernetes.Clientset, hostsFailed *sync.M
 		if inactiveHosts[node.Labels[k8s.HostnameLabel]] {
 			continue
 		}
-		if val, ok := node.Labels[k8s.IgnoreHostDuringUpgradeLabel]; ok && val == "true" {
+		if val, ok := node.Labels[k8s.IgnoreHostDuringUpgradeLabel]; ok && val == k8s.IgnoreLabelValue {
 			continue
 		}
 		nodeList = append(nodeList, node)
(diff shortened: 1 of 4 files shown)


services/node_util.go
```diff
@@ -114,7 +114,7 @@ func resetMaxUnavailable(maxUnavailable, lenInactiveHosts int, component string)
 	}
 
 	if lenInactiveHosts > 0 {
-		if maxUnavailable == lenInactiveHosts {
+		if lenInactiveHosts >= maxUnavailable {
 			return 0, fmt.Errorf("cannot proceed with upgrade of %s since %v host(s) are found to be inactive prior to upgrade", component, lenInactiveHosts)
```
deniseschannon: Nit: Is inactive the way we want to describe it? There are so many reasons why a host would not be available to upgrade. Or what if the upgrade failed?

services/node_util.go
```diff
@@ -96,7 +96,7 @@ func CalculateMaxUnavailable(maxUnavailableVal string, numHosts int) (int, error
 		logrus.Errorf("Unable to parse max_unavailable, should be a number or percentage of nodes, error: %v", err)
```
deniseschannon: Do we check if it's negative and error out?

services/node_util.go
```diff
@@ -79,25 +80,38 @@ func getNodeListForUpgrade(kubeClient *kubernetes.Clientset, hostsFailed *sync.M
...
 }
 
-func CalculateMaxUnavailable(maxUnavailableVal string, numHosts int) (int, error) {
+func CalculateMaxUnavailable(maxUnavailableVal string, numHosts int, role string) (int, error) {
 	// if maxUnavailable is given in percent, round down
 	maxUnavailableParsed := k8sutil.Parse(maxUnavailableVal)
 	logrus.Debugf("Provided value for maxUnavailable: %v", maxUnavailableParsed)
+	if maxUnavailableParsed.Type == k8sutil.Int {
+		if maxUnavailableParsed.IntVal <= 0 {
+			return 0, fmt.Errorf("invalid input for max_unavailable_%s: %v, value must be >= 1", role, maxUnavailableParsed.IntVal)
```
deniseschannon: Should it just be > 0 since it could be an int or percentage?

services/node_util.go
```diff
@@ -79,25 +80,38 @@ func getNodeListForUpgrade(kubeClient *kubernetes.Clientset, hostsFailed *sync.M
...
+		}
+	} else {
+		if strings.HasPrefix(maxUnavailableParsed.StrVal, "-") {
```
deniseschannon: Can this just be consolidated to check if it's negative or 0 and just ask that it's >0 as the error message?

services/node_util.go
deniseschannon: Could we update to indicate what "inactive" means in this case?

services/node_util.go
mrajashree: I'm checking separately for percentage type, `maxUnavailableParsed.IntVal` won't hold any percentage values

services/node_util.go
```diff
@@ -79,25 +80,35 @@ func getNodeListForUpgrade(kubeClient *kubernetes.Clientset, hostsFailed *sync.M
...
+func CalculateMaxUnavailable(maxUnavailableVal string, numHosts int, role string) (int, error) {
 	// if maxUnavailable is given in percent, round down
 	maxUnavailableParsed := k8sutil.Parse(maxUnavailableVal)
 	logrus.Debugf("Provided value for maxUnavailable: %v", maxUnavailableParsed)
+	if maxUnavailableParsed.Type == k8sutil.Int {
+		if maxUnavailableParsed.IntVal <= 0 {
+			return 0, fmt.Errorf("invalid input for max_unavailable_%s: %v, value must be > 0", role, maxUnavailableParsed.IntVal)
+		}
+	} else {
+		if strings.HasPrefix(maxUnavailableParsed.StrVal, "-") && maxUnavailableParsed.StrVal == "0%" {
```
deniseschannon: shouldn't this be an OR?

services/node_util.go
```diff
@@ -79,25 +80,35 @@ func getNodeListForUpgrade(kubeClient *kubernetes.Clientset, hostsFailed *sync.M
...
+			return 0, fmt.Errorf("invalid input for max_unavailable_%s: %v, percentage input should be > 0", role, maxUnavailableParsed.StrVal)
```
deniseschannon: Small Nit: But in theory, it should have the same message as above to say `value must be > 0`

deniseschannon:
https://github.com/rancher/rke/issues/1906

Issue #1944, opened 2020-03-05
core-dns addon pod is stuck in pending state


soumyalj:
**RKE version:** rke version v1.1.0-rc10

Steps:
1. Create a RKE cluster using K8s version v1.16.7-rancher1-2
Cluster config: 1 etcd 2 control node and 2 workers
2. Upgrade the cluster using K8s version v1.17.3-rancher1-2

One of the core-dns pod is stuck in pending state as below :

```
soumyas-MBP:rke soumya$ kubectl get pod --kubeconfig kube_config_clusterzeroworker.yml -l k8s-app=kube-dns -n kube-system 
NAME                       READY   STATUS    RESTARTS   AGE
coredns-5c59fd465f-lsp2b   1/1     Running   0          16m
coredns-7c5566588d-l6hh7   0/1     Pending   0          16m
coredns-7c5566588d-vxxbv   1/1     Running   0          15m
```
kubectl describe of the pending pod `coredns-7c5566588d-l6hh7`  has the below error 
```
Events:
  Type     Reason            Age        From               Message
  ----     ------            ----       ----               -------
  Warning  FailedScheduling  <unknown>  default-scheduler  0/5 nodes are available: 2 node(s) didn't match pod affinity/anti-affinity, 3 node(s) didn't match node selector.
  Warning  FailedScheduling  <unknown>  default-scheduler  0/5 nodes are available: 2 node(s) didn't match pod affinity/anti-affinity, 3 node(s) didn't match node selector.

```

Note: This issue does not happen in a cluster with 3 worker nodes

deniseschannon:
These errors aren't related to the RKE upgrade logic changes, but to the scheduling logic in the coredns template.

Oats87:
This appears to happen due to the logic we have for the autoscaler + coredns deployment.

https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.4/rke/templates/coredns.go#L275 indicates that `nodesPerReplica` is `4`, and according to the `cluster-proportional-autoscaler` logic of `ceil( nodes * 1/nodesPerReplica )` and `5` nodes in the cluster, we would end at a desired number of replicas as 2.

In conjunction with the `RollingUpdate` strategy being set with a `MaxUnavailable` of `1`, the third replica cannot be scheduled.

More investigation will be done into this. @soumyalj did this not occur in earlier RKE versions for CoreDNS upgrades?

deniseschannon:
[...]

PR #1947, opened 2020-03-07
MaxUnavailable and powered off hosts comparison

mrajashree:
1. Compare maxUnavailable with powered off hosts before attempting to reconcile
NotReady hosts https://github.com/rancher/rancher/issues/25841
2. Include powered off hosts as failed hosts for controlplane upgrade to return error https://github.com/rancher/rancher/issues/25599
3. Change coredns upgrade strategy. With addons changes it was changed to have the k8s
default value for a deployment of 25% maxUnavailable and maxSurge. This commit changes it
back to maxUnavailable of 1 to avoid dns addon upgrade issues https://github.com/rancher/rke/issues/1944


mrajashree:
Not ready for review

Issue #1928, opened 2020-02-27
Check certs expiration before restoring snapshot


maggieliu:
Before restoring etcd snapshot via RKE CLI, check if certs are expiring. If can't find cert or it's expired exit restoration.

deniseschannon:
Available with RKE v1.1.0-rc16

sowmyav27:
**Verified with rke `v1.1.0-rc16`**

1. Deploy a cluster expired with certs (using rke cli binary with certs expiry in 15 minutes)
2. Take a snapshot
3. Wait for the cluster certs to expire
4. restore from snapshot (taken in step 2) using rke cli 1.1.0-rc16 gives me error - 
```
./rke etcd snapshot-restore --name test_certs --config rancher-cluster.yml
WARN[0000] This is not an officially supported version (v1.1.0-rc16) of RKE. Please download the latest official release at https://github.com/rancher/rke/releases/latest 
INFO[0000] Running RKE version: v1.1.0-rc16             
INFO[0000] Restoring etcd snapshot test_certs           
FATA[0000] [etcd] Failed to restore etcd snapshot: invalid certs: Certificate [kube-node] failed verification: x509: certificate has expired or is not yet valid: Certificate [kube-etcd-172-31-4-104] failed verification: x509: certificate has expired or is not yet valid: Certificate [kube-apiserver] failed verification: x509: certificate has expired or is not yet valid: Certificate [kube-scheduler] failed verification: x509: certificate has expired or is not yet valid: Certificate [kube-etcd-172-31-0-167] failed verification: x509: certificate has expired or is not yet valid: Certificate [kube-controller-manager] failed verification: x509: certificate has expired or is not yet valid: Certificate [kube-proxy] failed verification: x509: certificate has expired or is not yet valid: Certificate [kube-etcd-172-31-6-140] failed verification: x509: certificate has expired or is not yet valid: Certificate [kube-ca] failed verification: x509: certificate has expired or is not yet valid: Certificate [kube-apiserver-proxy-client] failed verification: x509: certificate has expired or is not yet valid: Certificate [kube-apiserver-proxy-client] failed verification: x509: certificate has expired or is not yet valid 
```
5. Rotate certs - `./rke cert rotate --config rancher-cluster.yml` (rotates all certs except kube-ca)
6. Error seen when trying to restore to snapshot (taken in step 2) - 
```
./rke etcd snapshot-restore --name test_certs --config rancher-cluster.yml
[...]

PR #1953, opened 2020-03-11
Skip restore if no valid certs

rmweir:
**Issue:**
https://github.com/rancher/rke/issues/1928

### DIFF
--- a/cmd/etcd.go
+++ b/cmd/etcd.go
@@ -2,10 +2,12 @@ package cmd
 
 import (
 	"context"
+	"crypto/x509"
 	"fmt"
 	"strings"
 	"time"
 
+	"github.com/pkg/errors"
 	"github.com/rancher/rke/cluster"
 	"github.com/rancher/rke/hosts"
 	"github.com/rancher/rke/log"
@@ -129,6 +131,11 @@ func RestoreEtcdSnapshot(
 	if err != nil {
 		return APIURL, caCrt, clientCert, clientKey, nil, err
 	}
+
+	if err := validateCerts(rkeFullState.DesiredState); err != nil {
+		return APIURL, caCrt, clientCert, clientKey, nil, err
+	}
+
 	if err := checkLegacyCluster(ctx, kubeCluster, rkeFullState, flags); err != nil {
 		return APIURL, caCrt, clientCert, clientKey, nil, err
 	}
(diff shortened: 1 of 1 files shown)


cmd/etcd.go
```diff
@@ -129,6 +129,11 @@ func RestoreEtcdSnapshot(
 	if err != nil {
 		return APIURL, caCrt, clientCert, clientKey, nil, err
 	}
+
+	if err := validateETCDCerts(kubeCluster.EtcdHosts, rkeFullState.CurrentState); err != nil {
```
galal-hussein: I think this should be desired state not the current state

cmd/etcd.go
```diff
@@ -180,6 +185,20 @@ func RestoreEtcdSnapshot(
...
 }
 
+func validateETCDCerts(etcdHosts []*hosts.Host, state cluster.State) error {
+	if state.RancherKubernetesEngineConfig == nil {
+		// possibly already started a restore
+		return nil
+	}
+	for _, etcdHost := range etcdHosts {
+		certName := pki.GetCrtNameForHost(etcdHost, pki.EtcdCertName)
+		if cert := state.CertificatesBundle[certName].Certificate; cert == nil || cert.NotAfter.Before(time.Now().Add(time.Hour)) {
```
mrajashree: should we check for presence of they key certName in the CertificatesBundle map?

cmd/etcd.go
mrajashree: should we be passing desiredState?

cmd/etcd.go
```diff
@@ -180,6 +185,20 @@ func RestoreEtcdSnapshot(
 	return APIURL, caCrt, clientCert, clientKey, certs, err
 }
 
+func validateETCDCerts(etcdHosts []*hosts.Host, state cluster.State) error {
+	if state.RancherKubernetesEngineConfig == nil {
+		// possibly already started a restore
+		return nil
+	}
+	for _, etcdHost := range etcdHosts {
+		certName := pki.GetCrtNameForHost(etcdHost, pki.EtcdCertName)
```
galal-hussein: I am not sure that we need actually to check for the etcd cert, since it should be regenerated as a part of rke updates which will be executed as a part of the restore, meanwhile I think we should check for other certs

cmd/etcd.go
```diff
@@ -180,6 +187,42 @@ func RestoreEtcdSnapshot(
...
 
+func validateCerts(state cluster.State) error {
+	var failedErrs error
+
+	if state.RancherKubernetesEngineConfig == nil {
+		// possibly already started a restore
+		return nil
+	}
+	for name, certPKI := range state.CertificatesBundle {
+		if strings.Contains(name, "token") || strings.Contains(name, "header") || strings.Contains(name, "admin") {
```
superseb: Can we change this to match the constants in https://github.com/rancher/rke/blob/master/pki/constants.go?

superseb:
Just one comment, rest LGTM

PR #1963, opened 2020-03-13
Enable kube-api audit log for new k8s versions

superseb:


deniseschannon:
Instead of `CISDefaults`, can this be new defaults as of RKE v1.1?

Issue #1974, opened 2020-03-17
Cert rotate should not proceed if kube-ca has expired


sowmyav27:
**RKE version:** 
 v1.1.0-rc16

**Type/provider of hosts: (VirtualBox/Bare-metal/AWS/GCE/DO)**
AWS

**cluster.yml file:**
```
nodes:
  - address: <ip-1>
    internal_address: <p-ip-1>
    user: ubuntu
    role: [controlplane,worker,etcd]
    ssh_key_path: <>
  - address: <ip-2>
    internal_address: <p-ip-2>
    user: ubuntu
    role: [controlplane,worker,etcd]
    ssh_key_path: <>
  - address: <ip-3>
    internal_address: <p-ip-3>
    user: ubuntu
    role: [controlplane,worker,etcd]
    ssh_key_path: <>

services:
  etcd:
    snapshot: true
    creation: 6h
    retention: 24h
```

**Steps to Reproduce:**

- Deploy a cluster expired with certs (using rke cli binary with certs expiry in 15 minutes)
- Wait for the cluster certs to expire
- Do a cert rotate
- Rotate certs - ./rke cert rotate --config rancher-cluster.yml (rotates all certs except kube-ca)

**Results:**
- certs get rotate but kube-apiserver on the nodes in restarting state and the cluster is not accessible via kubectl.
- There is panic in the logs of the kube-apiserver
```
+ '[' kube-apiserver = kubelet ']'
[...]

PR #1977, opened 2020-03-18
Validate kube-ca cert before rotating certs

mrajashree:
https://github.com/rancher/rke/issues/1974

Problem: `rke cert rotate` rotates all certificates, but it rotates the kube-ca cert only if the flag 
`--rotate-ca` is used. RKE rotates certs even if kube-ca has expired, and expired kube-ca shouldn't be generating certs. 
Solution: RKE shouldn't rotate certs if `--rotate-ca` is not passed to `rke cert rotate` without checking the expiration of kube-ca

### DIFF
--- a/cmd/cert.go
+++ b/cmd/cert.go
@@ -2,10 +2,9 @@ package cmd
 
 import (
 	"context"
+	"crypto/x509"
 	"fmt"
 	"time"
-
-	"github.com/sirupsen/logrus"
 
 	"github.com/rancher/rke/cluster"
 	"github.com/rancher/rke/hosts"
@@ -14,6 +13,7 @@ import (
 	"github.com/rancher/rke/pki/cert"
 	"github.com/rancher/rke/services"
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
 )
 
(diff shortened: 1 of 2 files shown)


luthermonson:
fix the conflict and LGTM

cmd/cert.go
```diff
@@ -236,6 +236,21 @@ func rotateRKECertificates(ctx context.Context, kubeCluster *cluster.Cluster, fl
...
+		}
+		caCert := caCertPKI.Certificate
+		if caCert == nil {
+			return nil, fmt.Errorf("[rotateRKECertificates] certificate [%s] is nil", pki.CACertName)
+		}
+		certPool := x509.NewCertPool()
+		certPool.AddCert(caCert)
+		if _, err := caCert.Verify(x509.VerifyOptions{Roots: certPool, KeyUsages: []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth}}); err != nil {
+			return nil, fmt.Errorf("[rotateRKECertificates] failed to rotate certs: %v", err)
+		}
```
rmweir: nit: could probably put this in common and reuse it at least, gives us a reusable function for something we might need to do elsewhere. Just worth considering but not necessary.

cmd/cert.go
```diff
@@ -236,6 +236,21 @@ func rotateRKECertificates(ctx context.Context, kubeCluster *cluster.Cluster, fl
 		return nil, fmt.Errorf("Failed to rotate certificates: can't find old certificates")
 	}
 	currentCluster.RotateCertificates = kubeCluster.RotateCertificates
+	if !kubeCluster.RotateCertificates.CACertificates {
+		caCertPKI, ok := rkeFullState.CurrentState.CertificatesBundle[pki.CACertName]
+		if !ok {
+			return nil, fmt.Errorf("[rotateRKECertificates] certificate [%s] not found", pki.CACertName)
```
superseb: The errors are user facing and should not use a technical format with the function name. See line 236 for an example.

cmd/cert.go
```diff
@@ -236,6 +236,21 @@ func rotateRKECertificates(ctx context.Context, kubeCluster *cluster.Cluster, fl
...
+		}
+		caCert := caCertPKI.Certificate
+		if caCert == nil {
+			return nil, fmt.Errorf("[rotateRKECertificates] certificate [%s] is nil", pki.CACertName)
+		}
+		certPool := x509.NewCertPool()
+		certPool.AddCert(caCert)
+		if _, err := caCert.Verify(x509.VerifyOptions{Roots: certPool, KeyUsages: []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth}}); err != nil {
+			return nil, fmt.Errorf("[rotateRKECertificates] failed to rotate certs: %v", err)
```
superseb: This should advise the user to use the flag to rotate CA

cmd/cert.go
mrajashree: I was considering that earlier, but since from `rotateRKECertificates` we'd like to return separate error messages for when cert is nil vs when Verify returns an error for it, it didn't make sense to put this in a separate function, and then determine the type of error that function has returned.

PR #1989, opened 2020-03-26
Use v3 type for addons updateStrategy

mrajashree:


cluster/addons.go
```diff
@@ -322,7 +325,10 @@ func (c *Cluster) deployCoreDNS(ctx context.Context, data map[string]interface{}
 		UpstreamNameservers:    c.DNS.UpstreamNameservers,
 		ReverseCIDRs:           c.DNS.ReverseCIDRs,
 		NodeSelector:           c.DNS.NodeSelector,
-		UpdateStrategy:         c.DNS.UpdateStrategy,
+		UpdateStrategy: &appsv1.DeploymentStrategy{
+			Type:          c.DNS.UpdateStrategy.Strategy,
```
ibuildthecloud: I don't know the full context of this code, but is it possible to get a panic here dereferencing nil?

mrajashree: @ibuildthecloud RKE first sets it to default values in cluster/defaults.go https://github.com/rancher/rke/blob/597bc5f2a76f0c896fc75ff15fdea00cdcdbb32d/cluster/defaults.go#L619:19

mrajashree: @ibuildthecloud `c.DNS.UpdateStrategy` or UpdateStrategy field for any addon won't be nil because if not provided, RKE defaults it to RollingUpdate when setting cluster defaults [here](https://github.com/rancher/rke/blob/master/cluster/defaults.go#L619:19)

ibuildthecloud: Does the same apply for any other place in this PR where multiple fields are dereferenced at once?

mrajashree: Yes, except for nodeLocal dns which doesn't get default values, changed it to add a nil check https://github.com/rancher/rke/pull/1989#diff-e072bfb1478c2c4b66ce16353b1b3027R665

Issue #1916, opened 2020-02-19
Removing a user addon in the cluster.yml file and performing "rke up" does not remove the addon from the cluster


soumyalj:
**RKE version:** v1.1.0-rc6

Steps:
1. Create a RKE cluster including a user defined addon in the cluster.yml file.
```
addons: |-
    ---
    apiVersion: v1
    kind: Pod
    metadata:
      name: my-nginx
      namespace: default
    spec:
      containers:
      - name: my-nginx
        image: nginx
        ports:
        - containerPort: 80

```
2. After the cluster is up and running, verify the addon is created successfully.
3. In the cluster.yml file, remove the addon section and perform a  rke up. After the rke up is complete, the user-addons do not get removed from the cluster.
The rke logs display that user addons are not defined, but they still exist in the cluster

```
NFO[0034] [addons] Successfully saved ConfigMap for addon rke-metrics-addon to Kubernetes 
INFO[0034] [addons] Executing deploy job rke-metrics-addon 
INFO[0035] [addons] Metrics Server deployed successfully 
INFO[0035] [ingress] Setting up nginx ingress controller 
INFO[0035] [addons] Saving ConfigMap for addon rke-ingress-controller to Kubernetes 
INFO[0035] [addons] Successfully saved ConfigMap for addon rke-ingress-controller to Kubernetes 
INFO[0035] [addons] Executing deploy job rke-ingress-controller 
INFO[0035] [ingress] ingress controller nginx deployed successfully 
INFO[0035] [addons] Setting up user addons              
INFO[0035] [addons] no user addons defined              
INFO[0035] Finished building Kubernetes cluster successfully 


```
```
soumyas-MBP:rke soumya$ kubectl get pod --kubeconfig kube_config_clusterzeroworker.yml
NAME       READY   STATUS    RESTARTS   AGE
my-nginx   1/1     Running   2          69m
```

cbluth:
i have a feeling that this issue is related: https://github.com/rancher/rke/issues/1913

deniseschannon:
Available to test with RKE v1.2.0-rc1 (master) and RKE v1.1.2-rc1 (Rancher 2.4)

soumyalj:
Tested with rke version v1.1.2-rc1 and rke version v1.2.0-rc1
1. Create a RKE cluster including a user defined addon in the addons.yml file.
2. After the cluster is up and running, verify the addon is created successfully.
3. In the addons.yml file, remove the addon section and perform  rke up. 
[...]

PR #1990, opened 2020-03-27
Remove user addons when not present

superseb:
https://github.com/rancher/rke/issues/1916

This implements similar logic to configuring `provider: none` to monitoring/ingress, if there are no addons defined, check for the addon job and remove it. The logic that needed changing was that `-o name` provides you with `<kind>/<name>` which works for `kubectl delete` but it doesn't include namespace so resources in the non-default namespace (which is `kube-system`) were not removed.

Example:

When deleting the `ingress` addon, the output of the command how it is today is:

```
kubectl get -f ingress.yaml -o name
namespace/ingress-nginx
configmap/nginx-configuration
configmap/tcp-services
configmap/udp-services
serviceaccount/nginx-ingress-serviceaccount
clusterrole.rbac.authorization.k8s.io/nginx-ingress-clusterrole
role.rbac.authorization.k8s.io/nginx-ingress-role
rolebinding.rbac.authorization.k8s.io/nginx-ingress-role-nisa-binding
clusterrolebinding.rbac.authorization.k8s.io/nginx-ingress-clusterrole-nisa-binding
daemonset.apps/nginx-ingress-controller
deployment.apps/default-http-backend
service/default-http-backend
```

Which wouldn't work if we didn't include the namespace in there, as all the resource are being deleted are not deleted in the namespace but because the namespace is in there, everything gets deleted and the rest is ignored (``--ignore-not-found=true`).

For the other addons, they are in `kube-system` which is the default namespace for `kubectl` commands so that also works.

For the user addons, the namespace resource will not always be in the definition and without it, it cannot be deleted (as it defaults to `kube-system` and might be created in `default` or basically wherever. The output for addons how it is today:

```
pod/nginx
```

Which does not include the namespace. The fix is to list all parameters needed to properly delete them (and exclude namespace parameter if namespace is `<none>` (not namespaced)):

```
kubectl delete Namespace ingress-nginx                                                                                                                         
kubectl -n ingress-nginx delete ConfigMap nginx-configuration                                                                                                  
kubectl -n ingress-nginx delete ConfigMap tcp-services                                                                                                                                                                                                                                                                        kubectl -n ingress-nginx delete ConfigMap udp-services  
kubectl -n ingress-nginx delete ServiceAccount nginx-ingress-serviceaccount                                                                                    
kubectl delete ClusterRole nginx-ingress-clusterrole                                                                                                           
kubectl -n ingress-nginx delete Role nginx-ingress-role                                                                                                        
kubectl -n ingress-nginx delete RoleBinding nginx-ingress-role-nisa-binding                                                                                    
kubectl delete ClusterRoleBinding nginx-ingress-clusterrole-nisa-binding                                                                                       
kubectl -n ingress-nginx delete DaemonSet nginx-ingress-controller                                                                                             
kubectl -n ingress-nginx delete Deployment default-http-backend                                                                                                
kubectl -n ingress-nginx delete Service default-http-backend                              
```

### DIFF
--- a/cluster/addons.go
+++ b/cluster/addons.go
@@ -33,6 +33,8 @@ const (
 
 	IngressAddonJobName            = "rke-ingress-controller-deploy-job"
 	MetricsServerAddonJobName      = "rke-metrics-addon-deploy-job"
+	UserAddonJobName               = "rke-user-addon-deploy-job"
+	UserAddonIncludeJobName        = "rke-user-includes-addons-deploy-job"
 	MetricsServerAddonResourceName = "rke-metrics-addon"
 	NginxIngressAddonAppName       = "ingress-nginx"
 	KubeDNSAddonAppName            = "kube-dns"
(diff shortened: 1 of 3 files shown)


cluster/addons.go
```diff
@@ -33,6 +33,8 @@ const (
 
 	IngressAddonJobName            = "rke-ingress-controller-deploy-job"
 	MetricsServerAddonJobName      = "rke-metrics-addon-deploy-job"
+	UserAddonJobName               = "rke-user-addon-deploy-job"
+	UserAddonIncludeJobName        = "rke-user-includes-addons-deploy-job"
```
kinarashah: v nit: should this be renamed to `user-addons-include` since the addon field is addons Include

superseb: I have to double-check this doesn't cause it to not find the job and re-deploy

superseb: Changing this would not re-deploy, but would not apply the change to remove as it won't find the job by the old name as it will look for the new name. Also, we don't want every cluster to create new jobs for user addons deployment.

PR #1994, opened 2020-03-27
[release/v1.1] update bindata/data.json

superseb:


rancherbot (approved):

Issue #2014, opened 2020-04-03
[Backport 2.3] Add nodelocal DNS option


maggieliu:
Backport nodelocal DNS to 2.3: https://github.com/rancher/rke/issues/1050

jiaqiluo:
on `v2.3-head`, with the KDM points to the `dev-v2.3` by default, here is the list of k8s versions showing in the UI
<img width="896" alt="Screen Shot 2020-04-20 at 12 29 21 PM" src="https://user-images.githubusercontent.com/6218999/79791794-bea6d800-8302-11ea-9fd7-f2f279e2f9dc.png">

we expected to see the following versions
```
v1.15.11-rancher1-3
v1.16.8-rancher1-3
v1.17.4-rancher1-3
```

Also, since we are adding new miner versions `1.16.9` and `1.17.5` which will make the `rancher1-3` versions not showing in UI.

mrajashree:
We are removing `v1.16.8-rancher1-3` and `v1.17.4-rancher1-3`, so only `1.16.9` and `1.17.5` will be presented

jiaqiluo:
Failed to validate the lcoal nodeDNS option on `v2.3.7-rc2` for all 3 k8s versions:
v1.17.5-rancher1-1
v1.16.9-rancher1-1
v1.15.11-rancher1-3

I set this while provisioning the cluster
``` dns:
    node_selector: null
    nodelocal:
      ip_address: 169.254.20.10
      node_selector: null
    reversecidrs: null
    stubdomains: null
    upstreamnameservers: null
```

but in the control plan node the address is not applied.
```root@jiaqi-1-15-c1:/home/ubuntu# sudo docker inspect kubelet | grep -e  'cluster-dns'
            "--cluster-dns=10.43.0.10",
                "--cluster-dns=10.43.0.10",
```

jiaqiluo:
The enhancement is validated on `v2.3.7-rc3` with all 3 k8s versions:
```
v1.17.5-rancher1-1
v1.16.9-rancher1-1
v1.15.11-rancher1-3
```


Set the nodeLocal ip_address to `169.254.20.10 ` when provisioning the cluster
Confirm the `--cluster-dns` on the `kubelet` container is set to the designated address (need to ssh into nodes)
```
ubuntu@ip-172-31-25-27:~$ docker inspect kubelet | grep -e  'cluster-dns'
            "--cluster-dns=169.254.20.10",
                "--cluster-dns=169.254.20.10",
```
[...]

PR #2025, opened 2020-04-17
[release/v1.0] Add nodelocal DNS

superseb:
https://github.com/rancher/rke/issues/2014

Only changes needed to keep using the same template is https://github.com/rancher/rke/pull/2025/files#diff-e072bfb1478c2c4b66ce16353b1b3027R104 and https://github.com/rancher/rke/pull/2025/files#diff-e072bfb1478c2c4b66ce16353b1b3027R630. This functionality was added in v2.4.0 as part of zero downtime and I marked it as string and put in the nil value to trigger the default block in the template.


cluster/addons.go
```diff
@@ -573,11 +584,63 @@ func (c *Cluster) deployDNS(ctx context.Context, data map[string]interface{}) er
...
+		}
+	}
+	if c.DNS.Nodelocal != nil && c.DNS.Nodelocal.IPAddress != "" {
+		if err := c.deployNodelocal(ctx, data); err != nil {
+			if err, ok := err.(*addonError); ok && err.isCritical {
+				return err
+			}
+			log.Warnf(ctx, "Failed to deploy addon execute job [%s]: %v", getAddonResourceName(Nodelocal), err)
+		}
+		return nil
```
luthermonson, line 616: probably dont need this one, it'll fall to the one outside of this block

superseb: I agree, but it being a backport PR I'd rather file an issue and change it everywhere so avoid differences.

PR #2044, opened 2020-05-04
Add support for k8s 1.18

aiyengar2:
This PR addresses the change in defaults addressed in https://github.com/rancherlabs/rancher-security/issues/412 and adds in the new data necessary for the KDM changes in https://github.com/rancher/kontainer-driver-metadata/pull/227.

Dependent PRs: https://github.com/rancher/kontainer-driver-metadata/pull/227

Related Issues: https://github.com/rancher/rancher/issues/25117, https://github.com/rancherlabs/rancher-security/issues/412

PR #2051, opened 2020-05-11
bump RKEVersionDev and log it if not supported version

kinarashah:
Tests failing cuz dev version not updated https://github.com/rancher/rke/pull/2044

### DIFF
--- a/main.go
+++ b/main.go
@@ -48,7 +48,8 @@ func mainErr() error {
 			metadata.RKEVersion = app.Version
 			return nil
 		}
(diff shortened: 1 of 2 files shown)


superseb:
Can we not fix this by adding logic how this version is parsed/handled versus keeping the need of bump of version(s)?

kinarashah:
> Can we not fix this by adding logic how this version is parsed/handled versus keeping the need of bump of version(s)?

Yep sure, that's the idea but bumped it up rn cuz it was required for 1.18 to get in for this PR https://github.com/rancher/rke/pull/2044 
@aiyengar2 has updated RKE Version in his PR, keeping this open to revisit this logic

PR #2058, opened 2020-05-19
Dont relabel volumes on upstream Docker & SELinux

superseb:
https://github.com/rancher/rancher/issues/26871


superseb:
Accompanied by https://github.com/rancher/rancher/pull/27280

PR #2079, opened 2020-05-27
Revert default encryption in k8s 1.18

aiyengar2:


metadata/metadata.go
```diff
@@ -90,7 +90,7 @@ func readFile(file string) ([]byte, error) {
 	return ioutil.ReadFile(file)
 }
 
-const RKEVersionDev = "v0.2.3"
+const RKEVersionDev = "v1.1.1-rc100"
```
deniseschannon: Why is this being undone?

aiyengar2: The first commit reverted all changes from https://github.com/rancher/rke/pull/2044, which set the default encryption for k8s 1.18 clusters and bumped the RKEVersionDev for k8s 1.18 to account for the new minRKEVersion: https://github.com/rancher/kontainer-driver-metadata/blob/2d451d5eaf91384e1a5a066a70813dfc5194d48d/rke/k8s_version_info.go#L148

The change in the second commit ensures that we only undo the default encryption and not the RKEVersionDev bump.

PR #2095, opened 2020-05-30
[0.2] vendor update types for weave 2.6.4

kinarashah:


rancherbot (approved):

PR #2098, opened 2020-06-01
Check rollout status in integration test

superseb:
This is to avoid issues where provisioning succeeds (`rke up`), which means the components can be deployed using Docker and needed manifests are technically valid but the resources are not getting ready/healthy. Although this is something we should look at in improving in RKE, the first step is to make sure PRs won't go in which break cluster from becoming healthy (for example, changing system images without supporting templates)

### DIFF
--- a/dind/dind.go
+++ b/dind/dind.go
@@ -63,6 +63,7 @@ func StartUpDindContainer(ctx context.Context, dindAddress, dindNetwork, dindSto
 				"sh",
 				"-c",
 				"mount --make-shared / && " +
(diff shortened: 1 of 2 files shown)


ryansann (dismissed):

superseb:
This was updated to correctly determine if a version failed for rollout. Also adds another `mount --make-shared` for `/sys` because the newly introduced Calico templates require `/sys` for eBPF.

ryansann (approved):

Issue #2123, opened 2020-06-16
Remove unneeded advertised port 4001 for etcd


dgiebert:
**RKE version:**
v1.1.2

**Docker version: (`docker version`,`docker info` preferred)**
19.03.5

**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)**
4.14.138-rancher

**Type/provider of hosts: (VirtualBox/Bare-metal/AWS/GCE/DO)**
Bare-metal

**cluster.yml file:**
Snippet:
```
services:
    etcd:
      backup_config:
        interval_hours: 12
        retention: 6
        s3backupconfig:
          access_key: ""
          secret_key: ""
          bucket_name: ""
          region: ""
          folder: ""
          endpoint: ""
      extra_args:
        election-timeout: '5000'
        heartbeat-interval: '500'
      gid: 1001
      uid: 1001
```

**Results:**
docker inspect etcd:
```
                "/usr/local/bin/etcd",
                "--heartbeat-interval=500",
                "--initial-cluster-token=etcd-cluster-1",
                "--initial-cluster=etcd-master-1=https://10.44.0.1:2380,etcd-master-2=https://10.44.0.2:2380,etcd-master-3=https://10.44.0.3:2380",
                "--enable-v2=true",
                "--election-timeout=5000",
                "--peer-key-file=/etc/kubernetes/ssl/kube-etcd-10-44-0-3-key.pem",
                "--data-dir=/var/lib/rancher/etcd/",
                "--listen-client-urls=https://10.44.0.3:2379",
                "--trusted-ca-file=/etc/kubernetes/ssl/kube-ca.pem",
                "--peer-trusted-ca-file=/etc/kubernetes/ssl/kube-ca.pem",
                "--peer-client-cert-auth=true",
                "--advertise-client-urls=https://10.44.0.3:2379,https://10.44.0.3:4001",
                "--key-file=/etc/kubernetes/ssl/kube-etcd-10-44-0-3-key.pem",
                "--peer-cert-file=/etc/kubernetes/ssl/kube-etcd-10-44-0-3.pem",
                "--client-cert-auth=true",
                "--cert-file=/etc/kubernetes/ssl/kube-etcd-10-44-0-3.pem",
                "--name=etcd-master-3",
                "--initial-advertise-peer-urls=https://10.44.0.3:2380",
                "--listen-peer-urls=https://10.44.0.3:2380",
                "--initial-cluster-state=new"
```

etcdctl list members:
```
[...]

PR #2127, opened 2020-06-17
Removes etcd port 4001

dgiebert:
https://github.com/rancher/rke/issues/2123


superseb (changes requested):
This needs to be scoped to 1.19 and up to make sure it doesn't replace all etcd containers on all existing clusters on provision. Let me know if you want to do this else we need to create a new PR.

PR #2128, opened 2020-06-17
Conditionally include vSphere cloud provider secret config

nikkelma:
Include vSphere cloud provider secret config instead of literal credentials if provided. Depends on rancher/types#1165.


nikkelma:
Is this the entire set of changes needed to support these two fields introduced in v1.11? Unsure if the in-tree or out-of-tree cloud provider is used.

Issue #1652, opened 2019-09-19
rke 0.2.8 fails to deploy k8s 1.14.6, rke-network-plugin deploy fails + workaround


pasikarkkainen:
rke v0.2.8 doesn't seem to be able to deploy new kubernetes v1.14.6-rancher1 clusters. Earlier kubernetes versions seem to deploy/work OK for me. Problems show up when I'm trying to deploy the latest k8s v1.14.6 version (which is default in rke v0.2.8), as rke's "rke-network-plugin" deployment task/job always fails with a FATAL error.

Environment:
- docker hosts: CentOS 7.6 VMs, running native docker 1.13.1-102 (selinux enabled).
- rke v0.2.8.
- kubernetes v1.14.6-rancher1 (default).
- networking: canal (default).

I'm using stock rke "cluster.yml" configuration generated with "rke config" for 3 nodes, except I've added bastion host info to cluster.yml, to be able to access the 3x docker hosts (fresh CentOS 7.6 VMs) I have.

"rke --debug up" always fails like this:

INFO[0271] [addons] Executing deploy job rke-network-plugin
DEBU[0271] [k8s] waiting for job rke-network-plugin-deploy-job to complete..
FATA[0301] Failed to get job complete status for job rke-network-plugin-deploy-job in namespace kube-system

The FATAL error shows up in around ~20 seconds after the "rke-network-plugin" deployment job was started. If I check the logs on the docker hosts, all seems good, there are no errors that I can find, and networking seems to be succesfully configured/enabled.

I spent quite many hours debugging and troubleshooting this until I figured the logs contained lines about "docker.io/rancher/pause:3.1" container image not being there, so it was pulled as part of the "rke-network-plugin" deploy job.. 

..which made me wonder if this was some kind of timing/timeout issue. So I decided to pre-pull the container image on all the docker hosts before running "rke up", and that was it, now "rke-network-plugin" deployment job/task succeeds properly (in around 5 seconds) !!

So it seems the "rke-network-plugin" deployment job clearly is broken with the "complete status" detection. If there's a maximum timeout configured, it needs to be increased, as that would allow the "rke up" to succeed without me first pre-pulling the container image manually on all the hosts.

I didn't take a look at the rke sources yet, but hopefully this is enough information for the first brave soul who wants to investigate and fix this issue :)

Some of the similar/related issues:
- https://github.com/rancher/rancher/issues/19713
- https://github.com/rancher/rke/issues/1461
[...]

PR #2141, opened 2020-06-23
Change Default Addon Job Timeout to 45s

leflambeur:
In recent tickets, I have seen the rke-network-job in vSphere clusters has failed as it has timed out after just 30s at its default value. (Usually at 31-32s)

Increasing the addon timeout has worked to resolve this per customer but this change should be introduced so that in future other customers do not see this.

In a recent ticket, we tried 60s but seemed excessive from our testing so 45s should be a reasonable compromise.

If this change needs to be extended anywhere else in the code base please let me know

gz#10699
gz#9105

### DIFF
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -14,7 +14,7 @@ import (
 const (
 	DefaultRetries          = 5
 	DefaultSleepSeconds     = 5
-	DefaultTimeout          = 30
+	DefaultTimeout          = 45
 	K8sWrapTransportTimeout = 30
 )
 


pasikarkkainen:
This PR should fix https://github.com/rancher/rke/issues/2143 and https://github.com/rancher/rke/issues/1652 among others

pasikarkkainen:
This PR is still valid and needed, the default addon job timeout in rke is too short.

superseb:
@leflambeur What seemed excessive about 60 seconds as there is a retry mechanism polling the state every time? Do you mean it never reached beyond 45 seconds?

leflambeur:
@superseb More a feeling than a science

Default is 30, doubling that seemed like a big jump so I met it in the middle,

It's also a long time to wait if there are other issues that prevent those jobs running

It's almost 5mins

leflambeur:
From the experience of sitting on customer calls, waiting for it to fail after 60s

leflambeur:
It really just needs to be 1-3s greater so +15s is more than generous

superseb:
@leflambeur Can you include updating https://github.com/rancher/rke/blob/master/types/rke_types.go#L51 as well

leflambeur:
Think I did that right? Types wasn't a thing when I did my original change so I had to rebase with master, hoping it's not overwritten the original change but a cursory look seems like it hasn't

superseb:
@leflambeur can you squash commits to a single one and push?

superseb (approved):

paynejacob (approved):

Issue #1336, opened 2019-05-08
Restoring snapshot without pki bundle or rkestate file


galal-hussein:
For etcd restore to work correctly the certificates generated originally for the cluster must be present, in rke 0.1 we saved the certs as pki.bundle.tar.gz and in 0.2 we moved to cluster.rkestate file, we expect the state file to be present during the restore, however we also save a copy of the state as a configmap with the name `full-cluster-state` which we might fetch from etcd snapshot directly, the steps to implement this might include:

1- restore the etcd snapshot on any node from the cluster.yml
2- fetch the state using etcd client
3- write the state locally on disk
4- continue with restoration

maggieliu:
Discussed Solution:
* Bundle in RKE state file & Cluster.yml file in the snapshot going forward
* When restoring:
          - User can pass in an option for RKE state file location, which will overwrite the one bundled in the snapshot.
          - Check if the certs is valid or not. If it's expired, prompt for certs rotation

soumyalj:
@superseb : Could you please provide details on the final implemented solution? 
1. Do we expect the cluster.yml to be bundled in the snapshot.zip file ? 
2. Should the user be able to provide a location for the rke-state file ? I did not see this option listed in etcd-restore --help
3. Are we checking for cert validity ? 
4. How will this new change affect the etcd snapshot/restore functionality in Rancher ?

superseb:
The implementation is:

- etcd snapshot zip files created using the latest version will contain two files, the actual etcd snapshot file and the `.rkestate` file
- When restoring this snapshot (zip file), RKE will first attempt to recover the `.rkestate` file from the snapshot zip file. If it is found, it will be retrieved, written locally and used to restore the snapshot. If it is not found, it will fallback to the local `.rkestate` file (basically the logic how it always was will be used, which is, use `.rkestate` file if found, if not, treat the cluster as being upgraded from before `.rkestate` file existed)

A description of what I tested is in https://github.com/rancher/rke/pull/2152, it comes down to testing the new functionality (so without `.rkestate` present) but also preserving functionality with both the old snapshot zip files (which do no contain `.rkestate`) and clusters that do no have the new functionality (new rke-tools) but should still be able to restore.

soumyalj:
**Tested with rke version v1.1.5-rc2**

**Use case 1:**
1. Create a cluster using  rke version v1.1.5-rc2
2. Create a pod and create a snapshot snp1 
3. Create another pod and create a snapshot snp2
[...]

PR #2152, opened 2020-07-09
Able to include and extract state file in snapshot

superseb:
In combination with https://github.com/rancher/rke-tools/pull/100

How this works for saving etcd snapshots:

- Added step that deploys the statefile
- In rke-tools, it will check if there is a state file present, and if yes, it will be included in the snapshot

How this works for restoring etcd snapshots:

- Setup a temporary cluster to create dialers to the nodes (because we dont have the state file)
- Try to extract the state file from the snapshot and read it
- If found, write it locally and use it to restore
- If not found, fall back to the local state file

Most of the decisions are token to keep backwards compatibility, mainly the steps of rke-tools checking if the state file is present, so even when newer versions are used, it will still successfully create a snapshot. For restoring, it will just error out if an older rke-tools image is used as the command to extract the state file is not recognized and will fall back to local file.

The following usecases were tested:

- Create and restore a snapshot created with these changes
- Create and restore a snapshot created with these changes and with the local `.rkestate` file removed
- Create a snapshot with an older version (without included statefile) and restore
- Create and restore a snapshot with the current RKE version and the new rke-tools image

### DIFF
--- a/cmd/etcd.go
+++ b/cmd/etcd.go
@@ -107,6 +107,10 @@ func SnapshotSaveEtcdHosts(
 	flags cluster.ExternalFlags, snapshotName string) error {
 
 	log.Infof(ctx, "Starting saving snapshot on etcd hosts")
+
+	stateFilePath := cluster.GetStateFilePath(flags.ClusterFilePath, flags.ConfigDir)
+	rkeFullState, _ := cluster.ReadStateFile(ctx, stateFilePath)
+
 	kubeCluster, err := cluster.InitClusterObject(ctx, rkeConfig, flags, "")
 	if err != nil {
 		return err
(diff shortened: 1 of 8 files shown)


maggieliu:
Related to issue: https://github.com/rancher/rke/issues/1336

cmd/etcd.go
```diff
@@ -107,6 +107,10 @@ func SnapshotSaveEtcdHosts(
 	flags cluster.ExternalFlags, snapshotName string) error {
 
 	log.Infof(ctx, "Starting saving snapshot on etcd hosts")
+
+	stateFilePath := cluster.GetStateFilePath(flags.ClusterFilePath, flags.ConfigDir)
+	rkeFullState, _ := cluster.ReadStateFile(ctx, stateFilePath)
```
luthermonson, line 112: why are we always eating this error instead of returning? it's in several places and just seems weird. ok if it's on purpose but everywhere i saw it we could return it...

superseb: Because the choice was made to seamlessly upgrade from RKE v0.1.x to v0.2.x which introduced the statefile, so if it's not there, it doesn't mean an error as it can be a cluster to be upgraded to contain/support a statefile. Now we know this probably could have been better but we also have to think of Rancher where we can't interact with cluster provisioning (e.g. ask the user a question before proceeding)

PR #2155, opened 2020-07-11
Remove dependency on rancher/types

ibuildthecloud:


types/rke_types.go
```diff
@@ -203,30 +202,6 @@ type RKEConfigNode struct {
...
 
-type RkeK8sSystemImage struct {
-	types.Namespaced
-	metav1.TypeMeta   `json:",inline"`
-	metav1.ObjectMeta `json:"metadata,omitempty"`
-
-	SystemImages RKESystemImages `yaml:"system_images" json:"systemImages,omitempty"`
-}
-
-type RkeK8sServiceOption struct {
```
galal-hussein, line 214: just one question regarding the removes, where the addons and system images going to be defined?

PR #2158, opened 2020-07-15
Update base images to Ubuntu 20.04

superseb:
This aligns with that `rancher/rancher` uses


superseb:
Updated to 20.04 as `rancher/rancher` is now on 20.04

PR #2159, opened 2020-07-15
Initial commit for ACI template generation

shastrinator:
https://github.com/rancher/rke/issues/2271

Add arguments and defaults to ACI template

Signed-off-by: Kiran Shastri <shastrinator@gmail.com>


cluster/defaults.go
```diff
@@ -71,6 +71,17 @@ const (
 
 	DefaultCanalFlexVolPluginDirectory = "/usr/libexec/kubernetes/kubelet-plugins/volume/exec/nodeagent~uds"
 
+	DefaultApicRefreshTime                           = 1200
+	DefaultOVSMemoryLimit                            = "1Gi"
+	DefaultImagePullPolicy                           = "Always"
+	DefaultServiceMonitorInterval                    = 0
+	DefaultPBRTrackingNonSnat                        = "false"
+	DefaultInstallIstio                              = "false"
+	DefaultIstioProfile                              = "demo"
```
superseb, line 80: Should we set this to `default` and let the user choose if he wants to use `demo` profile?

cluster/defaults.go
```diff
@@ -71,6 +71,17 @@ const (
...
+	DefaultDropLogEnable                             = "true"
```
superseb, line 81: Is DropLog and debug going to be spammy for a default setup? If these are the ACI defaults then it's fine, otherwise please set the default setting/level.

cluster/defaults.go
```diff
@@ -554,6 +571,17 @@ func (c *Cluster) setClusterNetworkDefaults() {
 	if c.Network.WeaveNetworkProvider != nil {
 		networkPluginConfigDefaultsMap[WeavePassword] = c.Network.WeaveNetworkProvider.Password
 	}
+	if c.Network.AciNetworkProvider != nil {
```
superseb, line 574: Why is using `networkPluginConfigDefaultsMap` being skipped here like the other providers?

cluster/network.go
```diff
@@ -284,9 +339,88 @@ func (c *Cluster) doWeaveDeploy(ctx context.Context, data map[string]interface{}
...
+func (c *Cluster) doAciDeploy(ctx context.Context, data map[string]interface{}) error {
+	refreshTime, err := atoiWithDefault(c.Network.AciNetworkProvider.ApicRefreshTime, DefaultApicRefreshTime)
+	if err != nil {
+		return err
+	}
+	serviceMonitorInterval, err := atoiWithDefault(c.Network.AciNetworkProvider.ServiceMonitorInterval, DefaultServiceMonitorInterval)
+	if err != nil {
+		return err
+	}
+	_, clusterCIDR, _ := net.ParseCIDR(c.ClusterCIDR)
```
superseb, line 351: How are input errors handled here? Errors are ignored and there is no validation in `cluster/validation` added

superseb:
Please provide an example how this would be used in the `cluster.yml`.

cluster/defaults.go
shastrinator: IstioProfile of "demo" is the ACI CNI default. But these settings are for ACI-CNI installed istio, which is disabled by default, and Rancher installed Istio would be preferred for customer usage.

cluster/defaults.go
shastrinator: Thanks, dropLog enabled by default is an ACI-CNI default setting, other debugs should be info level, fixed these

cluster/network.go
shastrinator: Thanks, added the validation back for the ignored cases. Cluster/validation seems to be for common(to all CNI) network options, so not adding there.

cluster/defaults.go
shastrinator: Copying to networkPluginConfigDefaultsMap, then to Network.Options[] and checking whether NetworkProvider Option is set and then setting it to Network.Option[] seemed circuitous instead of setting NetworkProviderOption to the default. Besides, moving it to Network.Option seems like polluting the global namespace with CNI specific options, hence followed this path. Please let me know if this is still what you want.

shastrinator: Additionally ACI has more parameters that are not options, no defaults can be given to those. In essence, I am combining all parameters with or without defaults under aciNetworkProvider and not using NetworkOptions.

shastrinator:
> Please provide an example how this would be used in the `cluster.yml`.
Below, I have pasted what I currently use in my cluster.yml:
kubernetes_version: v1.17.9-rancher1-3
network:
  plugin: ACI
  aci_network_provider:
    system_id: "rancher"
    apic_hosts: ["\"10.30.120.180\""]
    token: "3415a616-50d6-45de-851e-4a419e3540b2"
    apic_user_name: "rancher"
    apic_user_key: "LS0tLS1CRUdJTiBQUklWQVRFIEtFWS0tLS0tCk1JSUNkZ0lCQURBTkJna3Foa2lHOXcwQkFRRUZBQVNDQW1Bd2dnSmNBZ0VBQW9HQkFMNnZmRHpZYy8rQTJmRUUKL2xjWU5wdDV5aU95aVhNK0hXY3QvUlMrMFBqSnRqVVh5dDFBbExQQTgwWXF5YUtFUmxHNGM4TkZhaEh2WEREVwpDWngzb2ZsMHdzR3N1MzE2N1ZlaFlYSS9Bb2k2THVOY09McXJ4RGF6N1VGdVkva25Uc3pCS2VjOGZ4MHVGOFkwClZzcFBmL21sN3dVVjFCOG1HV1ZyZFExNDdYZWhBZ01CQUFFQ2dZRUFvVkpycERwbm9MRE5NV2hOR1RnRjh2Rk4KZjNmcmExRVFwUnMyTXBmZ3lhbzFlYU9BWUVEanpWT2ZQU3l6MjhkV2lvcW5mWHU3YWlMNzl1RU56bzh6dXhYdQpET0dwZVNRaHhlSUZ4RS9qaGdCZDFFL09iaWNPbXVPN3R2RzMrZUh4Y0lESHlLRkg3YjVJRTFHNkk2bVRudC9mCkFvT2tCUlRqeUlKbnZJcEY2YVVDUVFEdlk5NHhOek5TNzY0K1ZzNGJhcXlZOWZKL21ETFBWMTI2Q1NIZXNpWWcKVmMyVEM0dkx3TldZUzV3RklxMFVhaHVwbStFRWR2U0JHRVVXVnJENXJCTGpBa0VBeStxQ05Yb2NGQmlHUHhSMApyUW5WN3h1MS83ekQxemFuNXpGOWlaMDVhZXdQM3NvU3VWaVNkTjByOW9IS3ZjNVNYQTdlanNPcEEwWXZDWkVBCmNiVGVxd0pBRHBUdXUrQzQ1NXhoNkxvdzQ4cWpxTTJTYitNOWRSUEZTY2Uwb0hRSmpuNnJmd0Y5Q25ROGpvSk0KaHFPT0YyZy9OeDJmbnozL3RyMWJ0L2lmM3hWME1RSkFFUGVCUUF2Wi9ENFNKWHVCUGNpY3VWbDM2a0I5bHhHZwpKTmQwVjVmeWFBTFJTenRMRllydUhMVXViRU5lYUM2Ym9iZEhBMS82VnlCbTE0ZnVIb2ZySXdKQVhrWDlhUERUCnhpY3BGRXV1ZEtZU1AxbDlHb3hma0pHTFRIZWRpQ3FaaVNGYjFqZW15eHh2RkNTbm55VkpEbWphOVprWmxQWVEKcGZIQWVvQ3ZiMXdoVnc9PQotLS0tLUVORCBQUklWQVRFIEtFWS0tLS0tCg=="
    apic_user_crt: "LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUI2RENDQVZFQ0FnUG9NQTBHQ1NxR1NJYjNEUUVCQlFVQU1Ed3hDekFKQmdOVkJBWVRBbFZUTVJZd0ZBWUQKVlFRS0RBMURhWE5qYnlCVGVYTjBaVzF6TVJVd0V3WURWUVFEREF4VmMyVnlJSEpoYm1Ob1pYSXdIaGNOTWpBdwpOakEwTWpBME1qVXdXaGNOTXpBd05qQXpNRGcwTWpVd1dqQThNUXN3Q1FZRFZRUUdFd0pWVXpFV01CUUdBMVVFCkNnd05RMmx6WTI4Z1UzbHpkR1Z0Y3pFVk1CTUdBMVVFQXd3TVZYTmxjaUJ5WVc1amFHVnlNSUdmTUEwR0NTcUcKU0liM0RRRUJBUVVBQTRHTkFEQ0JpUUtCZ1FDK3IzdzgySFAvZ05ueEJQNVhHRGFiZWNvanNvbHpQaDFuTGYwVQp2dEQ0eWJZMUY4cmRRSlN6d1BOR0tzbWloRVpSdUhQRFJXb1I3MXd3MWdtY2Q2SDVkTUxCckx0OWV1MVhvV0Z5ClB3S0l1aTdqWERpNnE4UTJzKzFCYm1QNUowN013U25uUEg4ZExoZkdORmJLVDMvNXBlOEZGZFFmSmhsbGEzVU4KZU8xM29RSURBUUFCTUEwR0NTcUdTSWIzRFFFQkJRVUFBNEdCQUFCV0hQYnVpa1Ivb0hjYm1kUTVDQk1zajRtawpvWGFmcGhmVTI3cWtYdTlGcFUwM0VDQ2Y0YzFpM3FoeEkzaU1iRmdOMkw3ZU1IS1ZEd29jRHlQNGN6YmlEWkpuCjlseHJ5OVNtaEdUbGxDRCthcnpwTjN2UkZxc2lHeGdVYXg4bVR3RWtCWDNlZno4cE1CZ09WQ1F2c2daRzZQejEKVE5FdFRHK3V5QkRMV2RoYQotLS0tLUVORCBDRVJUSUZJQ0FURS0tLS0tCg=="
    apic_refresh_time: 1200
    encap_type: "vxlan"
    mcast_range_start: "225.38.1.1"
    mcast_range_end: "225.38.255.255"
    aep: "hypf-aep"
    vrf_name: "k8s28_vrf"
    vrf_tenant: "common"
    l3out: "k8s28"
    node_subnet: "192.168.28.1/24"
    l3out_external_networks: ["\"k8s28_net\""]
    extern_dynamic: "10.3.0.1/24"
    extern_static: "10.4.0.1/24"
    node_svc_subnet: "10.5.0.1/24"
    kube_api_vlan: 55
    service_vlan: 56
    infra_vlan: 3901
  options: {}
  mtu: 1600
  node_selector: {}
  update_strategy: null

cluster/network.go
```diff
@@ -284,9 +340,98 @@ func (c *Cluster) doWeaveDeploy(ctx context.Context, data map[string]interface{}
 	return c.doAddonDeploy(ctx, pluginYaml, NetworkPluginResourceName, true)
 }
 
+func (c *Cluster) doAciDeploy(ctx context.Context, data map[string]interface{}) error {
+	refreshTime, err := atoiWithDefault(c.Network.AciNetworkProvider.ApicRefreshTime, DefaultApicRefreshTime)
```
superseb, line 344: What is the reason this (and the one below) is not in `cluster/defaults.go`? This currently panics when not provided. I think the whole part for defaults and validation should go into `cluster/defaults.go` and `cluster/validation.go`.

shastrinator: Thanks for getting back. I tried this just now that you mentioned, It doesn't panic when apic_refresh_time is not given. This line is there for that exact reason. I edited the cluster.yaml I am using now exactly in the post above.
I have validated the options that are required for ACI and these don't apply to other CNIs which don't have the physical fabric involved. Keeping them common will not help other CNIs. We use the other common network options like MTU and pod subnet that are required by all CNIs. We can discuss more about this offline over email if you prefer

cluster/defaults.go
superseb: Although I couldn't find a hard technical requirement for the `networkPluginConfigDefaultsMap`, I don't want to divert from the use for one plugin, every other plugin uses it so I would like to use the same mechanism for all plugins.

cluster/network.go
superseb: I ran this PR with just specifying the network plugin `ACI` and nothing else and it panics on this line.

The problem here is that if any of these defaults error out, the cluster is provisioned but the network provisioning errors out while this can be easily caught by defaults/validation.

cluster/network.go
```diff
@@ -70,6 +71,9 @@ const (
 
 	WeaveNetworkPlugin  = "weave"
 	WeaveNetworkAppName = "weave-net"
+
+	AciNetworkPlugin = "ACI"
```
superseb, line 75: All the network plugins are lowercase and the convention in the code seems to be `Aci` which is also conform the use of `aci` here.

cluster/network.go
shastrinator: Ok. For ACI, not all arguments have defaults and defaults cannot be provided for all arguments because they pertain to specific values like the address of the ACI controller, credentials for it, other subnets. Running without any arguments is bound to fail althought not at this argument. We have a tool called acc-provision that generates this portion of cluster.yml, so customers dont have to do this by hand. https://github.com/noironetworks/acc-provision.

shastrinator: Ok.. I will move these to cluster/defaults.go as you mentioned and submit a PR shortly.

cluster/network.go
shastrinator: ACI is an acronym(Application Centric Infrastructure), so it is capitalized. I will change the rest of the arguments to use the capitalized convention as well, hope this is ok..

shastrinator: This is now changed back to "aci"

shastrinator:
Hi Seb, I have updated the CNI template with more parameters and added validation and am running tests at my end. Can you please continue your review?

shastrinator:
I had to do some more iterations as I brought in a lot of template changes and there were quite a few tab and metadata issues in the template. Tests are now looking good at my end and I am done with my changes. Please continue with your review.

superseb:
Thanks! It seems that there might be something wrong with image population in the network deploy YAML, did you deploy successfully with the last changes?

I will do a final review soon.

shastrinator:
> Thanks! It seems that there might be something wrong with image population in the network deploy YAML, did you deploy successfully with the last changes?
> 
> I will do a final review soon.
Thanks. I did not face any issues with image population and was able to deploy. What was the issue you ran into?  I found one more issue with the template from my tests and I will push that update.

shastrinator:
Done at this point. Do let me know what issue you faced. Thanks.

cluster/validation.go
```diff
@@ -67,13 +67,59 @@ func validateAuthOptions(c *Cluster) error {
...
 func validateNetworkOptions(c *Cluster) error {
-	if c.Network.Plugin != NoNetworkPlugin && c.Network.Plugin != FlannelNetworkPlugin && c.Network.Plugin != CalicoNetworkPlugin && c.Network.Plugin != CanalNetworkPlugin && c.Network.Plugin != WeaveNetworkPlugin {
+	if c.Network.Plugin != NoNetworkPlugin && c.Network.Plugin != FlannelNetworkPlugin && c.Network.Plugin != CalicoNetworkPlugin && c.Network.Plugin != CanalNetworkPlugin && c.Network.Plugin != WeaveNetworkPlugin && c.Network.Plugin != AciNetworkPlugin {
 		return fmt.Errorf("Network plugin [%s] is not supported", c.Network.Plugin)
 	}
 	if c.Network.Plugin == FlannelNetworkPlugin && c.Network.MTU != 0 {
 		return fmt.Errorf("Network plugin [%s] does not support configuring MTU", FlannelNetworkPlugin)
 	}
+	if c.Network.Plugin == AciNetworkPlugin {
+		//Skipping Cloud options for now
```
superseb, line 92: Does `for now` mean in the end commit or is this subject to change? if this is going to be committed, it needs a note why this is skipped and what the plan is

shastrinator: I have to do more testing with this template in cloud environment before I add validation. I will make another commit when I test cloud version.

cluster/validation.go
```diff
@@ -67,13 +67,59 @@ func validateAuthOptions(c *Cluster) error {
...
+			AciNodeSubnet, AciAEP, AciVRFName, AciVRFTenant, AciL3Out, AciDynamicExternalSubnet,
+			AciStaticExternalSubnet, AciServiceGraphSubnet, AciKubeAPIVlan, AciServiceVlan, AciInfraVlan,
+			AciNodeSubnet, AciMcastRangeStart, AciMcastRangeEnd}
+		for _, v := range networkOptionsList {
+			val, ok := c.Network.Options[v]
+			if !ok || val == "" {
+				v = transformAciNetworkOption(v)
+				return fmt.Errorf("Network plugin aci: %s under aci_network_provider is not provided", strings.TrimPrefix(v, "aci_"))
+			}
+		}
```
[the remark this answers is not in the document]
shastrinator: Added description for options.

shastrinator:
Hi Sebastiaan, It seems that with the version v1.18.6-rancher1-3, kubelet is started as a pod in the etcd bring up stage and it seems to require CNI to be loaded, network provider is applied at a later stage. How is this supposed to work- I see that this version does not come up locally with/with-out my changes.

cluster/validation.go
shastrinator: As discussed, added validation to disable cloud options. Please continue your review

shastrinator:
Split this PR into 2:
Part1: #2339

PR #2160, opened 2020-07-15
[v1.0] update kdm for v1.16.13 and v1.17.9

kinarashah:


rancherbot (approved):

PR #2161, opened 2020-07-15
[v1.1] update kdm for v1.16.13, v1.17.9, v1.18.6

kinarashah:


rancherbot (approved):

PR #2172, opened 2020-07-21
Add win_ Params for prefix path, env, args and binds

luthermonson:
Problem: When building a hybrid cluster with windows nodes there is only a single set of overrides you can use per service. This limits configuring the node as service args and prefix_path sometimes need to be specific for the different OS.
Solution: Add support for `win_` prefixed parameters for cluster level `path_prefix` and service level `extra_args`, `extra_env` and `extra_binds`. Params will work as before, passing in the non `win_` prefixed params, IF you set the `win_` prefixed params it willy only use those meaning you will need to duplicate the params in both config sections of your rke cluster yaml.


luthermonson:
https://github.com/rancher/rancher/issues/25108

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.13
 
 replace (
 	github.com/knative/pkg => github.com/rancher/pkg v0.0.0-20190514055449-b30ab9de040e
+	github.com/rancher/types => github.com/luthermonson/types v0.0.0-20200721015930-23162793857e
```
ryansann, line 7: Was this used for testing? Probably want to remove it before merging

luthermonson: yup, it's updated now

luthermonson:
@superseb yup exactly

PR #2186, opened 2020-08-03
Add restore flag to use local state

superseb:
https://github.com/rancher/rke/issues/1336

This adds the flag `--use-local-state` for the `rke etcd restore-snapshot` command to use the local statefile and not check the snapshot archive for an included statefile


superseb:
Sorry for re-request, made some changes:

* Added logging for when the flag is used so its clear to the user
* Don't swallow the error on reading the state file if the new flag was used to explicitly use the local state file

PR #2206, opened 2020-08-18
add filepath.Clean to prefix path

luthermonson:


luthermonson:
https://github.com/rancher/rancher/issues/28440

PR #2208, opened 2020-08-19
added windows path cleaner

luthermonson:
Since filepath.Clean() is being called in the rancher master it'll always be in Linux, had to write a quick cleaner to use.


kinarashah (dismissed):

ryansann (dismissed):

luthermonson:
https://github.com/rancher/rancher/issues/28440

Issue #2194, opened 2020-08-10
/lib/modules not mounting read-only on Fedora CoreOS


kschamplin:
**RKE version:**
v1.1.4
**Docker version: (`docker version`,`docker info` preferred)**
```
[core@squirtle ~]$ docker info
Client:
 Debug Mode: false

Server:
 Containers: 10
  Running: 7
  Paused: 0
  Stopped: 3
 Images: 4
 Server Version: 19.03.11
 Storage Driver: overlay2
  Backing Filesystem: xfs
  Supports d_type: true
  Native Overlay Diff: true
 Logging Driver: journald
 Cgroup Driver: systemd
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: runc
 Default Runtime: runc
 Init Binary: /usr/libexec/docker/docker-init
 containerd version: 
 runc version: fbdbaf85ecbc0e077f336c03062710435607dbf1
 init version: 
 Security Options:
  seccomp
   Profile: default
  selinux
 Kernel Version: 5.7.8-200.fc32.x86_64
 Operating System: Fedora CoreOS 32.20200715.3.0
 OSType: linux
 Architecture: x86_64
 CPUs: 4
 Total Memory: 15.63GiB
 Name: squirtle
 ID: T6WG:NF2V:6Q4R:YTML:JQKA:ZQE6:JNBJ:MQG3:VVO3:FAFM:JXRB:2ZCE
 Docker Root Dir: /var/lib/docker
 Debug Mode: false
 Registry: https://index.docker.io/v1/
 Labels:
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: true
```
**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)**
```
[core@squirtle ~]$ uname -r
5.7.8-200.fc32.x86_64
```
```
[core@squirtle ~]$ cat /etc/os-release
NAME=Fedora
VERSION="32.20200715.3.0 (CoreOS)"
ID=fedora
VERSION_ID=32
VERSION_CODENAME=""
PLATFORM_ID="platform:f32"
PRETTY_NAME="Fedora CoreOS 32.20200715.3.0"
ANSI_COLOR="0;34"
LOGO=fedora-logo-icon
[...]

PR #2214, opened 2020-08-21
cluster/plan: don't relabel /lib/modules by default

vbatts:
https://github.com/rancher/rke/issues/2194

As this logic went, it would relabel /lib/modules, except on enterprise
linux and when SELinux is enabled (even just permisive).

Flatcar Container Linux defaults to SELinux on, but permisive, and
`/lib/modules/` is a symlink to the read-only `/usr`.
So `./rke up` would fail on attempting to relabel /usr.

The prior work around is to set `SELINUX=disable` in
/etc/selinux/config.

Signed-off-by: Vincent Batts <vbatts@kinvolk.io>

### DIFF
--- a/cluster/plan.go
+++ b/cluster/plan.go
@@ -602,11 +602,7 @@ func (c *Cluster) BuildKubeProxyProcess(host *hosts.Host, serviceOptions v3.Kube
 			"/run:/run",
 		}
 
-		BindModules := "/lib/modules:/lib/modules:z,ro"
-		if hosts.IsEnterpriseLinuxHost(host) && hosts.IsDockerSELinuxEnabled(host) && !hosts.IsEnterpriseLinuxDocker(host) {
-			// Avoid relabing on Enterprise Linux with Docker SELinux and upstream Docker
-			BindModules = "/lib/modules:/lib/modules:ro"
-		}
+		BindModules := "/lib/modules:/lib/modules:ro"
 		Binds = append(Binds, BindModules)
 	}
 	Binds = append(Binds, host.GetExtraBinds(kubeproxy.BaseService)...)


ibuildthecloud:
@superseb @galal-hussein I can't understand a use case where we would ever need to relabel /lib/modules.  It seems to be on the platform where we support SELinux (RHEL) we specifically don't relabel. This logic seems to be correct in that we just never relabel /lib/modules.  I'd say as long as this doesn't break a default centos 7/8 install it should be fine.

superseb:
This was added in https://github.com/rancher/rke/pull/1724 and seems like a copy from the other bind. The reason the exception was added was because we want to avoid changes on cluster provisioning if nothing changed. This change will replace `kube-proxy` on every node besides the exception that is present today. I'm checking if there was technical requirement for the relabing in the original PR.

superseb (approved):

dirien:
hi @vbatts  and @superseb:

i get following error with flatcar linux:

`
[controlPlane] Failed to upgrade Control Plane: [[Failed to start [kube-proxy] container on host [xxx]: Error response from daemon: error setting label on mount source '/usr/lib64/modules': relabeling content in /usr is not allowed]]
`

your fix was about /usr/lib i have /usr/lib64. Is there a different fix needed?

vbatts:
@dirien ah, i just saw this message while looking at the other issue. Are you getting this error with this PR?

mikekuzak:
@vbatts I'm getting the same issue with Rancher 2.3.9 and 2.4.5.  I tried to run master-head to see if it fixes the issue but rancher is not even starting, getting different error ...

```
2020/10/05 22:59:59 [ERROR] error syncing 'system-default-registry': handler copy-settings: the server could not find the requested resource, requeuing
2020/10/05 22:59:59 [INFO] Deleting roleBinding clusterrolebinding-5bctc
2020/10/05 22:59:59 [INFO] Deleting roleBinding clusterrolebinding-cpfpb
2020/10/05 22:59:59 [INFO] Deleting roleBinding clusterrolebinding-cpfpb
E1005 22:59:59.868889       7 runtime.go:78] Observed a panic: "assignment to entry in nil map" (assignment to entry in nil map)
goroutine 5750 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x3c139e0, 0x4d2e320)
        /go/pkg/mod/github.com/rancher/apimachinery@v0.19.0-rancher1/pkg/util/runtime/runtime.go:74 +0xa3
```

dirien:
@vbatts  same as @mikekuzak. I have v2.4.8 running.

hwaastad:
> @vbatts same as @mikekuzak. I have v2.4.8 running.
Hi, I have seen the same issue on my flatcar instances. After some troubleshooting, i’ve downgraded to 2512.1.0 at cluster is OK.
On 2605.6.0 the labeling issue occured.
I’ll continue to trace down the exact version that breaks.

PR #2245, opened 2020-09-22
Don't advertise etcd port 4001 in v3.4.13 and up

superseb:
https://github.com/rancher/rke/issues/2123

This removes the unneeded port 4001 to be advertised starting with k8s 1.19 (to avoid having every etcd to be rolled when not updating to k8s 1.19)

### DIFF
--- a/cluster/plan.go
+++ b/cluster/plan.go
@@ -49,8 +49,9 @@ const (
 	KubeletDockerConfigPath    = "/var/lib/kubelet/config.json"
 
 	// MaxEtcdOldEnvVersion The versions are maxed out for minor versions because -rancher1 suffix will cause semver to think its older, example: v1.15.0 > v1.15.0-rancher1
-	MaxEtcdOldEnvVersion = "v3.2.99"
-	MaxK8s115Version     = "v1.15"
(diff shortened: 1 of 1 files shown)


ryansann (approved):

cluster/plan.go
```diff
@@ -49,8 +49,9 @@ const (
 	KubeletDockerConfigPath    = "/var/lib/kubelet/config.json"
 
 	// MaxEtcdOldEnvVersion The versions are maxed out for minor versions because -rancher1 suffix will cause semver to think its older, example: v1.15.0 > v1.15.0-rancher1
-	MaxEtcdOldEnvVersion = "v3.2.99"
-	MaxK8s115Version     = "v1.15"
+	MaxEtcdOldEnvVersion   = "v3.2.99"
+	MaxK8s115Version       = "v1.15"
+	MaxEtcdPort4001Version = "v3.4.3-rancher99"
```
kinarashah, line 54: should this be 3.4.13?

superseb: This is to handle the `-rancher` suffix which leads to `v3.14.3-rancher1` being marked as lower than `v3.14.3`

kinarashah (approved):

PR #2302, opened 2020-10-26
Updates to install Calico with the Operator

tmjd:
This is the initial work to update RKE to install Calico using the Tigera Operator. Once this is in I'll follow up with a PR to rancher/kontainer-driver-metadata.

### DIFF
--- a/cluster/defaults.go
+++ b/cluster/defaults.go
@@ -431,14 +431,22 @@ func newDefaultAdmissionConfiguration() (*apiserverv1alpha1.AdmissionConfigurati
 	return admissionConfiguration, nil
 }
 
-func (c *Cluster) setClusterImageDefaults() error {
-	var privRegURL string
+func (c *Cluster) getDefaultPrivateRegistryURL() string {
+	for _, privReg := range c.PrivateRegistries {
+		if privReg.IsDefault {
+			return privReg.URL
+		}
+	}
+	return ""
+}
 
+func (c *Cluster) setClusterImageDefaults() error {
 	imageDefaults, ok := metadata.K8sVersionToRKESystemImages[c.Version]
 	if !ok {
 		return nil
 	}
 
+	privRegURL := c.getDefaultPrivateRegistryURL()
 	for _, privReg := range c.PrivateRegistries {
 		if privReg.IsDefault {
 			privRegURL = privReg.URL
@@ -466,6 +474,7 @@ func (c *Cluster) setClusterImageDefaults() error {
 		&c.SystemImages.CalicoCtl:                 d(imageDefaults.CalicoCtl, privRegURL),
 		&c.SystemImages.CalicoControllers:         d(imageDefaults.CalicoControllers, privRegURL),
 		&c.SystemImages.CalicoFlexVol:             d(imageDefaults.CalicoFlexVol, privRegURL),
+		&c.SystemImages.TigeraOperator:            d(imageDefaults.TigeraOperator, privRegURL),
 		&c.SystemImages.CanalNode:                 d(imageDefaults.CanalNode, privRegURL),
 		&c.SystemImages.CanalCNI:                  d(imageDefaults.CanalCNI, privRegURL),
 		&c.SystemImages.CanalControllers:          d(imageDefaults.CanalControllers, privRegURL),
--- a/cluster/defaults.go
+++ b/cluster/defaults.go
@@ -447,12 +447,6 @@ func (c *Cluster) setClusterImageDefaults() error {
 	}
 
 	privRegURL := c.getDefaultPrivateRegistryURL()
-	for _, privReg := range c.PrivateRegistries {
-		if privReg.IsDefault {
-			privRegURL = privReg.URL
-			break
-		}
-	}
 	systemImagesDefaultsMap := map[*string]string{
 		&c.SystemImages.Alpine:                    d(imageDefaults.Alpine, privRegURL),
 		&c.SystemImages.NginxProxy:                d(imageDefaults.NginxProxy, privRegURL),
--- a/cluster/defaults.go
+++ b/cluster/defaults.go
@@ -67,6 +67,8 @@ const (
 	DefaultFlannelBackendVxLanPort = "8472"
 	DefaultFlannelBackendVxLanVNI  = "1"
 
+	DefaultCalicoInstallMethod          = "operator"
+	K8sVersionCalicoInstallMethod       = "1.19.3"
 	DefaultCalicoFlexVolPluginDirectory = "/usr/libexec/kubernetes/kubelet-plugins/volume/exec/nodeagent~uds"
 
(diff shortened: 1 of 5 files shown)


cluster/defaults.go
```diff
@@ -431,14 +431,22 @@ func newDefaultAdmissionConfiguration() (*apiserverv1alpha1.AdmissionConfigurati
 	return admissionConfiguration, nil
 }
 
-func (c *Cluster) setClusterImageDefaults() error {
-	var privRegURL string
+func (c *Cluster) getDefaultPrivateRegistryURL() string {
```
Oats87, line 434: What is the purpose for adding this in its own independent function, given that the original for loop is still in place?

cluster/network.go
```diff
@@ -204,12 +208,15 @@ func (c *Cluster) doCalicoDeploy(ctx context.Context, data map[string]interface{
 		FlexVolImg:       c.SystemImages.CalicoFlexVol,
 		RBACConfig:       c.Authorization.Mode,
 		NodeSelector:     c.Network.NodeSelector,
+		OperatorImage:    c.SystemImages.TigeraOperator,
+		Kubectl:          c.Services.KubeAPI.Image, // Use same image that is used in GetAddonsExecuteJob for doAddonDeploy
```
Oats87, line 212: This will end up being the `rancher/hyperkube` image. Is this desirable or would a more streamlined image make sense?

cluster/validation.go
```diff
@@ -316,17 +316,22 @@ func validateNetworkImages(c *Cluster) error {
...
-		}
-		if len(c.SystemImages.CalicoNode) == 0 {
-			return errors.New("calico image is not populated")
-		}
-		if len(c.SystemImages.CalicoControllers) == 0 {
-			return errors.New("calico controllers image is not populated")
+		// If TigeraOperator is specified then the versions used will be
+		// bundled with the Operator, so only check the Calico images when
+		// the Operator is not used.
+		if len(c.SystemImages.TigeraOperator) == 0 {
```
Oats87, line 322: There may be times when you would want to deploy Calico in the "legacy" manner rather than using the operator, but if the corresponding K8s version has the operator image defined then this sanity check may miss image misconfigurations.

Oats87 (changes requested):

cluster/defaults.go
tmjd: Thank you. I meant to remove the original loop. The reason I put it into a function was for use in cluster/network.go, which I see now that I messed up. I think I forgot to run through my manual test after making this change.

cluster/network.go
tmjd: It will be the hyperkube image. The deploying by `rke-network-plugin-deploy-job` uses that image specifically calling `kubectl` in it which is exactly what it is being used for for Calico.
If there was a Kubectl image that was a better fit here we should use that.

An alternative to using Kubectl would be if there was a multiple stage deployment option that would allow 1st applying a CRD and then creation of an instance of the CR. Using Kubectl is only necessary to stage creation of the CR so it happens after the CRD is created.

Oats87: Is the goal to have `rke` itself create the CRD via the `rke-network-plugin-deploy-job` then down the line use the `kubectl` image to create the necessary CR's for the operator?

cluster/validation.go
tmjd: Sorry if I've misunderstood how things work with RKE but is it possible to install with an 'older' set of manifests but using a 'newer' set of images?
I thought the images and templates were both driven off the k8s version, is that not correct?
Something I should clarify is that when using the TigeraOperator there is no need for the other Calico component images to be specified as they are bundled in the operator. 
Do you have a suggestion on what this check should look like? Basically for Calico installs using the operator, only the operator is needed (none of the Calico images) and older installs only need the Calico images.

Oats87: I'm not sure that forcing all users of the Calico network plugin to use the operator past a certain K8s version makes sense, given that we have a perfectly suitable mechanism to install Calico as it is. Generally, when a hard change is introduced like this into RKE, we provide the option to use a method that is already around. `kube-dns` to `CoreDNS` is a great example of this.

luthermonson:
does the operator support windows nodes? if this gets installed will we have to figure out how to make that with with rke-tools? we currently do everything with wins + flannel in powershell scripts during bootstrap, have you tested any of that @tmjd ?

cluster/network.go
tmjd: The goal is to create a CRD and an instance of the CR.
With the CRD and the CR in the manifest, the `rke-network-plugin-deploy-job` will fail on the first attempt of the CR (because the CRD was not previously applied) and then pass on the 2nd attempt of the job. (So I guess that is another option.)
So I was trying to achieve applying the CRD and CR without any errors by having in the manifest the CRD and a kubectl pod that creates the CR.
If you want to see the manifest I've been testing check out https://github.com/rancher/kontainer-driver-metadata/pull/394

tmjd:
@luthermonson What do you mean by support windows nodes? Just like the previous Calico install method it does not attempt to run the calico-node deamonset on windows nodes through the use of a nodeSelector. Here I am only changing the Calico install (not Canal) so I don't think "wins + flannel" would be a consideration for this PR.
Do you have some support aspect that you are specifically wondering about?

luthermonson:
@tmjd my understanding was calico now supports windows. wasnt sure how that was done and if that was taken into account here. sounds like it's not. the issue is right now when you create a cluster with rke we only support flanneld if you want a cluster with windows nodes. i saw a PR adding calico operator (i dont honestly know what it does) and was just checking if this included any consideration so we could allow windows clusters to use calico too. reason im asking, as i know it would take some work in the UI and rke tools to make sure it works 100% so i was going to offer some guidance if you wanted to make it work.

tmjd:
@luthermonson, you are correct that Calico now supports windows but, IIUC, due to the limitations of how Calico must be ran as a service on windows nodes we cannot just bring in Windows support with a new daemonset or with the operator. Calico must be ran as a service on the windows hosts. Thank you for your offer of guidance, I'll ping someone that is working on on windows support.
@lmm I believe you are looking into this so you may want to chat with @luthermonson .

cluster/validation.go
tmjd: I've made updates to allow sticking with the "Legacy" Calico manifest install method if desired.

Oats87 (commented):
@tmjd What happens when a user is running `1.19.2` (or really any earlier version of K8s) and upgrades to `1.19.3` and receives the new Calico installation method i.e. operator? Does the operator take over management of the daemonset from that point on?

tmjd:
The operator will orchestrate a rolling update from the previous components to the new components managed by the operator. It must orchestrate the rolling update because the components are moved to a new namespace. After the components are migrated it will remove the previous daemonset and deployment.

tmjd:
Are there further concerns about this PR or more updates needed?

PR #2323, opened 2020-11-09
Update KDM - add v.16.15-rancher1-2

mrajashree:
This change was generated after running
```
RANCHER_METADATA_URL=https://releases.rancher.com/kontainer-driver-metadata/release-v2.5/data.json go generate
```

PR #2324, opened 2020-11-09
Update KDM - add v.16.15-rancher1-2

mrajashree:
Forward port of https://github.com/rancher/rke/pull/2323


rancherbot (approved):

PR #2346, opened 2020-11-20
Set default http backend to be optional

nickgerace:
Set default http backend to be optional for ingress nginx. It will be enabled by default.

Relevant Issue: https://github.com/rancher/rancher/issues/25590
Relevant Comment: https://github.com/rancher/rancher/issues/25590#issuecomment-729248564
KDM Data PR: https://github.com/rancher/kontainer-driver-metadata/pull/425

### DIFF
--- a/cluster/addons.go
+++ b/cluster/addons.go
@@ -37,7 +37,6 @@ const (
 	UserAddonJobName               = "rke-user-addon-deploy-job"
 	UserAddonIncludeJobName        = "rke-user-includes-addons-deploy-job"
 	MetricsServerAddonResourceName = "rke-metrics-addon"
-	NginxIngressAddonAppName       = "ingress-nginx"
 	KubeDNSAddonAppName            = "kube-dns"
 	KubeDNSAutoscalerAppName       = "kube-dns-autoscaler"
 	CoreDNSAutoscalerAppName       = "coredns-autoscaler"
@@ -47,6 +46,10 @@ const (
 	CoreDNSProvider = "coredns"
 	KubeDNSProvider = "kube-dns"
 	Nodelocal       = "nodelocal"
+
+	NginxIngressAddonAppName                 = "ingress-nginx"
+	NginxIngressAddonDefaultBackendName      = "default-http-backend"
+	NginxIngressAddonDefaultBackendNamespace = "ingress-nginx"
 )
 
 var DNSProviders = []string{KubeDNSProvider, CoreDNSProvider}
@@ -66,6 +69,7 @@ type ingressOptions struct {
 	HTTPPort          int
 	HTTPSPort         int
 	NetworkMode       string
+	DefaultBackend    bool
 	UpdateStrategy    *appsv1.DaemonSetUpdateStrategy
 	Tolerations       []v1.Toleration
 }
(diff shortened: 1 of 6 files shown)


brendarearden (dismissed):
LGTM

superseb (changes requested):
The issue comment with the design shows `default-backend` as option, but I assume it will be `default_backend` based on this code?

I think we need to update the ingress config in `types/rke_types.go` as well.

Can we confirm that the behavior will be on disable, that the user needs to manually remove the deployment? (on existing setups)

nickgerace:
@superseb Thanks for the review!

> The issue comment with the design shows default-backend as option, but I assume it will be default_backend based on this code?
Correct. The other fields use underscores rather than dashes, so we are using an underscore to maintain consistency.

> I think we need to update the ingress config in types/rke_types.go as well.
Perhaps I am missing something, but I believe it has been updated: https://github.com/rancher/rke/pull/2346/files#diff-d6d39102a5c2432600e3ffac3abe79885d842926704befe5bdb3063be71518b7R443

> Can we confirm that the behavior will be on disable, that the user needs to manually remove the deployment? (on existing setups)
This is a good point. I am re-testing this workflow, and will give you answer.

nickgerace:
@superseb (cc: @cbron) - confirmed that `default-http-backend` is still installed if the flag is set to `false` on an existing installation.

nickgerace:
Added re-install logic for ingress-nginx based on whether or not default_backend is set to false.

When upgrading from an `rke` release to the local build of this PR...
```bash
[nickgerace at rancherbook in ~] (0) 
% rke --version
rke version v1.2.3
 
[nickgerace at rancherbook in ~/Developer/rke/src/github.com/rancher/rke] (0) (master-ingress)
% KUBECONFIG=kube_config_cluster.yml kubectl get services -n ingress-nginx
NAME                   TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)   AGE
default-http-backend   ClusterIP   <REDACTED>     <none>        80/TCP    89s
[nickgerace at rancherbook in ~/Developer/rke/src/github.com/rancher/rke] (0) (master-ingress)
% KUBECONFIG=kube_config_cluster.yml kubectl get deployments -n ingress-nginx
NAME                   READY   UP-TO-DATE   AVAILABLE   AGE
default-http-backend   1/1     1            1           99s
 
[nickgerace at rancherbook in ~/Developer/rke/src/github.com/rancher/rke] (0) (master-ingress)
% ./rke --version
rke version dev
 
[nickgerace at rancherbook in ~/Developer/rke/src/github.com/rancher/rke] (130) (master-ingress)
% KUBECONFIG=kube_config_cluster.yml kubectl get deployments -n ingress-nginx
No resources found in ingress-nginx namespace.
[nickgerace at rancherbook in ~/Developer/rke/src/github.com/rancher/rke] (0) (master-ingress)
% KUBECONFIG=kube_config_cluster.yml kubectl get services -n ingress-nginx
No resources found in ingress-nginx namespace.
```

nickgerace:
We now only remove the default backend deployment and service after the daemonset has been updated. This is to ensure that we do not cause downtime and update according to the daemonset's update strategy.

Now, the default backend deployment is removed _after_ updating the ingress addOn.

superseb (approved):

Issue #2217, opened 2020-08-24
kube-apiserver is not restarted after changing audit log configuration


soon-leong:
**RKE version:** 
1.1.4

**Docker version: (`docker version`,`docker info` preferred)**
19.03.12

**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)**
Debian GNU/Linux 9 (stretch)
Kernel Version: 4.14.132

**Type/provider of hosts: (VirtualBox/Bare-metal/AWS/GCE/DO)**
VMware ESXi Server

**cluster.yml file:**
cluster_name: test
nodes:
- address: x.x.x.x
  user: root
  role:
    - controlplane
    - etcd
    - worker
  ssh_key: |-
    -----BEGIN RSA PRIVATE KEY-----
    -----END RSA PRIVATE KEY-----
services:
  kube-api:
    audit_log:
      enabled: true
      configuration:
        path: /var/log/kube-audit/audit-log.json
        format: json
        policy:
          apiVersion: audit.k8s.io/v1
          kind: Policy
          rules:
            - level: Metadata
              resources:
                - group: "" # core
                  resources: ["secrets", "configmaps"]
                - group: authentication.k8s.io
                  resources: ["tokenreviews"]
              omitStages:
                - "RequestReceived"

**Steps to Reproduce:**
1. Run rke up to create the cluster.
2. Run tail -f /var/log/kube-audit/audit-log.json to view the events in the audit log
3. Edit cluster.yml. Change the audit level from Metadata to None.
4. Run rke up.

**Results:**
When the audit level is changed to None, the audit events should not be written to the audit log but they continued to appear in the audit log.

soon-leong:
@superseb any estimate when the bug fix can be available? Thanks!

soon-leong:
Which version will the bug fix be available? Thanks!

Ellenqs:
My test can reproduce this issue. details as below

**RKE version:**
v1.2.3

**Docker version:** (docker version,docker info preferred)
19.03

**Type/provider of hosts:** (VirtualBox/Bare-metal/AWS/GCE/DO)
AWS EC2

**YAML file used to test:**
//
nodes:
  - address: xxxxxxxx
    user: ubuntu
    role: [controlplane, worker, etcd]
    ssh_key_path: ~/path/to/key
services:
    kube-api:
      audit_log:
        enabled: true
        configuration:
          path: /var/log/kube-audit/audit-log.json
          format: json
          policy:
[...]

PR #2376, opened 2020-12-10
Add auditlog checksum to trigger restart on update

superseb:
https://github.com/rancher/rke/issues/2217

This adds functionality similar to cloud config checksum, we add the current checksum as an environment variable, so if it changes, the logic triggers a restart of the kube-apiserver (Which doesn't happen currently so the new config is deployed but kube-apiserver is not restarted to reload the config)

### DIFF
--- a/cluster/plan.go
+++ b/cluster/plan.go
@@ -21,6 +21,7 @@ import (
 	v3 "github.com/rancher/rke/types"
 	"github.com/rancher/rke/util"
 	"github.com/sirupsen/logrus"
+	"sigs.k8s.io/yaml"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


cluster/plan.go
```diff
@@ -265,6 +267,11 @@ func (c *Cluster) BuildKubeAPIProcess(host *hosts.Host, serviceOptions v3.Kubern
 	}
 	if c.Services.KubeAPI.AuditLog != nil && c.Services.KubeAPI.AuditLog.Enabled {
 		Binds = append(Binds, fmt.Sprintf("%s:/var/log/kube-audit:z", path.Join(host.PrefixPath, "/var/log/kube-audit")))
+		bytes, _ := yaml.Marshal(c.Services.KubeAPI.AuditLog.Configuration.Policy)
```
StrongMonkey, line 270: Can we print or log the error in case of the malform input?

superseb: Added, thanks

StrongMonkey (approved):

kinarashah (approved):

PR #2389, opened 2020-12-22
[v1.2] init metadata only if metadataInitialized is false

kinarashah:
https://github.com/rancher/rancher/issues/30238

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -705,9 +705,11 @@ func InitClusterObject(ctx context.Context, rkeConfig *v3.RancherKubernetesEngin
 		},
 	}
 	if !metadata.MetadataInitialized {
+		logrus.Debugf("metadataInitialized: [False] [%s]", rkeConfig.ClusterName)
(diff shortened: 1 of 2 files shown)


superseb (approved):

cluster/cluster.go
```diff
@@ -705,9 +705,11 @@ func InitClusterObject(ctx context.Context, rkeConfig *v3.RancherKubernetesEngin
 		},
 	}
 	if !metadata.MetadataInitialized {
+		logrus.Debug("metadataInitialized: [False] [%s]", rkeConfig.ClusterName)
 		if err := metadata.InitMetadata(ctx); err != nil {
 			return nil, err
 		}
+		logrus.Debugf("metadataInitialized: [%v] [%s]", rkeConfig.ClusterName)
```
StrongMonkey, line 712: This is already called out in CI. You should two parameter.

StrongMonkey (dismissed):

StrongMonkey (approved):

PR #2418, opened 2021-01-14
[v1.1] Added etcd snapshot timeout parameter

rawmind0:
Issue https://github.com/rancher/rancher/issues/30663

Requires https://github.com/rancher/types/pull/1196 : `Timeout` field has been added to `Services.Etcd. BackupConfig`. Once types PR is merged, this PR should be updated to use it.

Required by https://github.com/rancher/rancher/pull/30885

This PR adds the feature to set etcd snapshot timeout instead of the [default](https://github.com/rancher/rke/blob/v1.1.6/docker/docker.go#L495). 

To limit the changes at the rke logic and functions parameters, the `Services.Etcd. BackupConfig.Timeout` field is read at `cluster. SnapshotEtcd` and `cluster.RestoreEtcdSnapshot` functions, using execution context to propagate the argument. `docker.WaitForContainer` function has been modified to set default `WaitTimeout=300` or timeout provided by context value `wait_timeout` if provided

PR #2419, opened 2021-01-15
[v1.2] Added etcd snapshot timeout parameter

rawmind0:
forward port to https://github.com/rancher/rke/pull/2418

Issue https://github.com/rancher/rancher/issues/30663

Required by https://github.com/rancher/rancher/pull/30886

### DIFF
--- a/cluster/defaults.go
+++ b/cluster/defaults.go
@@ -54,6 +54,7 @@ const (
 	DefaultMonitoringProvider            = "metrics-server"
 	DefaultEtcdBackupConfigIntervalHours = 12
 	DefaultEtcdBackupConfigRetention     = 6
+	DefaultEtcdBackupConfigTimeout       = docker.WaitTimeout
 
 	DefaultDNSProvider = "kube-dns"
 	K8sVersionCoreDNS  = "1.14.0"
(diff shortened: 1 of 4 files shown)


superseb (commented):
Is there a preference to wrap Context vs using the passed variable (`c.Services.Etcd`) which has all the variables available, and having a default on WaitForContainer and override when `c.Services.Etcd.BackupConfig.Timeout` is set?

rawmind0:
> Is there a preference to wrap Context vs using the passed variable (c.Services.Etcd) which has all the variables available, and having a default on WaitForContainer and override when c.Services.Etcd.BackupConfig.Timeout is set?

@superseb , the PR is proposing something similar to that. The new field `c.Services.Etcd.BackupConfig.Timeout` is read at [SnapshotEtcd](https://github.com/rancher/rke/pull/2419/files#diff-41a89aabd48470824c598071dec0d405ca88be491745ff5a6bb8c5257295b37aR21) and at [RestoreEtcdSnapshot](https://github.com/rancher/rke/pull/2419/files#diff-41a89aabd48470824c598071dec0d405ca88be491745ff5a6bb8c5257295b37aR183) setting default value if not set. The use of context instead is due to propagate the timeout to the `WaitForContainer` function minimizing the change on logic or function parameters, as this function is not related to cluster object. https://github.com/rancher/rke/pull/2418#issue-555204607

PR #2436, opened 2021-01-27
kdm bump for windows 2004 and 20h2

luthermonson:
https://github.com/rancher/rancher/issues/30701


rancherbot (approved):

PR #2446, opened 2021-02-01
[1.2] Go generate for adding k8s 1.20

StrongMonkey:


rancherbot (approved):

PR #2477, opened 2021-02-28
bump rke tools

luthermonson:
https://github.com/rancher/rancher/issues/31316


rancherbot (approved):

PR #2497, opened 2021-03-29
[v1.2] update kdm data

kinarashah:
https://github.com/rancher/rancher/issues/31782


rancherbot (approved):

PR #2514, opened 2021-04-13
Add omitempty to yaml tags & some struct fields in types/rke_types.go

dcermak:
This pull request adds additional tags to the structs in [`types/rke_types.go`](https://github.com/rancher/rke/blob/master/types/rke_types.go) so that a json/yaml schema can be automatically generated using the [alecthomas/jsonschema](https://github.com/alecthomas/jsonschema) package.

This requires the following changes:
- most of the `yaml:` tags were lacking the `omitempty` entry, whereas the `json:` tags have it.
- some additional fields that are optional were also marked as `omitempty`
- port numbers are frequently specified as numbers as well, for this the additional tag `jsonschema:"oneof_type=string;number"` is added so that the resulting schema will accept port numbers and not only port strings
- Add the type alias `NumberStringMap` for maps that allow numbers as values (in their yaml/json representation). This type is used to identify maps where this applies and is just used to generate the schema.


dcermak:
@nickgerace Anything that I can do to get this PR merged?

nickgerace:
@dcermak I'll reach out to RKE SMEs to take a look.

dcermak:
@nickgerace What can I do to get this eventually merged?

PR #2528, opened 2021-04-29
update to k8s v1.21 and golang 1.16

kinarashah:
https://github.com/rancher/rancher/issues/31567


rancherbot (approved):

StrongMonkey (approved):

Issue #2492, opened 2021-03-25
etcd-snapshot save without local state file waits for state file to get copied


superseb:
**RKE version:**
v1.2.6

**Steps to Reproduce:**
1. Create `cluster.yml`
2. Run `rke up`
3. Remove `cluster.rkestate`
4. Run `rke etcd-snapshot save`

**Results:**
See the process trying to copy the state file and waiting for the state file to be copied (around 20-30 seconds) and still successfully create the snapshot but waiting for the state file is unnecessary.

We probably need to error out without a local state file to indicate the state file is missing and add a flag to override it without local state (in the case we just need the etcd snapshot) and skip the copy state step to save time.

gz#14991

superseb:
This needs to be tested so that without a state file, the logging for waiting for state file to be copied is not seen.

And a regular snapshot-save with state file present should still work as normal.

anupama2501:
Reproduced on rke version v1.2.6
1. On an aws ec2 instance, install docker
2. Create a cluster.yml, in the yaml file input the IP address, internal address of the instance
3. Run rke up and once the process is finished successfully, run rke etcd snapshot-save

```
WARN[0000] Name of the snapshot is not specified using [rke_etcd_snapshot_2021-05-28T08:17:05-07:00]
INFO[0000] Starting saving snapshot on etcd hosts
INFO[0000] [dialer] Setup tunnel for host []
INFO[0000] Removing container [cluster-state-deployer] on host [], try #1
INFO[0001] [remove/cluster-state-deployer] Successfully removed container on host []
INFO[0001] [state] Deploying state file to [/etc/kubernetes/rke_etcd_snapshot_2021-05-28T08:17:05-07:00.rkestate] on host []
INFO[0001] Image [rancher/rke-tools:v0.1.72] exists on host []
INFO[0002] Starting container [cluster-state-deployer] on host [], try #1
INFO[0002] [state] Successfully started [cluster-state-deployer] container on host []
INFO[0003] Waiting for [cluster-state-deployer] container to exit on host []
INFO[0003] Container [cluster-state-deployer] is still running on host []: stderr: [], stdout: [Waiting for file [/etc/kubernetes/cluster.rkestate] to be successfully copied to this container, retry count 1
]
INFO[0004] Waiting for [cluster-state-deployer] container to exit on host []
[...]

PR #2537, opened 2021-05-15
Dont deploy statefile if its not readable

superseb:
https://github.com/rancher/rke/issues/2492

### DIFF
--- a/cluster/etcd.go
+++ b/cluster/etcd.go
@@ -68,6 +68,16 @@ func (c *Cluster) DeployRestoreCerts(ctx context.Context, clusterCerts map[strin
 }
 
 func (c *Cluster) DeployStateFile(ctx context.Context, stateFilePath, snapshotName string) error {
+	stateFileExists, err := util.IsFileExists(stateFilePath)
+	if err != nil {
+		logrus.Warnf("Could not read cluster state file from [%s], error: [%v]. Snapshot will be created without cluster state file. You can retrieve the cluster state file using 'rke util get-state-file'", stateFilePath, err)
+		return nil
+	}
+	if !stateFileExists {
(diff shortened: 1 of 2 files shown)


cluster/etcd.go
```diff
@@ -68,6 +68,16 @@ func (c *Cluster) DeployRestoreCerts(ctx context.Context, clusterCerts map[strin
 }
 
 func (c *Cluster) DeployStateFile(ctx context.Context, stateFilePath, snapshotName string) error {
+	stateFileExists, err := util.IsFileExists(stateFilePath)
```
StrongMonkey, line 71: Does the `DeployStateFile` only get called when creating etcd snapshot?

superseb: Yes, but in case of multiple calls this would be the same requirement as we can't deploy it if the state file doesnt exist/cant be read locally

StrongMonkey: Ok, I just want to make sure we are giving the right information. If it gets called elsewhere we have to change the warning message accrodingly.

cluster/etcd.go
```diff
@@ -68,6 +68,16 @@ func (c *Cluster) DeployRestoreCerts(ctx context.Context, clusterCerts map[strin
...
+	if !stateFileExists {
+		logrus.Warnf("Could not read cluster state file from [%s], file does not exist. Snapshot will be created without cluster state file. You can retrieve the cluster state file using 'rke util get-state-file'", stateFilePath)
+		return nil
+	}
+	if err != nil {
```
StrongMonkey, line 76: Should check the error first. The `util.IsFileExists` always return false, err so if there is an error it won't show up.

StrongMonkey (approved):

PR #2541, opened 2021-05-17
Do not rewrite SELinux labels on volume mounts

superseb:
Requires https://github.com/rancher/rancher-selinux/pull/6

### DIFF
--- a/cluster/plan.go
+++ b/cluster/plan.go
@@ -264,10 +264,10 @@ func (c *Cluster) BuildKubeAPIProcess(host *hosts.Host, serviceOptions v3.Kubern
 		services.SidekickContainerName,
 	}
 	Binds := []string{
-		fmt.Sprintf("%s:/etc/kubernetes:z", path.Join(host.PrefixPath, "/etc/kubernetes")),
+		fmt.Sprintf("%s:/etc/kubernetes", path.Join(host.PrefixPath, "/etc/kubernetes")),
 	}
 	if c.Services.KubeAPI.AuditLog != nil && c.Services.KubeAPI.AuditLog.Enabled {
-		Binds = append(Binds, fmt.Sprintf("%s:/var/log/kube-audit:z", path.Join(host.PrefixPath, "/var/log/kube-audit")))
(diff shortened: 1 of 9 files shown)


cluster/plan.go
```diff
@@ -479,11 +479,11 @@ func (c *Cluster) BuildKubeletProcess(host *hosts.Host, serviceOptions v3.Kubern
 		}
 	} else {
 		Binds = []string{
-			fmt.Sprintf("%s:/etc/kubernetes:z", path.Join(host.PrefixPath, "/etc/kubernetes")),
+			fmt.Sprintf("%s:/etc/kubernetes", path.Join(host.PrefixPath, "/etc/kubernetes")),
 			"/etc/cni:/etc/cni:rw,z",
 			"/opt/cni:/opt/cni:rw,z",
```
cmurphy, line 484: There are still a few volumes like these that are using the `z` mount option, can these be fixed too?

cmurphy (commented):
I don't know if the goal is to remove some of them or all of them but there's a few more sprinkled around:

https://github.com/rancher/rke/blob/462a7ba01ebb38d2f45e4a7b96e1c86c7a774b4f/hosts/hosts.go#L309
https://github.com/rancher/rke/blob/462a7ba01ebb38d2f45e4a7b96e1c86c7a774b4f/cluster/file-deployer.go#L63
https://github.com/rancher/rke/blob/462a7ba01ebb38d2f45e4a7b96e1c86c7a774b4f/pki/deploy.go#L89
https://github.com/rancher/rke/blob/462a7ba01ebb38d2f45e4a7b96e1c86c7a774b4f/pki/pki.go#L115
https://github.com/rancher/rke/blob/462a7ba01ebb38d2f45e4a7b96e1c86c7a774b4f/services/controlplane.go#L383
https://github.com/rancher/rke/blob/462a7ba01ebb38d2f45e4a7b96e1c86c7a774b4f/services/etcd.go#L350

cmurphy (commented):
lgtm :+1:

kinarashah (approved):
Changes LGTM, @superseb all clusters will start upgrading because of binds change, right? we'd need to call this out.

PR #2545, opened 2021-05-19
Update cert filename validation to *.pem

alexander-hughes:
Users leveraging custom certs directories face errors when deploying
if the directory contains any files that do not end in .pem. This
change adds additional validation to ensure files are *.pem before
attempting further logic.

Signed-off-by: Alexander Hughes <Alexander.Hughes@pm.me>

### DIFF
--- a/pki/util.go
+++ b/pki/util.go
@@ -592,7 +592,7 @@ func ReadCertsAndKeysFromDir(certDir string) (map[string]CertificatePKI, error) 
 
 	for _, file := range files {
 		logrus.Debugf("[certificates] reading file %s from directory [%s]", file.Name(), certDir)
-		if !strings.HasSuffix(file.Name(), "-key.pem") && !strings.HasSuffix(file.Name(), "-csr.pem") {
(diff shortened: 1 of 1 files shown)


superseb:
I changed this to filter out key/csr because they used to be silently ignored and the directory should just include the custom certificate files and nothing else was the determination. I assume your use case is that the custom certificate directory contains non certificate files?

alexander-hughes:
> I changed this to filter out key/csr because they used to be silently ignored and the directory should just include the custom certificate files and nothing else was the determination. I assume your use case is that the custom certificate directory contains non certificate files?

Correct, in a previous version of RKE these were also silently ignored. Attempting to upgrade to a newer version of RKE causes deployments to fail with this error.

superseb (approved):

superseb:
Please create an issue for it and link the PR to it so we can track changes

alexander-hughes:
> Please create an issue for it and link the PR to it so we can track changes

Done, thanks.

superseb:
Thanks!

PR #2562, opened 2021-06-03
update EncryptionConfig when using CustomCerts

chaudhryfaisal:
addresses https://github.com/rancher/rke/issues/2559


superseb (changes requested):
This seems to be copy pasted from `rebuildExistingState`, can't we modify/use that function?

chaudhryfaisal:
> This seems to be copy pasted from `rebuildExistingState`, can't we modify/use that function?

are you suggesting to move duplicate code into new function? ex:

```go

func updateEncryptionConfig(kubeCluster *Cluster, oldState *FullState, newState *FullState) error {
	if isEncryptionEnabled(&kubeCluster.RancherKubernetesEngineConfig) {
		if oldState.DesiredState.EncryptionConfig != "" {
			newState.DesiredState.EncryptionConfig = oldState.DesiredState.EncryptionConfig
		} else {
			var err error
			if newState.DesiredState.EncryptionConfig, err = kubeCluster.getEncryptionProviderFile(); err != nil {
				return err
			}
		}
	}
	return nil
}
func rebuildExistingState(ctx context.Context, kubeCluster *Cluster, oldState, newState *FullState, flags ExternalFlags) error {
	...
	err := updateEncryptionConfig(kubeCluster, oldState, newState)
	return err
}

func RebuildState(ctx context.Context, kubeCluster *Cluster, oldState *FullState, flags ExternalFlags) (*FullState, error) {
	...
	if flags.CustomCerts {
		...
		err = updateEncryptionConfig(kubeCluster, oldState, newState)
		if err != nil {
			return nil, err
		}
		return newState, nil
	}
	...
}
```

also any preferences with 

```go
		err = updateEncryptionConfig(kubeCluster, oldState, newState)
		if err != nil {
			return nil, err
		}
		return newState, nil
		
		vs 
		
		err = updateEncryptionConfig(kubeCluster, oldState, newState)
		return newState, err
```

superseb:
Because of the `if flags.CustomCerts`, it returns early and never reaches `rebuildExistingState` which seems to do the same as the code you are adding in this PR. If `rebuildExistingState` is doing too much or if that breaks custom certs, then we can't really do much about it besides maybe like you suggested, implement a function that we can reuse in the function and in the `if` block.

chaudhryfaisal:
do you have a suggestion how you would like this to be refactored?

superseb:
I checked out the other functions but they involve certificate related actions which are not needed in case of custom certificates, so moving the code block to a new function that can be reused is fine.

chaudhryfaisal:
github is not picking up new changes from my repo, could you delete this PR so I can I open new one with clean history?

superseb:
I can close it (as can you), you can push a different branch on your repo and create a PR from there. If you want to push updates to this PR, you need to force push to the branch on your repo. (which seems to be master but I would recommend for a next time to create a branch for your changes and PR from the branch)

PR #2565, opened 2021-06-07
Add cridockerd

superseb:
https://github.com/rancher/rancher/issues/31943

### DIFF
--- a/cluster/plan.go
+++ b/cluster/plan.go
@@ -57,6 +57,8 @@ const (
 	MaxEtcdNoStrictTLSVersion = "v3.4.14-rancher99"
 
 	EncryptionProviderConfigArgument = "encryption-provider-config"
+
+	KubeletCRIDockerdNameEnv = "RKE_KUBELET_CRIDOCKERD"
 )
 
 var admissionControlOptionNames = []string{"enable-admission-plugins", "admission-control"}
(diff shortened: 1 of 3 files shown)


paynejacob (dismissed):

cluster/validation.go
```diff
@@ -583,3 +591,24 @@ func validateIngressImages(c *Cluster) error {
...
+
+func validateCRIDockerdOption(c *Cluster) error {
+	if c.EnableCRIDockerd != nil && *c.EnableCRIDockerd {
+		k8sVersion := c.RancherKubernetesEngineConfig.Version
+		toMatch, err := semver.Make(k8sVersion[1:])
+		if err != nil {
+			return fmt.Errorf("%s is not valid semver", k8sVersion)
+		}
+		logrus.Debugf("Checking CRIDockerd for cluster version [%s]", k8sVersion)
+		// default backend will be false for k8s 1.21 and up
```
kinarashah, line 603: nit: update this for cri dockerd

cluster/validation.go
```diff
@@ -583,3 +591,24 @@ func validateIngressImages(c *Cluster) error {
 	}
 	return nil
 }
+
+func validateCRIDockerdOption(c *Cluster) error {
```
kinarashah, line 595: should we also set this as default for valid k8s version?

cluster/plan.go
```diff
@@ -1096,3 +1108,14 @@ func appendArgs(command []string, args map[string]string) []string {
 	}
 	return command
 }
+
+func (c *Cluster) IsCRIDockerdEnabled() bool {
+	if c == nil {
+		return false
+	}
+	if c.EnableCRIDockerd != nil && *c.EnableCRIDockerd {
+		logrus.Infof("CRIDockerd is enabled")
```
kinarashah, line 1117: could we move this on cluster level log? since we call this from kubelet process, it'll get printed in info for every call to getPlan per node which'll be pretty often

kinarashah (dismissed):

superseb:
Rebased after the golangci-lint PR

kinarashah (approved):

Issue #2571, opened 2021-06-10
cert generate-csr does not reflect changes to nodes


chaudhryfaisal:
**RKE version:**  v1.2.3
**Steps to Reproduce:**
- `cat single.yml`
```yaml
nodes:
  - address: 10.157.150.174
    role: [ controlplane,etcd,worker ]
    hostname_override: k8s-node-01
    user: root
services:
  kubelet:
    generate_serving_certificate: true
```
- `rke cert generate-csr --config single.yml --cert-dir certs`
```bash
INFO[0000] Running RKE version: v1.2.3                  
INFO[0000] Generating Kubernetes cluster CSR certificates 
INFO[0000] [certificates] Generating Kubernetes API server csr 
INFO[0000] [certificates] Generating Kube Controller csr 
INFO[0000] [certificates] Generating Kube Scheduler csr 
INFO[0000] [certificates] Generating Kube Proxy csr     
INFO[0000] [certificates] Generating Node csr and key   
INFO[0001] [certificates] Generating admin csr and kubeconfig 
INFO[0001] [certificates] Generating Kubernetes API server proxy client csr 
INFO[0001] [certificates] Generating etcd-10.157.150.174 csr and key 
INFO[0001] [certificates] Generating kube-kubelet-10-157-150-174 Kubernetes Kubelet csr 
INFO[0001] Successfully Deployed certificates at [certs]
```
- `ls -1 certs`
```bash 
kube-admin-csr.pem
kube-admin-key.pem
kube-apiserver-csr.pem
kube-apiserver-key.pem
kube-apiserver-proxy-client-csr.pem
kube-apiserver-proxy-client-key.pem
kube-controller-manager-csr.pem
kube-controller-manager-key.pem
kube-etcd-10-157-150-174-csr.pem
kube-etcd-10-157-150-174-key.pem
kube-kubelet-10-157-150-174-csr.pem
kube-kubelet-10-157-150-174-key.pem
kube-node-csr.pem
kube-node-key.pem
kube-proxy-csr.pem
kube-proxy-key.pem
kube-scheduler-csr.pem
kube-scheduler-key.pem
```
- `cat multi.yml`
```yaml
nodes:
  - address: 10.157.150.174
    role: [ controlplane,etcd,worker ]
    hostname_override: k8s-node-01
    user: root
  - address: 10.157.150.175
    role: [ controlplane,etcd,worker ]
    hostname_override: k8s-node-02
    user: root
    port: 2222
[...]

PR #2572, opened 2021-06-10
fix custom certs csr generation for node changes

chaudhryfaisal:
fix for https://github.com/rancher/rke/issues/2571

### DIFF
--- a/pki/services.go
+++ b/pki/services.go
@@ -419,9 +419,12 @@ func GenerateEtcdCSRs(ctx context.Context, certs map[string]CertificatePKI, rkeC
 	for _, host := range etcdHosts {
 		etcdName := GetCrtNameForHost(host, EtcdCertName)
 		etcdCrt := certs[etcdName].Certificate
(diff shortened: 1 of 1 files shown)


superseb:
@chaudhryfaisal Thanks for the PR, the problem here is that we return as soon as we find a non empty etcd host CSR while we are looping through all of them and thus missing everything after a non empty etcd host CSR, correct?

chaudhryfaisal:
> @chaudhryfaisal Thanks for the PR, the problem here is that we return as soon as we find a non empty etcd host CSR while we are looping through all of them and thus missing everything after a non empty etcd host CSR, correct?

there were two issues, one for etcd which you described above , second for Kubelet

1. for etcd: code was looking for non empty csr where it should be doing deep comparison incase new nodes were added
1. for Kubelet: code was returning after first node cert comparison where it should also be comparing remaining nodes for changes / addition of new nodes

superseb (approved):

superseb:
Thanks!

PR #2574, opened 2021-06-12
Improve validate

superseb:
* Add go commands to validate to make sure go modules are all correct
* Change kdm source to dev-v2.6 (the vendored data is already `dev-v2.6`)
* Remove `build/bin` directory (`build/bin` was being removed in build step, causing dirty repo in validation because of `buil/bin/.gitkeep`)
* Moved `validate` to first step in CI so it fails directly if validation fails


paynejacob (approved):

kinarashah (approved):

superseb:
KDM changed

scripts/ci
```diff
@@ -3,8 +3,8 @@ set -e
 
 cd $(dirname $0)
 
+./validate
```
paynejacob, line 6: <3

paynejacob (approved):

PR #2583, opened 2021-06-23
Test all CNI in CI

superseb:
This runs in parallel with current CI to save time (else every CNI needs to be tested sequentially). The "downside" of this is that we need to build the binary again.

This will make testing new CNI and upgrades better as we now test each CNI + upgrades. This came up after updating addons for k8s v1.21.

This also adds `kubectl` output to each cluster provisioning + upgrade so you can check what images have been used in the test (to verify versions).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,3 @@
-./bin
 ./.dapper
 ./dist
 ./.trash-cache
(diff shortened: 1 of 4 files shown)


ryansann (approved):
Looks good

PR #2588, opened 2021-06-25
go generate

superseb:
https://github.com/rancher/rancher/issues/32188

Includes a fix for CI runs being reported as successful while they actually failed.


rancherbot (dismissed):

kinarashah (approved):

PR #2591, opened 2021-06-26
Clean up Drone CI steps

superseb:
- Scope steps to events and instances so they don't run needlessly.
- Remove artifacts as they are not used

### DIFF
--- a/scripts/package
+++ b/scripts/package
@@ -5,59 +5,6 @@ source $(dirname $0)/version
 
 cd $(dirname $0)/..
 
-DIST=$(pwd)/dist/artifacts
-
-mkdir -p $DIST/${VERSION} $DIST/latest
-
-for i in build/bin/*; do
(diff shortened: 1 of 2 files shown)


scripts/package
```diff
@@ -5,59 +5,6 @@ source $(dirname $0)/version
 
 cd $(dirname $0)/..
 
-DIST=$(pwd)/dist/artifacts
```
paynejacob, line 8: isn't this useful for people testing airgap installs locally?

superseb: `dist/artifacts` contains compressed archives of the `rke` binary, and are not used (are not release assets on GitHub). What or who will be missing these artifacts?

paynejacob: I see, i thought this included the build.

paynejacob (approved):

PR #2595, opened 2021-07-01
Make it possible to configure dualstack

superseb:
https://github.com/rancher/rancher/issues/33107

### DIFF
--- a/cluster/validation.go
+++ b/cluster/validation.go
@@ -208,6 +208,31 @@ func validateNetworkOptions(c *Cluster) error {
 	if c.Network.Plugin == FlannelNetworkPlugin && c.Network.MTU != 0 {
 		return fmt.Errorf("Network plugin [%s] does not support configuring MTU", FlannelNetworkPlugin)
 	}
+	dualStack := false
+	serviceClusterRanges := strings.Split(c.Services.KubeAPI.ServiceClusterIPRange, ",")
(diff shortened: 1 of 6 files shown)


cluster/validation.go
```diff
@@ -208,6 +208,31 @@ func validateNetworkOptions(c *Cluster) error {
 	if c.Network.Plugin == FlannelNetworkPlugin && c.Network.MTU != 0 {
 		return fmt.Errorf("Network plugin [%s] does not support configuring MTU", FlannelNetworkPlugin)
 	}
+	dualStack := false
+	serviceClusterRanges := strings.Split(c.Services.KubeAPI.ServiceClusterIPRange, ",")
+	if len(serviceClusterRanges) > 1 {
+		logrus.Debugf("Found more than 1 service cluster IP range, assuming dual stack")
```
paynejacob, line 214: are there any side effects if this assumption is wrong?  Are there any other cases where a service could have multiple ips?

paynejacob (approved):

cluster/validation.go
superseb: These are configured as ranges (CIDRs), having multiple is only possible if you have dual stack (I wouldn't know how the cluster would decide what range (CIDR) to use when you create a service when its configured with multiple.

Issue #2597, opened 2021-07-02
RKE strips multiple certs from pem file when using intermediate/sub ca 


chaudhryfaisal:
RKE strips multiple certs from pem file when using intermediate/sub ca with sub ca and root ca in kube-ca.pem file

**RKE version:** 1.2.8

**cluster.yml file:**
```yaml
cluster_name: k8s-cluster-with-ca-chain
kubernetes_version: v1.19.7-rancher1-1
nodes:
  - address: 127.0.0.1
    role: [ controlplane,etcd,worker ]
    user: ubuntu
```
**Steps to Reproduce:**
1. generate CSRs: `rke cert generate-csr`
2. Sign CSRs with intermediate/sub ca 
3. initialize **cluster.rkestate**: `rke up -init --custom-certs`

**Results:**
1. check **kube-ca.pem**
```bash
# cat kube-ca.pem
-----BEGIN CERTIFICATE-----
MIIDqTCCApGgAwIBAgIUMiXn5PbZBgP9Oygd7Mx4MULILSIwDQYJKoZIhvcNAQEL
BQAwWjELMAkGA1UEBhMCVVMxCzAJBgNVBAgTAk5ZMQwwCgYDVQQHEwNMSUMxDDAK
BgNVBAoTA0s4UzEMMAoGA1UECxMDUktFMRQwEgYDVQQDEwtSS0UgUm9vdCBDQTAg
Fw0yMTA3MDIxNDE2MDBaGA8yMDUxMDYyNTE0MTYwMFowWTELMAkGA1UEBhMCVVMx
CzAJBgNVBAgTAk5ZMQwwCgYDVQQHEwNMSUMxDDAKBgNVBAoTA0s4UzEMMAoGA1UE
CxMDUktFMRMwEQYDVQQDEwpSS0UgU3ViIENBMIIBIjANBgkqhkiG9w0BAQEFAAOC
AQ8AMIIBCgKCAQEAztg5dQzx0s+VJqEToPmlbcGkIu1/r+ZoNO9J0oL5EX095PCb
jE5gbT6xsw7rWAS2mwF770BRt6ToprXeY2MfstsqDBGddvAdVp3VUXKhPWoabL7P
DdAITSEE6uy3C7zvw39el9IKo4pWsjn+PeICgnr/4TgMszxo2MOE04SH1pGVcwam
[...]

Issue #1834, opened 2019-12-17
Support certificate chains when using custom certs


byo:
**RKE version:**
1.0.0

When specifying custom certificates, RKE assumes that the specified CA directly signs final certificates for kubernetes components. There's no way to specify custom certificates using intermediate CA:
* validation of certificates fails due to issuer name mismatch (i.e. https://github.com/rancher/rke/blob/adc5941/pki/util.go#L773)
* there's no way to inform kubernetes components to serve whole certificate chain instead of just their own certificate

Such intermediate-CA setup would be beneficial in case of intermediate CA rotation as is in our case - it's only needed to supply the long-lived, offline root CA to clients to trust the rotated intermediate certificates.

elmariofredo:
@byo do I understand it correctly that you want to just past intermediate CA to rke and let rke to generate component certs for you same as kubeadm does https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-certs/#external-ca-mode?

byo:
@elmariofredo My use-case is when rke does not generate certificates nor keys at all, those are created outside and passed through `--custom-certs` flag:
```
rke up --config <config_file> --custom-certs --cert-dir <certs_dir>`
```

dje4om:
I tried to provide RKE custom certs a kube-ca.pem with the chain CA (the root CA and intermediate CA) in the same file, wich is valid for openssl, but when rke deploy the certificates on nodes, the kube-ca.pem keep only one cert, so all the chain is invalid.

stoff1973:
@elmariofredo Can we apply the "KISS" principle?
Because, IMHO, I don't think RKE has any good reason to modify these certificates while I explicitly use the "--custom-certs" option.
"Custom" should be understood as "I've craft these files so I don't want RKE change anything in them"

elmariofredo:
Thanks @byo and @stoff1973 for reply, as I mentioned I like how kubeadm is handling certificate generation for you from custom CA. 

I understand that it's simpler to apply KISS here, but this leads to clusters running on self signed certificates which is IMHO security bad practice when company have its own CA.

stoff1973:
For me, there is no relationship between providing a complete chain (CA + intermediate) and using a self-signed certificate.
Talking about kubeadm, there is some commits about this issue :
https://github.com/kubernetes/kubernetes/pull/97266

stoff1973:
[...]

PR #2598, opened 2021-07-02
process pem file with multiple certs when ca contains a chain

chaudhryfaisal:
process pem file with multiple certs when ca contains a chain ( root and sub ca ) and push pem file with chain to nodes instead of only first cert in file

addresses https://github.com/rancher/rke/issues/2597 and https://github.com/rancher/rke/issues/1834


superseb:
Superseded by https://github.com/rancher/rke/pull/2742

PR #2644, opened 2021-08-08
[release/v1.3] go generate

kinarashah:
rancher/rancher#33756


rancherbot (approved):

Issue #2686, opened 2021-09-01
Secrets encryption config breaks on rke up


cmmarslender:
**RKE version:**

1.3.0

**Docker version: (`docker version`,`docker info` preferred)**

**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)**

Mac & Linux both show the issue

**cluster.yml file:**

Relevant section in kube-api

```yaml
  kube-api:
    secrets_encryption_config:
      enabled: true
      custom_config:
        apiVersion: apiserver.config.k8s.io/v1
        kind: EncryptionConfiguration
        resources:
          - resources:
              - secrets
            providers:
              - aescbc:
                  keys:
                    - name: key1
                      secret: "......"
              - identity: { }
```

**Steps to Reproduce:**

Run rke up after setting secrets_encryption_config [per docs](https://rancher.com/docs/rke/latest/en/config-options/secrets-encryption/)

**Results:**

`FATA[0000] Failed to parse cluster file: unexpected type: *schema.GroupVersionKind`

removing the apiVersion line changes to this error instead:

`FATA[0000] Failed to parse cluster file: error decoding data: Object 'apiVersion' is missing...`

superseb:
Does it work when you change `custom_config` to `customConfig`?

cmmarslender:
It doesn't error any more, but I dont think its being loaded properly at that point - the yaml parsing annotation specifies `custom_config` as the key it needs https://github.com/rancher/rke/blob/54dc689ac040076881a5b4af50326d5d15efb2c6/types/rke_types.go#L991

superseb:
Ok I can run some checks, so it doesnt use the custom configuration when it is used this way?

cmmarslender:
Doesn't appear to load it otherwise.

I'm stepping through this with a go debugger, and the error is coming from here:
https://github.com/rancher/rke/blob/release/v1.3/cluster/encryption.go#L578

There is a commit that changed some stuff in this method that wasn't in 1.2.11 (the last version I used) but I'm not sure if its relevant yet https://github.com/rancher/rke/commit/d07658cf9b5a4b9a1891a5477b9919ecc121e1e6

cmmarslender:
Tested the current version of PR #2690 last night and it all seemed to work correctly - encryption config is still set and the cluster upgraded fine, can read secrets, etc

cmmarslender:
[...]

PR #2690, opened 2021-09-04
Fix issues with custom secrets encryption

cmmarslender:
This allows the `parseCustomConfig` method to return successfully now. I'm not entirely why the previous method failed, since the only difference I can see between the non-versioned and the v1 version of `EncryptionConfiguration` is the json annotations, but.. since the json annotations are now there, the decoder seems a bit unnecessary to use - can just unmarshal the json directly into the object.

Not sure what branch you'd prefer this PR against - if you need it updated, let me know

Resolves https://github.com/rancher/rke/issues/2686

### DIFF
--- a/cluster/encryption.go
+++ b/cluster/encryption.go
@@ -510,8 +510,7 @@ func (c *Cluster) readEncryptionCustomConfig() (string, error) {
 		return "", nil
 	}
 
-	return templates.CompileTemplateFromMap(templates.CustomEncryptionProviderFile,
-		struct{ CustomConfig string }{CustomConfig: string(yamlConfig)})
+	return string(yamlConfig), nil
 }
 
 func resolveCustomEncryptionConfig(clusterFile string) (string, *apiserverconfigv1.EncryptionConfiguration, error) {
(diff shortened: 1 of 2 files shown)


cmmarslender:
Ok this does not completely work - generates a config file that starts with duplicates on top - will work on revising:
```apiVersion: apiserver.config.k8s.io/v1
kind: EncryptionConfiguration
apiVersion: apiserver.config.k8s.io/v1
kind: EncryptionConfiguration
resources:
- providers:
  - aescbc:```

**Comment by cmmarslender** [comment]
Resolved the last issue with the most recent commit, and it seems to have deployed to the cluster just fine. I did get the following in the logs, but it appears due to the old template that injected the version/kind having an extra trailing newline compared to the new direct yaml marshal only having the single newline. It was otherwise an identical encryption configuration file

```
INFO[0069] [controlplane] Encryption provider config has changed; reconciling cluster's encryption provider configuration 
INFO[0149] [controlplane] Restarting kube-apiserver on controlplane nodes.. 
```

superseb:
Thanks for the PR, it will be covered in https://github.com/rancher/rke/pull/2701

PR #2713, opened 2021-10-08
fix etcd healthy checking

lshgdut:
Aussuming that I have 3 etcd nodes(A, B, C) to upgrade. 

- node A is unhealthy
- node B, C is healthy

When I try to upgrade cluster by ` rke up`, the result is successful with warning message.

```
time="2021-10-08T10:48:57+08:00" level=info msg="[etcd] Successfully started etcd plane.. Checking etcd cluster health"
time="2021-10-08T10:53:29+08:00" level=warning msg="[etcd] host [10.113.77.16] failed to check etcd health: failed to get /health for host [10.113.77.16]: Get \"https://10.113.77.16:2379/health\": net/http: TLS handshake timeout"
time="2021-10-08T10:53:29+08:00" level=info msg="[etcd] etcd host [10.113.77.15] reported healthy=true"
```

But I think it should be failed according to the error message in the return statement.

### DIFF
--- a/services/etcd.go
+++ b/services/etcd.go
@@ -89,12 +89,12 @@ func RunEtcdPlane(
 		_, _, healthCheckURL := GetProcessConfig(etcdNodePlanMap[host.Address].Processes[EtcdContainerName], host)
 		healthError = isEtcdHealthy(localConnDialerFactory, host, clientCert, clientKey, healthCheckURL)
 		if healthError == nil {
-			break
+			continue
 		}
 		logrus.Warn(healthError)
 		hosts = append(hosts, host.Address)
 	}
-	if healthError != nil {
+	if hosts != nil {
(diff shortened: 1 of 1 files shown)


superseb:
The design chosen was to continue as soon as a healthy cluster member is found. To change this, we need a new design and a proper way to check and what we want to do in case of members not reporting healthy etc as this is quite a big change. Can you create an issue for this?

lshgdut:
> The design chosen was to continue as soon as a healthy cluster member is found. To change this, we need a new design and a proper way to check and what we want to do in case of members not reporting healthy etc as this is quite a big change. Can you create an issue for this?

It's my mistakes on upgrading tests. Certificates of etcd changed on upgrading first time, but I use old `cluster.yaml` to upgrade again. 

Sorry for  that.

PR #2725, opened 2021-10-27
[release/v1.2] Update kdm data

pennyscissors:
https://github.com/rancher/rancher/issues/35164

### DIFF
--- a/metadata/metadata.go
+++ b/metadata/metadata.go
@@ -95,7 +95,7 @@ func readFile(file string) ([]byte, error) {
 	return ioutil.ReadFile(file)
 }
 
-const RKEVersionDev = "v1.2.8-rc0"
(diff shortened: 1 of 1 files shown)


codegen/codegen.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 const (
-	defaultURL = "https://releases.rancher.com/kontainer-driver-metadata/dev-v2.5/data.json"
+	defaultURL = "https://raw.githubusercontent.com/rancher/kontainer-driver-metadata/dev-v2.5-1021-patches/data/data.json"
```
kinarashah: we can keep the URL as is, just changing `dev-v2.5` to `dev-v2.5-1021`-patches would work the same?

pennyscissors: let's see, I'll update it

PR #2729, opened 2021-10-28
[release/v1.3] Add compatibility with k8s v1.22

superseb:
https://github.com/rancher/rancher/issues/33114

- Changed os label to avoid deprecated warnings, only applies to 1.22 and up to avoid any regressions
- Changed health check for kube-controller-manager as it is now HTTPS


addons/addons.go
```diff
@@ -6,26 +6,45 @@ import (
...
+	toMatch, err := semver.Make(k8sVersion[1:])
+	if err != nil {
+		logrus.Warnf("Cluster version [%s] can not be parsed as semver", k8sVersion)
+	}
+
+	logrus.Debugf("Checking addon job OS label for cluster version [%s]", k8sVersion)
+	// kubernetes.io/os should be used 1.22.0 and up
+	OSLabelRange, err := semver.ParseRange(">=1.22.0-rancher0")
+	if err != nil {
+		logrus.Warnf("Failed to parse semver range for checking OS label for addon job")
```
nickgerace: Maybe we should fail here, or at least return?

superseb:
In addition to above, NGINX 1.0.4 was added to 1.22 and higher, needing the `DefaultIngressClass` to make sure we set a default ingress class which is now required (see https://kubernetes.github.io/ingress-nginx/#what-is-an-ingressclass-and-why-is-it-important-for-users-of-ingress-nginx-controller-now). In case users do no want that behavior, it can be switched off.

nickgerace (approved):

kinarashah (dismissed):

kinarashah (approved):

PR #2733, opened 2021-11-01
Support harvester cloud provider

yaocw2020:
Related issue: 
https://github.com/rancher/rancher/issues/35000
https://github.com/harvester/harvester/issues/1396


types/rke_types.go
```diff
@@ -556,6 +556,9 @@ type CloudProvider struct {
 	OpenstackCloudProvider *OpenstackCloudProvider `yaml:"openstackCloudProvider,omitempty" json:"openstackCloudProvider,omitempty"`
 	// VsphereCloudProvider
 	VsphereCloudProvider *VsphereCloudProvider `yaml:"vsphereCloudProvider,omitempty" json:"vsphereCloudProvider,omitempty"`
+	// HarvesterCloudProvider is a string returned by calling the Harvester API "/v1/harvester/kubeconfig"
+	// We should synchronize this field to rancher
+	HarvesterCloudProvider string `yaml:"harvesterCloudProvider,omitempty" json:"harvesterCloudProvider,omitempty"`
```
gitlawr: I think there should be related generated codes for this change?

yaocw2020: Yes, we should bump the rke package in Rancher.

guangbochen: I think lawr is referring to the generated files in this repo, e.g https://github.com/rancher/rke/blob/master/types/zz_generated_deepcopy.go

yaocw2020: Rancher refers rke/types here.  https://github.com/rancher/rancher/blob/ea7823e463847ac1faf3dfdd62dbab7eec48bde1/pkg/apis/management.cattle.io/v3/cluster_types.go#L109

yaocw2020: There should be the related generated codes, but I can't generate code by the command `go generate`. I am still figuring out the reason.

yaocw2020: `HarvesterCloudProvider` is a string so it does not need a deepcopy method. @gitlawr @guangbochen

gitlawr (approved):
LGTM

nickgerace (approved):

types/rke_types.go
StrongMonkey: Can this be a go struct so that it match the format with other cloud providers? Any reason to use string here? Also you get a benefit of adding more fields in harvester cloud provider later without having to tweak the api.

StrongMonkey (changes requested):
Can this be a go struct so that it match the format with other cloud providers? Any reason to use string here? Also you get a benefit of adding more fields in harvester cloud provider later without having to tweak the api.

yaocw2020:
> Can this be a go struct so that it match the format with other cloud providers? Any reason to use string here? Also you get a benefit of adding more fields in harvester cloud provider later without having to tweak the api.

The cloud config for the Harvester cloud provider is actually a kubeconfig generated by calling the Harvester api.

StrongMonkey:
But can you actually make it a go struct similar to other cloud providers? Just for consistency and extensibility.

guangbochen:
@yaocw2020 run the `go generate` should be required with the current struct type change.

yaocw2020:
`go generate` will be functional if the project codes are under the go src path.

StrongMonkey (approved):

nickgerace (approved):
@StrongMonkey I think we are good to merge, right?

guangbochen:
Hey folks, wondering if we should also submit the go mod bump on the `rancher/rancher` after PR merge? thanks in advance.

nickgerace:
@guangbochen yes. I think opening a PR now there could be helpful (`release/v2.6`)

StrongMonkey:
cc @guangbochen assigning @kinarashah as she is the committer and can help coordinate the merge.

kinarashah (approved):
@guangbochen Looks good, do you need RKE CLI RC for this? Lemme know and I can tag one. Otherwise, can just pull in the commit in Rancher go.mod.

guangbochen:
CLI is not necessary for now, just bump the go.mod should be good enough, thanks.

PR #2734, opened 2021-11-02
vendor bumps for k8s 1.22

kinarashah:


go.mod
```diff
@@ -36,19 +32,21 @@ require (
...
+	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
+	google.golang.org/grpc v1.38.0
 	gopkg.in/yaml.v2 v2.4.0
-	k8s.io/api v0.21.0
-	k8s.io/apimachinery v0.21.0
-	k8s.io/apiserver v0.21.0
-	k8s.io/client-go v0.21.0
+	k8s.io/api v0.22.3
+	k8s.io/apimachinery v0.22.3
+	k8s.io/apiserver v0.22.0
```
superseb: do these need to divert?

kinarashah: Nope, thanks - I updated it all to v0.22.3. Confirmed with Darren that for 1.21, we introduced v0.21.0 and then it was never updated but there's no specific reason for not updating to latest.

superseb (approved):

PR #2750, opened 2021-11-19
[release/v1.3] Do not rewrite SELinux labels on volume mounts

superseb:


cluster/plan.go
```diff
@@ -281,6 +283,12 @@ func (c *Cluster) BuildKubeAPIProcess(host *hosts.Host, serviceOptions v3.Kubern
 			fmt.Sprintf("%s=%s", AuditLogConfigSumEnv, getStringChecksum(string(bytes))))
 	}
 
+	matchedRange, _ := util.SemVerMatchRange(c.Version, util.SemVerK8sVersion122OrHigher)
```
nickgerace: There's a few spots where we don't log the error for `SemVerMatchRange`. Even if we expect it to pass 100% of the time, perhaps we should consider debug logging the returned error if non-nil

cluster/selinux.go
```diff
@@ -0,0 +1,79 @@
...
+	for retries := 0; retries < 3; retries++ {
+		logrus.Infof("[selinux] Checking if host [%s] recognizes SELinux label [%s], try #%d", host.Address, SELinuxLabel, retries+1)
+		if err = docker.DoRemoveContainer(ctx, host.DClient, SELinuxCheckContainer, host.Address); err != nil {
+			return err
+		}
+		if err = docker.DoRunOnetimeContainer(ctx, host.DClient, imageCfg, hostCfg, SELinuxCheckContainer, host.Address, "selinux", prsMap); err != nil {
+			if strings.Contains(err.Error(), "invalid argument") {
+				return fmt.Errorf("[selinux] Host [%s] does not recognize SELinux label [%s]. This is required for Kubernetes version [%s]. Please install rancher-selinux RPM package and try again", host.Address, SELinuxLabel, util.SemVerK8sVersion122OrHigher)
+			}
+			continue
```
nickgerace, line 73: Are we potentially skipping an error here on purpose? What happens if `"invalid argument"` is not contained in the error?

util/util.go
```diff
@@ -30,6 +32,54 @@ func StrToSemVer(version string) (*semver.Version, error) {
 	return v, nil
 }
 
+func SemVerMatchRange(k8sVersion, versionRange string) (bool, error) {
+	toMatch, err := sv.Make(k8sVersion[1:])
+	if err != nil {
+		return false, fmt.Errorf("Cluster version [%s] can not be parsed as semver: %v", k8sVersion, err)
+	}
```
nickgerace: I could be wrong here, but are we checking the length? It could be possible to have the `k8sVersion[1:]` result in a non-zero length right?

util/util.go
```diff
@@ -30,6 +32,54 @@ func StrToSemVer(version string) (*semver.Version, error) {
...
+					}
+					cleanBindOptions = append(cleanBindOptions, bindOption)
+
+				}
+			}
+			var newString string
+			if len(cleanBindOptions) > 0 {
+				newString = fmt.Sprintf("%s%s", element[:lastIndex], fmt.Sprintf(":%s", strings.Join(cleanBindOptions, ",")))
+			} else {
+				newString = fmt.Sprintf("%s%s", element[:lastIndex], fmt.Sprintf("%s", strings.Join(cleanBindOptions, ",")))
```
nickgerace: I do not believe we need the `Sprintf` here. Could be wrong.
```go
fmt.Sprintf("%s%s", element[:lastIndex], fmt.Sprintf("%s", strings.Join(cleanBindOptions, ",")))

# could be...

fmt.Sprintf("%s%s", element[:lastIndex], strings.Join(cleanBindOptions, ","))
```

util/util_test.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestRemoveZFromBinds(t *testing.T) {
+	binds := []string{"/etc/kubernetes:/etc/kubernetes:z", "/var/log/kube-audit:/var/log/kube-audit:rw,z", "/var/lib/test:/var/lib/test:Z,ro", "/usr/local/lib/test:/usr/local/lib/test:ro,z,noexec", "/etc/normalz:/etc/normalz"}
+	expectedBinds := []string{"/etc/kubernetes:/etc/kubernetes", "/var/log/kube-audit:/var/log/kube-audit:rw", "/var/lib/test:/var/lib/test:ro", "/usr/local/lib/test:/usr/local/lib/test:ro,noexec", "/etc/normalz:/etc/normalz"}
```
nickgerace, line 11: Is it worth testing empty binds as well and asserting for failure?

cluster/plan.go
superseb: I didn't because the places that allow for returning the error will and it will be caught (at least that was the thought)

cluster/selinux.go
superseb: We are letting it retry in case it is not the known error (or actually the error we are looking for).

util/util.go
superseb: We have `validateVersion` in `cluster/validation.go` to validate this, I'm fine with adding another check in case of better safe than sorry.

util/util.go
superseb: True, this came from a previous attempt where I need to concat the joined value with something else and then no longer needed it.

cluster/plan.go
nickgerace: I'm a bit confused still. Is the idea that even if an error occurs, the consumers of `matchedRange` will know that the returned value is a "failed" one? If that's the case, maybe we shouldn't even return an error at all?

cluster/selinux.go
nickgerace: Gotcha, that's cool! Totally fine if you don't want to, but that could be a good comment to leave at the top of the function or somewhere in there. I do not mind either way.

util/util.go
nickgerace: That might be a good idea even if it's not "possible". We could also write a common "split" utility that the two locations share.

util/util_test.go
superseb: which case are we talking about here? `[]string{}` through the function and then it should the same after? or an empty string element?

nickgerace: Apologies, I'll be precise: I'm curious what happens if we feed `RemoveZFromBinds` an empty string slice and if we should perform an assertion ("good" or "bad") based on that.

cluster/plan.go
superseb: The function does not allow to properly return errors so I removed it and depended on the other invocations (as we added this in many places), but I added log lines for it because it is probably more clear with a log than just swallowing errors.

cluster/selinux.go
superseb: Added

util/util_test.go
superseb: done

nickgerace (approved):

PR #2753, opened 2021-11-24
[release/v1.3] Bump Golang and golangci-lint to match Rancher

superseb:


nickgerace (commented):
We hit an issue in Fleet where upgrading golangci-lint passed v1.41.1 make the `gofmt` linter check for Go 1.17, which causes linting failure if the generated code has not been regenerated with Go 1.17 "build lines": https://go.dev/doc/go1.17

Did this linter pass locally?

nickgerace (approved):
Talked to @superseb offline: this issue likely occurred in Fleet due to something unique in its generator setup and Go build tags. MERGE 🚀

PR #2763, opened 2021-12-04
add extra args for 1.22

kinarashah:
controller manager must start with proper authorization and authentication kubeconfig in args starting 1.22. k8s 1.22 has
disabled insecure serving for kube controller manager. https://github.com/kubernetes/kubernetes/pull/96216 

insecure serving for kube scheduler is going to be disabled starting 1.24, so updating the args for it too. 
https://github.com/kubernetes/kubernetes/pull/102857

Related Issue: https://github.com/rancher/rancher/issues/35751

### DIFF
--- a/cluster/plan.go
+++ b/cluster/plan.go
@@ -381,6 +381,8 @@ func (c *Cluster) BuildKubeControllerProcess(host *hosts.Host, serviceOptions v3
 
 	if matchedRange {
 		Binds = util.RemoveZFromBinds(Binds)
+		CommandArgs["authentication-kubeconfig"] = CommandArgs["kubeconfig"]
+		CommandArgs["authorization-kubeconfig"] = CommandArgs["kubeconfig"]
 	}
 
 	for arg, value := range c.Services.KubeController.ExtraArgs {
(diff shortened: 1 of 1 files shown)


brendarearden (approved):
LGTM

aiyengar2 (approved):
LGTM, might need to backport to 2.5

kinarashah:
@aiyengar2 this is only for >=1.22 and 2.5 doesn't have 1.22, so don't think we need to backport? When we change this for all k8s versions, we'd need PRs to both release/v1.3 and release/v1.2 at that time.

aiyengar2:
> @aiyengar2 this is only for >=1.22 and 2.5 doesn't have 1.22, so don't think we need to backport? When we change this for all k8s versions, we'd need PRs to both release/v1.3 and release/v1.2 at that time.

Technically, the HTTPs port is broken for all versions, including those below 1.22. Whether we want to fix it below 1.22 or just say that users below 1.22 should use the HTTP port (10252, 10251) is something we should clarify; if that is not necessary, then yes I agree with you that we do not need to backport.

kinarashah:
@aiyengar2 Going to use the GH issue to track updates for other k8s versions, merging this as is to unblock monitoring on v1.22 clusters.

PR #2777, opened 2021-12-13
[release/v1.3] Add GitHub Actions workflow for go-generate

superseb:
This adds GitHub Actions workflow which does:
- Check out repository
- Runs `scripts/go-generate` (added so it can be used using make in Actions and locally)
- Checks for changes in the repository (maybe `data.json` is not available yet as it needs to be uploaded in KDM Drone pipeline step)
- If changes found, create pull request and report created pull request URL

Added script to run using `make`,  added `./data` to `DAPPER_OUTPUT` so Dapper copies the changed files after running `make go-generate`


rawmind0 (approved):
LGTM

kinarashah (approved):

PR #2783, opened 2021-12-16
[release/v1.3] Add job complete check in integration

superseb:
* In case nginx admission-create and admission-patch jobs fail, this was not caught before.


rawmind0 (approved):
LGTM

PR #2787, opened 2021-12-24
Update KDM data

rayandas:


rancherbot (approved):

snasovich (approved):
Changes match in principle what was done for similar previous PRs (e.g. https://github.com/rancher/rke/pull/2694) and include latest KDM changes per https://github.com/rancher/kontainer-driver-metadata/pull/765.

@rayandas , do we need PR to `release/v1.2` too as was done before? I'm not too familiar with the process.

PR #2796, opened 2022-01-04
Retry query job status when addon is removing during apply

jakefhyde:
This adds an additional check to ensure that a job isn't currently being deleted before attempting to create a new one and proceed with reconciliation. If the job is being deleted, we now attempt to wait for either completion, or timeout and attempt reconciliation again.

Related Issue: rancher/rancher#35750

### DIFF
--- a/k8s/job.go
+++ b/k8s/job.go
@@ -16,6 +16,7 @@ import (
 type JobStatus struct {
 	Completed bool
 	Created   bool
+	Removing  bool
 }
 
 func ApplyK8sSystemJob(jobYaml, kubeConfigPath string, k8sWrapTransport transport.WrapperFunc, timeout int, addonUpdated bool) error {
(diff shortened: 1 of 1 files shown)


jakefhyde:
/backport "RKE v1.x - Backlog - Rancher v2.x" release/v1.2

thedadams (commented):
On line 54 (with the changes), it looks like if the job exists and is not completed, then Rancher will try to remove the existing job. When deleting the existing job, Rancher will try to wait for its deletion (which seems to be what your changes are doing).

Can your changes be absorbed into this existing logic? For example, changing line 54 to:
`if addonUpdated || (jobStatus.Created && (!jobStatus.Completed || jobStatus.Removing) {`

snasovich:
> On line 54 (with the changes), it looks like if the job exists and is not completed, then Rancher will try to remove the existing job. When deleting the existing job, Rancher will try to wait for its deletion (which seems to be what your changes are doing).
> 
> Can your changes be absorbed into this existing logic? For example, changing line 54 to: `if addonUpdated || (jobStatus.Created && (!jobStatus.Completed || jobStatus.Removing) {`

If the job is already removing, is there any point in trying to delete it again (which is what I believe the proposed change will do)? The way it's in the PR now seems to make sense to me - if the job is removing, wait for the removal is completed before attempting to proceed forward with job creation.

thedadams:
The logic of what this change is doing already exists; my proposal is to use the existing logic.

Deleting a job that is already deleting is fine. Then after the delete call is made, that function will wait (similarly to the changes here).

snasovich:
@thedadams , I understand it's fine as in "it won't cause issues", but still seems pointless to be trying to call delete on something that's already deleting.

thedadams (approved):

snasovich (commented):
The changes LGTM; however, let's wait for Kinara's approval before we merge.

@kinarashah , please review it with priority once you're back. Thank you.

snasovich (approved):

kinarashah (approved):

PR #2803, opened 2022-01-10
delete hostname-override for aws cloud provider

kinarashah:
aws cloud provider assigns private dns as the node name, don't override it since it cannot be changed.

Found while debugging https://github.com/rancher/rancher/issues/30363, kubelet constantly logs 
`
I0108 20:44:28.094203   15255 setters.go:198] "Replacing cloudprovider-reported hostname with overridden hostname" cloudProviderHostname="ip-xxx-xx-xx-xx.us-west-2.compute.internal" overriddenHostname="kinara-cpe-nodes1"
`


jiaqiluo (approved):
LGTM 💯

atsai1220:
Will this affect node templates? I believe it is related: https://github.com/rancher/rancher/issues/22416

superseb (approved):

kinarashah:
@atsai1220 Yeah, the fix would work for RKE clusters provisioned by node templates as well. Thanks for bringing up the issue, I've self assigned it.

wd:
Even though this is already merged, I want to point out that this will cause the issue below. 

When you have an existing cluster(provisioned before this PR) on AWS with the `hostname-override` has been set, the node will register with that hostname. 

After you upgraded the RKE engine(i.e. the rke command), when you do some upgrade to your cluster, RKE will recreate the kubelet container without the `hostname-override`, now the hostname will not match the old one. 

RKE will print the error here https://github.com/rancher/rke/blob/v1.3.5/k8s/node.go#L72, because of the corporation here https://github.com/rancher/rke/blob/v1.3.5/k8s/node.go#L55 will fail. 
```
[controlPlane] Failed to upgrade Control Plane: [[[controlplane] Error getting node ip-172-19-x-y us-east-2.compute.internal: "ip-172-19-x-y.us-east-2.compute.internal" not found]]
```

wd:
The fixes for this issue is run the command like below to set the hostname back.
```
kubectl label no ip-172-19-x-y.us-east-2.compute.internal kubernetes.io/hostname=ip-172-19-x-y.us-east-2.compute.internal --overwrite
```

herrbpl:
This patch btw kills your cluster Nodeport services if you had cluster installed with hostname override.

herrbpl:
This is bad move to remove this. Because it seems kube-proxy when run in docker, uses hostname info from underlying distroy and this always does not match to what 
` echo $(curl --silent "http://169.254.169.254/latest/meta-data/hostname")`
gives you.
For example, on ubuntu, hostname gives on aws:
```bash
root@ip-172-31-112-194:~# echo $(curl --silent "http://169.254.169.254/latest/meta-data/hostname")
ip-172-31-112-194.eu-central-1.compute.internal
root@ip-172-31-112-194:~# hostname
ip-172-31-112-194
```
while on Amazon Linux it gives:
```bash
[root@ip-172-31-29-94 ~]# echo $(curl --silent "http://169.254.169.254/latest/meta-data/hostname")
ip-172-31-29-94.eu-central-1.compute.internal
[root@ip-172-31-29-94 ~]# hostname
ip-172-31-29-94.eu-central-1.compute.internal
[root@ip-172-31-29-94 ~]#
```
Taking away option to configure this for kube-proxy is screwing all those ppl who have this value different

herrbpl:
Looks like workaround is to use 
```bash
hostnamectl set-hostname $(curl --silent "http://169.254.169.254/latest/meta-data/hostname")
```
in userdata of launch template/etc before custom rancher node is joined.

herrbpl:
Also, in working cluster, following can be used on RKE1 provisioned cluster nodes on Ubuntu (for ex using Ansible) to update kube-proxy config so that hostname matches ec2 metadata and NodePorts work again:

```bash
#!/bin/bash
apt update && apt install jq -y
hostnamectl set-hostname $(curl --silent "http://169.254.169.254/latest/meta-data/hostname")

podpath=$(dirname $(docker inspect kube-proxy | jq  '.[0].HostnamePath' -r))

configpath=${podpath}/config.v2.json
hostnamepath=${podpath}/hostname
hostname=$(hostname)
cat $configpath | jq --arg hostname "$hostname" '.Config.Hostname=$hostname' -c -r > /tmp/config.v2.json
service docker stop
mv "$configpath" "$configpath.old"
mv /tmp/config.v2.json "$configpath"
echo "$hostname" > "$hostnamepath"
service docker start
```

kinarashah:
@herrbpl Thanks for pointing this out with the steps you used as a workaround, I am working on fixing this and might revert it altogether. Before this change, did you configure the hostname-override yourself? Did it also match the node name that kubelet would register it as from the cloud provider? My understanding is that both hostname overrides on both kubelet and kube-proxy should match so wanted to confirm.

herrbpl:
@kinarashah Yes, we added custom rancher nodes with following command:

`sudo docker run -d --privileged --restart=unless-stopped --net=host -v /etc/kubernetes:/etc/kubernetes -v /var/run:/var/run rancher/rancher-agent:v2.6.3 --server https://REDACTED --token REDACTED --node-name `hostname -f` --address awslocal --internal-address awslocal --worker --label ... --taint ... `

precisely because by on ubuntu, hostname gave back short name. At time we did not think to patch hostname and had no suspicion it is so important, instead we went with provided argument to configure hostname for kubelet/kube-proxy. When option dropped with upgrade from 2.6.3 to 2.6.4 we had trouble with getting traafic reaching to our ingress controllers. This was very nasty experience and because under time pressure, we went with reinstall, no time to investigate uring duress. 

In rancher, nodes were registered with full FQDN.

wd:
> This patch btw kills your cluster Nodeport services if you had cluster installed with hostname override. We lost considerable amount of $ because of this particular change to remove repeating warning from logs.

Yes. We added this argument to fix the NodePort issue previously. Fortunately, we don't use NodePort anymore. So it is not a big issue to us.

kinarashah:
Since RKE relies on the hostname label to match with the hostname-override label even if it doesn't match the node name when cloud provider is enabled, the best approach here would be to revert back to original behavior for kubelet. 

Tho, the bigger issue is how to manage situations where users have already fixed the setups by adjusting the label to match the node name - it'll run into issues again when the hostname-override starts getting respected.

kinarashah:
@herrbpl @wd I've made a few changes around this for kubelet and kube-proxy and is merged to be released in the next release - https://github.com/rancher/rke/pull/3136. Let me know if you see any issues with this. Tried to cover all use cases, tho I understand that it might require manual intervention for some users to set the extra field depending on the usage. Thanks once again for the feedback.

Issue #1669, opened 2019-09-26
Allow for configuration of more details of certs


Tejeev:
**RKE version:**
2.8

**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)**
ubuntu 18.04

What kind of request is this (question/bug/enhancement/feature request):
Feature Request

Request
We would like to configure more details when generating certs with rke cert generate-csr
Some details that we would like to add:
- Locality
- State/Province
- Country

We would like to configure the following but understand they are not configurable as they are used by Kubernetes components (please correct me if I'm mistaken):
- Common Name
- Organization
- OU
- Key size

2easy:
+1 

I'd add encryption type as well. We'd like to use `ecdsa` certs and keys.

PR #2742, opened 2021-11-11
process pem file with multiple certs when ca contains a chain

chaudhryfaisal:
process pem file with multiple certs when ca contains a chain ( root and sub ca ) and push pem file with chain to nodes instead of only first cert in file

addresses https://github.com/rancher/rke/issues/2597 and https://github.com/rancher/rke/issues/1834

existing PR https://github.com/rancher/rke/pull/2598

PR #2816, opened 2022-01-31
extend pki with support for ed25519/ecdsa certs and certificate chains

Agalin:
Fixes https://github.com/rancher/rke/issues/2597 and https://github.com/rancher/rke/issues/1834.
It would be a start of https://github.com/rancher/rke/issues/1669 as well.

While there is support for reading and writing *ecdsa* certificates,
any private keys that are loaded from files (either from custom certs
or already deployed on the nodes) are being cast to `*rsa.PrivateKey`.
This results in crashes when *ecdsa* certificates are being used.

Similarly, there is built in support for reading certificate chains from
a single certificate file, but then only the leaf certificate is being
processed throughout the whole chain and ultimately written/saved on the
nodes, effectively dropping all cert chain information. This is a major
PITA when using custom certificates with intermediates in the chain.

This commit attempts to fix both this issues:
- a new `PrivateKey` interface is introduced to allow passing around
  both rsa and ecdsa private keys through the pki code
- the `CertificatePKI` type is extended to also hold the certificate
  chain and code is adjusted accordingly to work with that
- internally (rke) generated certificates are still hardcoded to use
  rsa keys, but defaults have been changed to use 4096 bit keys instead
  of the much less secure 2048 bit keys

Additionally support has been added for Ed25519. ECDSA keys use NIST P
curves. Ed25519 offers better security and is faster than NIST P curves.
It is a good idea to have support for them, both in terms of loading from
custom certs as well as generating internally.

All necessary methods to support Ed25519 keys have been added/updated
in this commit. To make it possible to use Ed25519 keys internally
generated by rke appropriate configuration options need to be added to
change the default key type (hardcoded to rsa for backward
compatibility).

### DIFF
--- a/cluster/certificates.go
+++ b/cluster/certificates.go
@@ -2,7 +2,6 @@ package cluster
 
 import (
 	"context"
-	"crypto/rsa"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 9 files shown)


chaudhryfaisal:
there is PR to process pem file with multiple certs when ca contains a chain https://github.com/rancher/rke/pull/2742

Agalin:
Implementation of chain support differs quite heavily so we'd rather wait for #2742 being merged to master before this PR is changed to work with it.

Agalin:
Still valid. Really hope either this or #2742 gets merged soon.

PR #2830, opened 2022-02-09
update k8s to 1.23.3

kinarashah:


maxsokolovsky (commented):
LGTM! I built RKE locally, just in case, everything is good!

maxsokolovsky (approved):

PR #2851, opened 2022-02-25
update golang to 1.17.6 and k8s to 1.23.3

kinarashah:


codegen/codegen.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 const (
-	defaultURL = "https://releases.rancher.com/kontainer-driver-metadata/dev-v2.6/data.json"
+	defaultURL = "https://raw.githubusercontent.com/rancher/kontainer-driver-metadata/59bf03888f4f38d5aef44756802bf9830b049d61/data/data.json"
```
paynejacob: Do we want to move the default away from the release site?

kinarashah: Nope, thanks for calling it out - I used my dev commit for checking if tests pass and forgot updating it.

kinarashah: I'll update it back to dev-v2.6 after I merge KDM changes

cluster/plan.go
```diff
@@ -833,9 +834,26 @@ func (c *Cluster) BuildSchedulerProcess(host *hosts.Host, serviceOptions v3.Kube
...
+	k8sTag, err := util.GetImageTagFromImage(c.SystemImages.Kubernetes)
+	if err != nil {
+		logrus.Warn(err)
 	}
+	k8sSemVer, err := util.StrToSemVer(k8sTag)
+	if err != nil {
+		logrus.Warn(err)
+	}
+	healthCheck := v3.HealthCheck{}
+	if k8sSemVer.LessThan(*maxK8s122Version) {
```
thedadams: Since we are more-or-less ignoring the error from `util.StrToSemVer`, do we want to add a `nil`-check for `k8sSemVer` here as well to avoid a possible nil-pointer dereference?

thedadams (dismissed):

paynejacob (approved):

jiaqiluo (approved):

Issue #2840, opened 2022-02-16
rke crashes if rotate_encryption_key is set when creating a new cluster


jiaqiluo:
**RKE version:**
v1.3.7
**Docker version: (`docker version`,`docker info` preferred)**
n/a
**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)**
n/a
**Type/provider of hosts: (VirtualBox/Bare-metal/AWS/GCE/DO)**
n/a
**cluster.yml file:**
```
nodes:
  - address:  xxx
    internal_address: xxx
    user: ubuntu
    role: [controlplane,worker,etcd]
    ssh_key_path: xxx
rotate_encryption_key: true
```

**Steps to Reproduce:**
run `rke up`

**Results:**

RKE crashes with the following error
```
INFO[0016] Rotating cluster secrets encryption key
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x2780c9e]

goroutine 1 [running]:
github.com/rancher/rke/cluster.InitClusterObject(0x300a758, 0xc0000440a0, 0x0, 0x0, 0x0, 0x7ff7bfeffa52, 0x15, 0x0, 0x0, 0x0, ...)
	/go/src/github.com/rancher/rke/cluster/cluster.go:705 +0xde
github.com/rancher/rke/cmd.RotateEncryptionKey(0x300a758, 0xc0000440a0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x7ff7bfeffa52, 0x15, ...)
	/go/src/github.com/rancher/rke/cmd/encryption.go:85 +0x185
github.com/rancher/rke/cmd.ClusterUp(0x300a758, 0xc0000440a0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x7ff7bfeffa52, 0x15, 0x0, ...)
	/go/src/github.com/rancher/rke/cmd/up.go:107 +0x35b
github.com/rancher/rke/cmd.clusterUpFromCli(0xc00015de40, 0x0, 0xc00015de40)
	/go/src/github.com/rancher/rke/cmd/up.go:309 +0x645
github.com/urfave/cli.HandleAction(0x29493e0, 0x2d8e618, 0xc00015de40, 0xc00047d300, 0x0)
	/go/pkg/mod/github.com/urfave/cli@v1.20.0/app.go:490 +0x82
[...]

PR #2855, opened 2022-03-02
fix issue around rotating the encryption key 

jiaqiluo:
Issue: 
- https://github.com/rancher/rancher/issues/36333 
- https://github.com/rancher/rke/issues/2840

Problem:
The bug is found in Rancher.
RKE will fail to rotate the keys in the following scenario:
- provision an RKE cluster with secrets Encryption disabled, which is the default behavior 
- wait for the cluster to be available 
- edit the cluster to set `rotateEncryptionKey: true` and `secretsEncryptionConfig.enable: true`
  
Solutions:
We rotate the encryption key only when updating an existing cluster that has secret Encryption enabled; all other cases will be handled later by reconciling the encryption provider config.  In this particular case, new certs will be generated when enabling the secret encryption.

Tests:
The tests are done in Rancher, I expect to see the same result with standalone RKE.
The steps are exactly the same as the above.
Results: the cluster is updated successfully, rancher logs show the certs are rotated. 

TODO:

- [ ] once the PR is merged, update rancher's dependencies


cmd/up.go
```diff
@@ -103,9 +103,10 @@ func ClusterUp(ctx context.Context, dialersOptions hosts.DialersOptions, flags c
 		return rebuildClusterWithRotatedCertificates(ctx, dialersOptions, flags, svcOptionsData)
 	}
 	// if we need to rotate the encryption key, do so and then return
+	// note that we rotate the encryption key only when updating an existing cluster that has secret encryption enabled
+	// all other cases will be handled later by reconciling the encryption provider config
 	if kubeCluster.RancherKubernetesEngineConfig.RotateEncryptionKey {
-		// rotate the encryption key only when updating an existing cluster
-		if clusterState.CurrentState.RancherKubernetesEngineConfig != nil {
+		if clusterState.CurrentState.RancherKubernetesEngineConfig != nil && clusterState.CurrentState.RancherKubernetesEngineConfig.Services.KubeAPI.SecretsEncryptionConfig.Enabled {
```
thedadams: `clusterState.CurrentState.RancherKubernetesEngineConfig.Services.KubeAPI.SecretsEncryptionConfig` is a pointer. Should we have a `nil`-check on that too?

paynejacob (approved):

cmd/up.go
jiaqiluo: Good catch! 
Fixed.

jiaqiluo: I will squash commits into one after getting your approval.

thedadams (dismissed):

thedadams (approved):

jakefhyde (dismissed):

PR #2869, opened 2022-03-15
fix auth kubeconfig not passed to kube-scheduler args

kinarashah:
**Problem:** 
authorization and authentication kubeconfig are set in commandArgs after command is generated for kube-scheduler. 

Issue always existed, but insecure serving was disabled for kube-scheduler starting 1.23, causing https://github.com/rancher/rancher/issues/36899


aiyengar2 (approved):
LGTM

PR #1279, opened 2019-04-10
Always re-run add-on deployment jobs to ensure they truley stay in sync.

jlamillan:
RKE add-ons (network, dns, etc.) can get out of sync [[1](https://github.com/rancher/rke/issues/815),[2](https://github.com/rancher/rke/issues/1278)] with respect to the `cluster.yaml` file regardless of whether or not the `cluster.yaml` has been updated (e.g. `kubectl delete deployment kube-dns`).  

This PR ensures RKE add-ons are brought back in sync each time the `up` command is run by simply re-running the add-on deploy-jobs and letting `kubectl apply` do the heavy lifting of determining whether or not changes are actually required. If nothing has changed, re-running the jobs should not be a problem since they will be noops.

PR #2900, opened 2022-04-08
Add logic to fix coreDNS add-on config issue while cluster upgrade

rayandas:
Fixes: https://github.com/rancher/rancher/issues/36985

For every user addon, rke runs an `rke-user-addon-deploy-job` job, which applies the user add-on configs to the cluster. Although when the cluster version gets updated, the user addon needs to be re-applied.

Currently if the user addons are previously applied, they get ignored subsequently, even if the k8s version is being upgraded. Hence, this PR changes that behavior to force-apply the user addons for each `rke up` command run. This will make sure that even if the k8s version of the cluster is upgraded, the user addon changes do not get lost.

### DIFF
--- a/cluster/addons.go
+++ b/cluster/addons.go
@@ -186,6 +186,18 @@ func (c *Cluster) deployK8sAddOns(ctx context.Context, data map[string]interface
 func (c *Cluster) deployUserAddOns(ctx context.Context) error {
 	log.Infof(ctx, "[addons] Setting up user addons")
 	if c.Addons != "" {
+		addonJobExists, err := addons.AddonJobExists(UserAddonJobName, c.LocalKubeConfigPath, c.K8sWrapTransport)
+		if err != nil {
+			return nil
+		}
+		if addonJobExists {
+			log.Infof(ctx, "[addons] Removing old user addons")
+			if err := c.doAddonDelete(ctx, UserAddonResourceName, false); err != nil {
+				return err
+			}
+
+			log.Infof(ctx, "[addons] Old user addons removed successfully")
+		}
 		if err := c.doAddonDeploy(ctx, c.Addons, UserAddonResourceName, false); err != nil {
 			return err
 		}


annablender (approved):
LGTM thank you @rayandas

annablender (approved):

cluster/addons.go
```diff
@@ -186,6 +186,18 @@ func (c *Cluster) deployK8sAddOns(ctx context.Context, data map[string]interface
 func (c *Cluster) deployUserAddOns(ctx context.Context) error {
 	log.Infof(ctx, "[addons] Setting up user addons")
 	if c.Addons != "" {
+		addonJobExists, err := addons.AddonJobExists(UserAddonJobName, c.LocalKubeConfigPath, c.K8sWrapTransport)
+		if err != nil {
+			return nil
```
snasovich, line 191: Why do we return `nil` here (I don't think we should ignore any error here).

superseb:
So there a few things to be considered in this case:

- The original issue as reported is expected. I don't really understand why there is KB article for something that won't last until next upgrade, this seems like postponing an annoying surprise to the user. The way it was designed (as far as I know as I did not design it), the add-ons are deployed with a job and saved as a configmap. And the state is determined by checking if the provided configuration is different from the one stored in the configmap (or if the job did not succeed). So what happens currently is working as expected.
- Which brings me to the 2nd part, the KB article solution uses user addons to modify a resource that is in our system add-ons. The way this is normally handled is by adding the configuration option to the template of the add-on, so that you can define how the add-ons should behave in a single place (and that the config is applied even if you are upgrading).

So I assume it was determined to change this design choice of RKE, which is quite a big change in behavior for users as even though it looks like the right choice to make sure it is kept in sync, it is still a major change in how RKE behaves and there is no real way to test all usecases that might be affected by this.

If it needs to be changed then there are two choices:
- Change behavior for all addons (similar to an attempt here https://github.com/rancher/rke/pull/1279)
- Scope the change of behavior to user addons only (which is this PR)

The current PR basically copies the logic that is used to remove jobs if someone removes the add-on from the config, so it gets deleted. Technically fine and it probably works. Another option would be to add a check for the user-addon-job in https://github.com/rancher/rke/blob/release/v1.3/k8s/job.go#L54 so that it is in the place where the condition for replacing are described and it will all be in the same place.

Let me know if there are any questions or if something doesn't make sense.

cluster/addons.go
rayandas: ah! right the error should be returned here.

rayandas:
@superseb Yeah your points makes sense. I removed the old logic (what this PR has) and added a condition in `rke/k8s/job.go` to change the behaviour of all addons as you suggested. Also tested it locally. It's working fine. I can raise a PR to revert this PR's changes and add new changes. 
cc: @kinarashah

kinarashah:
@rayandas @superseb  Given that we're fixing an issue happening due to user passing a config map in the user addon section (vs the entire user addon yaml), maybe it'd make sense for us to reduce the scope of the fix to user addons only. 

I agree that technically the end behavior remains the same because `kubectl apply` runs after the job is deleted. But this also introduces a change in behavior for our system addons via `ApplyK8sSystemJob` which is working fine as is right now. If customers run into issues with delete and redeploy on upgrade for system addons, it'll be a regression in for system addons that had no issues in the first place. What do you think?

PR #2904, opened 2022-04-18
Fix coreDNS addons issue

rayandas:
Raised this PR as per the suggestions here: https://github.com/rancher/rke/pull/2900#issuecomment-1100071645
I tested it, working fine.
issue: https://github.com/rancher/rancher/issues/36985

### DIFF
--- a/cluster/addons.go
+++ b/cluster/addons.go
@@ -186,18 +186,6 @@ func (c *Cluster) deployK8sAddOns(ctx context.Context, data map[string]interface
 func (c *Cluster) deployUserAddOns(ctx context.Context) error {
 	log.Infof(ctx, "[addons] Setting up user addons")
 	if c.Addons != "" {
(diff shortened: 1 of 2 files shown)


superseb (approved):
The referenced comment provides two choices based on the impact of the change. One was to apply to all, the other to user add-ons. Although technically it should be fine to run it always, it is still a change in design/working and too big for a patch release in my opinion.

But as patch releases have consisted of bigger changes than those, I guess that choice was made. (this is an assumption as there is no explanation of the choice in the issue or PR)

snasovich:
Closing this based on https://github.com/rancher/rke/pull/2900#issuecomment-1102708230

However, we still need https://github.com/rancher/rke/pull/2900/files#r850613706 addressed. @rayandas , could you please make the change and open a new PR?

PR #2908, opened 2022-04-20
Add fields `ExtraArgsArray` and `WindowsExtraArgsArray`

HarrisonWAffel:
Issue: https://github.com/rancher/rancher/issues/25500
 
# Problem
When setting up [service account token volume projection](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection) within the cluster.yaml, users were not able to specify a flag multiple times with different values within the `ExtraArgs` section of the `kubeapi` service. This prevents users from specifying multiple `service-account-key-file`’s, multiple `service-account-issuer`’s, or multiple `api-audiences`. This issue may also exist for the other services, and is not specific to service account projection (although that is the requested usecase)

# Solution
In an effort to preserve backwards compatibility, two new fields have been added to the `BaseService`: `ExtraArgsArray`/`extra_args_array` & `WindowsExtraArgsArray`/`win_extra_args_array`. These fields have a type of `map[string][]string`, as opposed to the type of `map[string]string` which is used for the existing `extraArgs` field. These new fields allow users to specify arguments multiple times with different values. These new fields can be used within any of the services within the cluster.yaml. _If anyone has any better names for these two fields please let me know, I'm happy to change them._ 

# Testing
I was able to utilize the new field within the `kubeapi` service when creating or modifying a cluster. I was able to observe updates / failures when specifying valid or invalid values for the arguments. Here are the test cases I have performed:

Test case 1) 
 + Create a new DO cluster (1 node, all roles)
 + Once available, SSH into the node
   + run the following commands 
     + `cp /etc/kubernetes/ssl/kube-apiserver-key.pem /etc/kubernetes/ssl/kube-apiserver-key2.pem`
     + `cp /etc/kubernetes/ssl/kube-apiserver-key.pem /etc/kubernetes/ssl/kube-apiserver-key3.pem`
 + modify the cluster.yaml and replace the `kubeapi` section with the following 
```
kube-api:
      always_pull_images: false
      extra_args_array:
        service-account-key-file:
          - /etc/kubernetes/ssl/kube-apiserver-key.pem
          - /etc/kubernetes/ssl/kube-apiserver-key2.pem
          - /etc/kubernetes/ssl/kube-apiserver-key3.pem
      pod_security_policy: false
      secrets_encryption_config:
        enabled: false
      service_node_port_range: 30000-32767
```
 + apply changes and observe cluster go into `active` state

Test case 2) 
 + create new downstream cluster in DO (1 node, all roles)
 + modify the cluster yaml BEFORE creating the cluster
 + Replace the `kube-api` field with the following value 
```kube-api:
      always_pull_images: false
      extra_args_array:
        service-account-key-file:
          - /etc/kubernetes/ssl/kube-apiserver-key.pem
          - /etc/kubernetes/ssl/blah.pem
      pod_security_policy: false
      secrets_encryption_config:
        enabled: false
      service_node_port_range: 30000-32767
```
+ Save the cluster 
+ Observe cluster errors out with the following error (expected) ```
[controlPlane] Failed to upgrade Control Plane: [[Failed to verify healthcheck: Failed to check https://localhost:6443/healthz for service [kube-apiserver] on host [104.131.53.240]: Get "https://localhost:6443/healthz": dial tcp [::1]:6443: connect: connection refused, log: Error: open /etc/kubernetes/ssl/blah.pem: no such file or directory]]```

If there are other test cases I should attempt before merging this PR let me know!

### DIFF
--- a/cluster/plan.go
+++ b/cluster/plan.go
@@ -188,6 +188,8 @@ func (c *Cluster) BuildKubeAPIProcess(host *hosts.Host, serviceOptions v3.Kubern
 		"tls-cert-file":                pki.GetCertPath(pki.KubeAPICertName),
 		"tls-private-key-file":         pki.GetKeyPath(pki.KubeAPICertName),
 	}
+	CommandArrayArgs := make(map[string][]string, len(c.Services.KubeAPI.ExtraArgsArray))
+
 	if len(c.CloudProvider.Name) > 0 {
 		CommandArgs["cloud-config"] = cloudConfigFileName
 	}
@@ -300,10 +302,12 @@ func (c *Cluster) BuildKubeAPIProcess(host *hosts.Host, serviceOptions v3.Kubern
 		}
 	}
 
-	for arg, value := range CommandArgs {
-		cmd := fmt.Sprintf("--%s=%s", arg, value)
-		Command = append(Command, cmd)
+	for arg, value := range c.Services.KubeAPI.ExtraArgsArray {
+		CommandArrayArgs[arg] = value
 	}
+
+	Command = appendArgs(Command, CommandArgs)
+	Command = appendArrayArgs(Command, CommandArrayArgs)
 
 	Binds = append(Binds, c.Services.KubeAPI.ExtraBinds...)
 
@@ -339,6 +343,7 @@ func (c *Cluster) BuildKubeControllerProcess(host *hosts.Host, serviceOptions v3
 		"service-account-private-key-file": pki.GetKeyPath(pki.ServiceAccountTokenKeyName),
 		"service-cluster-ip-range":         c.Services.KubeController.ServiceClusterIPRange,
 	}
+	CommandArrayArgs := make(map[string][]string, len(c.Services.KubeAPI.ExtraArgsArray))
 	// Best security practice is to listen on localhost, but DinD uses private container network instead of Host.
 	if c.DinD {
 		CommandArgs["address"] = "0.0.0.0"
@@ -392,10 +397,13 @@ func (c *Cluster) BuildKubeControllerProcess(host *hosts.Host, serviceOptions v3
 		}
 	}
 
-	for arg, value := range CommandArgs {
-		cmd := fmt.Sprintf("--%s=%s", arg, value)
-		Command = append(Command, cmd)
+	for arg, value := range c.Services.KubeController.ExtraArgsArray {
+		CommandArrayArgs[arg] = value
 	}
+
+	Command = appendArgs(Command, CommandArgs)
+	Command = appendArrayArgs(Command, CommandArrayArgs)
+
 	k8sTag, err := util.GetImageTagFromImage(c.SystemImages.Kubernetes)
 	if err != nil {
 		logrus.Warn(err)
@@ -595,6 +603,14 @@ func (c *Cluster) BuildKubeletProcess(host *hosts.Host, serviceOptions v3.Kubern
 	for arg, value := range host.GetExtraArgs(kubelet.BaseService) {
 		CommandArgs[arg] = value
 	}
+	extraArgsArray := host.GetExtraArgsArray(kubelet.BaseService)
+	CommandArrayArgs := make(map[string][]string, len(extraArgsArray))
+	for arg, value := range extraArgsArray {
+		CommandArrayArgs[arg] = value
+	}
+
+	Command = appendArgs(Command, CommandArgs)
+	Command = appendArrayArgs(Command, CommandArrayArgs)
 
 	// If nodelocal DNS is configured, set cluster-dns to local IP
 	if c.DNS.Nodelocal != nil && c.DNS.Nodelocal.IPAddress != "" {
@@ -608,7 +624,7 @@ func (c *Cluster) BuildKubeletProcess(host *hosts.Host, serviceOptions v3.Kubern
 
 	return v3.Process{
 		Name:                    services.KubeletContainerName,
-		Command:                 appendArgs(Command, CommandArgs),
+		Command:                 Command,
 		VolumesFrom:             VolumesFrom,
 		Binds:                   getUniqStringList(Binds),
 		Env:                     getUniqStringList(Env),
@@ -707,13 +723,22 @@ func (c *Cluster) BuildKubeProxyProcess(host *hosts.Host, serviceOptions v3.Kube
 		CommandArgs[arg] = value
 	}
 
+	extraArgsArray := host.GetExtraArgsArray(kubeproxy.BaseService)
+	CommandArrayArgs := make(map[string][]string, len(extraArgsArray))
+	for arg, value := range extraArgsArray {
+		CommandArrayArgs[arg] = value
+	}
+
+	Command = appendArgs(Command, CommandArgs)
+	Command = appendArrayArgs(Command, CommandArrayArgs)
+
 	healthCheck := v3.HealthCheck{
 		URL: services.GetHealthCheckURL(false, services.KubeproxyPort),
 	}
 	registryAuthConfig, _, _ := docker.GetImageRegistryConfig(kubeproxy.Image, c.PrivateRegistriesMap)
 	return v3.Process{
 		Name:                    services.KubeproxyContainerName,
-		Command:                 appendArgs(Command, CommandArgs),
+		Command:                 Command,
 		VolumesFrom:             VolumesFrom,
 		Binds:                   getUniqStringList(Binds),
 		Env:                     getUniqStringList(Env),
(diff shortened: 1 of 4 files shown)


cluster/plan.go
```diff
@@ -392,6 +393,12 @@ func (c *Cluster) BuildKubeControllerProcess(host *hosts.Host, serviceOptions v3
...
 
+	for arg, value := range c.Services.KubeController.ExtraArgsArray {
+		CommandArgs[arg] = value
+	}
+
+	Command = appendArgs(Command, CommandArgs)
+
 	for arg, value := range CommandArgs {
 		cmd := fmt.Sprintf("--%s=%s", arg, value)
 		Command = append(Command, cmd)
```
jiaqiluo: It looks like the above call to appenArgs has done the work, so we can drop this section.

cluster/plan.go
```diff
@@ -596,6 +603,12 @@ func (c *Cluster) BuildKubeletProcess(host *hosts.Host, serviceOptions v3.Kubern
 		CommandArgs[arg] = value
 	}
 
+	for arg, value := range host.GetExtraArgsArray(kubelet.BaseService) {
+		CommandArgs[arg] = value
+	}
+
+	Command = appendArgs(Command, CommandArgs)
```
jiaqiluo, line 612: we should drop this line because the same function will be called at line 624 
```
Command: appendArgs(Command, CommandArgs),
```

cluster/plan.go
```diff
@@ -1044,8 +1066,16 @@ func (c *Cluster) BuildEtcdProcess(host *hosts.Host, etcdHosts []*hosts.Host, se
...
+		case []string:
+			for _, v := range value.([]string) {
+				cmd := fmt.Sprintf("--%s=%s", arg, v)
+				args = append(args, cmd)
+			}
+		case string:
+			cmd := fmt.Sprintf("--%s=%s", arg, value)
+			args = append(args, cmd)
+		}
 	}
```
jiaqiluo: can we simplify this section like the below?
```
args := appendArgs(args, CommaArgs)
```

hosts/hosts.go
```diff
@@ -420,6 +420,15 @@ func (h *Host) GetExtraArgs(service v3.BaseService) map[string]string {
 	}
 }
 
+func (h *Host) GetExtraArgsArray(service v3.BaseService) map[string][]string {
+	switch {
+	case h.OS() == "windows" && len(service.WindowsExtraArgsArray) > 0:
```
jiaqiluo: If the `ExtraArgsArray` is supposed to be used when the OS is NOT windows, in the case where we have `ExtraArgsArray` but not `WindowsExtraArgsArray`, will it return the `ExtraArgsArray`? 

```suggestion
	case h.OS() == "windows":
```

jiaqiluo (changes requested):

hosts/hosts.go
HarrisonWAffel: Yes it will, this functions in the same fashion as the function above it `GetExtraArgs`

cluster/plan.go
```diff
@@ -953,7 +965,7 @@ func (c *Cluster) BuildEtcdProcess(host *hosts.Host, etcdHosts []*hosts.Host, se
 		listenAddress = "0.0.0.0"
 	}
 
-	CommandArgs := map[string]string{
+	CommandArgs := map[string]interface{}{
```
kinarashah: Is there a particular reason why we want to convert string to interface here? I'd prefer keeping commandArgs and commandArgsArray separate to make it explicit vs having to use switch case on types. Also this'd help ensure we only need to test this new field. 

I do like how you updated it to use one function instead of repeating the same logic for all plans 👍🏼 (maybe we could have 2 funcs instead of one)

hosts/hosts.go
jiaqiluo: I am ok with it if that is how we handle the same case at all places.

jiaqiluo (approved):
LGTM

cluster/plan.go
```diff
@@ -188,6 +188,8 @@ func (c *Cluster) BuildKubeAPIProcess(host *hosts.Host, serviceOptions v3.Kubern
 		"tls-cert-file":                pki.GetCertPath(pki.KubeAPICertName),
 		"tls-private-key-file":         pki.GetKeyPath(pki.KubeAPICertName),
 	}
+	CommandArrayArgs := map[string][]string{}
```
thedadams:
```suggestion
	CommandArrayArgs := make(map[string][]string, len(c.Services.KubeAPI.ExtraArgsArray))
```

cluster/plan.go
```diff
@@ -339,6 +343,7 @@ func (c *Cluster) BuildKubeControllerProcess(host *hosts.Host, serviceOptions v3
 		"service-account-private-key-file": pki.GetKeyPath(pki.ServiceAccountTokenKeyName),
 		"service-cluster-ip-range":         c.Services.KubeController.ServiceClusterIPRange,
 	}
+	CommandArrayArgs := map[string][]string{}
```
thedadams:
```suggestion
	CommandArrayArgs := make(map[string][]string, len(c.Services.KubeController.ExtraArgsArray))
```

cluster/plan.go
```diff
@@ -451,6 +459,7 @@ func (c *Cluster) BuildKubeletProcess(host *hosts.Host, serviceOptions v3.Kubern
 		"pod-infra-container-image": kubelet.InfraContainerImage,
 		"root-dir":                  path.Join(host.PrefixPath, "/var/lib/kubelet"),
 	}
+	CommandArrayArgs := map[string][]string{}
```
thedadams: Can we do something similar here as with my previous suggestions to save on unnecessary allocations?

cluster/plan.go
```diff
@@ -633,6 +649,7 @@ func (c *Cluster) BuildKubeProxyProcess(host *hosts.Host, serviceOptions v3.Kube
 		"hostname-override": host.HostnameOverride,
 		"kubeconfig":        pki.GetConfigPath(pki.KubeProxyCertName),
 	}
+	CommandArrayArgs := map[string][]string{}
```
thedadams: Can we do something similar here as with my previous suggestions to save on unnecessary allocations?

cluster/plan.go
```diff
@@ -786,7 +810,7 @@ func (c *Cluster) BuildSchedulerProcess(host *hosts.Host, serviceOptions v3.Kube
 	CommandArgs := map[string]string{
 		"kubeconfig": pki.GetConfigPath(pki.KubeSchedulerCertName),
 	}
-
+	CommandArrayArgs := map[string][]string{}
```
thedadams:
```suggestion
	CommandArrayArgs := make(map[string][]string, len(c.Services.Scheduler.ExtraArgsArray))
```

cluster/plan.go
```diff
@@ -969,6 +995,7 @@ func (c *Cluster) BuildEtcdProcess(host *hosts.Host, etcdHosts []*hosts.Host, se
 		"peer-cert-file":              pki.GetCertPath(nodeName),
 		"peer-key-file":               pki.GetKeyPath(nodeName),
 	}
+	CommandArrayArgs := map[string][]string{}
```
thedadams:
```suggestion
	CommandArrayArgs := make(map[string][]string, len(c.Services.Etcd.ExtraArgsArray))
```

hosts/hosts.go
```diff
@@ -420,6 +420,15 @@ func (h *Host) GetExtraArgs(service v3.BaseService) map[string]string {
...
+		return service.WindowsExtraArgsArray
+	default:
+		return service.ExtraArgsArray
+	}
+}
```
thedadams: nit: switch is not necessary here.
```suggestion
	if h.OS() == "windows" && len(service.WindowsExtraArgsArray) > 0 {
		return service.WindowsExtraArgsArray
	}

	return service.ExtraArgsArray
}
```

types/zz_generated_deepcopy.go
```diff
@@ -249,6 +249,21 @@ func (in *BaseService) DeepCopyInto(out *BaseService) {
 			(*out)[key] = val
 		}
 	}
+	if in.ExtraArgsArray != nil {
```
thedadams, line 252: `go generate` results should be in a separate commit.

thedadams (changes requested):

types/zz_generated_deepcopy.go
HarrisonWAffel: fixed

thedadams (approved):
Looks good. Please don't squash these commits on merge.

jiaqiluo (approved):
LGTM, 
before merging, could you run the tests again to make sure it still works as expected?

jakefhyde (approved):

PR #2915, opened 2022-04-27
Revert user addon logic for coredns issue 

kinarashah:
The safest option with https://github.com/rancher/rancher/issues/37461 unresolved is to revert changes so we can release RKE v1.3.10 for April k8s patch releases. The fix was originally for https://github.com/rancher/rancher/issues/36985 and during testing we found https://github.com/rancher/rancher/issues/37461.


snasovich (dismissed):

rayandas (approved):
lgtm

snasovich (approved):

PR #2952, opened 2022-06-10
Added retry for snapshot and s3 upload, verify etcd running on host

jakefhyde:
https://github.com/rancher/rancher/issues/37639

- Added option to quiet noisy container logs
  - In other words, some rke operations (such as taking a snapshot and uploading it) produce very few logs, so printing these logs for containers that have a potentially long run time floods the logs and isn't very useless by itself. Now, for taking a snapshot, the stdout and stderr of the container are only logged when they are different from previous iterations.
- Improved logic to ensure that saving a snapshot is considered successful when at least one snapshot is taken, and at least one node was able to upload to s3. The failures are still logged, however previously this would've resulted in snapshots that failed to upload to s3 being deleted from the node (fixed in `rke-tools`), so rke has been updated to reflect this. Additionally, rke will now fall back to using the snapshot on the node. This may change if rancher is to separate local and s3 backups for rke1 like rke2, but will have considerations regarding retention (although more than likely we will track retention separately for local/s3 snapshots much like how we are for failed snapshots (Needs Discussion)). There are already considerations regarding the desync between the rancher etcdbackup objects, and the actual backups in both s3 and on the nodes, and a solutation to backpopulate these objects similar to rke2 may actually be desired here in order to minimize the risk of a user restoring an impossible snapshot.
- Renamed `IsContainerRunning` to `DoesContainerExist` which is what was previously being checked; added `IscontainerRunning` for when you really want to make sure a container is running (like `etcd` before taking a snapshot).

Depends on https://github.com/rancher/rke-tools/pull/153 and subsequent KDM update.


docker/docker.go
```diff
@@ -164,7 +164,7 @@ func DoRollingUpdateContainer(ctx context.Context, dClient *client.Client, image
 		return fmt.Errorf("[%s] Failed rolling update of container: docker client is nil for container [%s] on host [%s]", plane, containerName, hostname)
 	}
 	logrus.Debugf("[%s] Checking for deployed [%s]", plane, containerName)
-	isRunning, err := IsContainerRunning(ctx, dClient, hostname, containerName, false)
+	isRunning, err := DoesContainerExist(ctx, dClient, hostname, containerName, false)
```
jiaqiluo: It is better to make the variable name matches the actual meaning to reduce confusion. 
```suggestion
	exist, err := DoesContainerExist(ctx, dClient, hostname, containerName, false)
```

pki/deploy.go
```diff
@@ -126,7 +126,7 @@ func DeployStateOnPlaneHost(ctx context.Context, host *hosts.Host, stateDownload
 func doRunDeployer(ctx context.Context, host *hosts.Host, containerEnv []string, certDownloaderImage string, prsMap map[string]v3.PrivateRegistry, k8sVersion string) error {
 	// remove existing container. Only way it's still here is if previous deployment failed
 	isRunning := false
-	isRunning, err := docker.IsContainerRunning(ctx, host.DClient, host.Address, CrtDownloaderContainer, true)
+	isRunning, err := docker.DoesContainerExist(ctx, host.DClient, host.Address, CrtDownloaderContainer, true)
```
jiaqiluo: It is better to make the variable name matches the actual meaning to reduce confusion.
```suggestion
	exist, err := docker.DoesContainerExist(ctx, host.DClient, host.Address, CrtDownloaderContainer, true)
```

pki/deploy.go
```diff
@@ -188,7 +188,7 @@ func doRunDeployer(ctx context.Context, host *hosts.Host, containerEnv []string,
 	}
 	logrus.Debugf("[certificates] Successfully started Certificate deployer container: %s", CrtDownloaderContainer)
 	for {
-		isDeployerRunning, err := docker.IsContainerRunning(ctx, host.DClient, host.Address, CrtDownloaderContainer, false)
+		isDeployerRunning, err := docker.DoesContainerExist(ctx, host.DClient, host.Address, CrtDownloaderContainer, false)
```
jiaqiluo, line 190: It is better to make the variable name matches the actual meaning to reduce confusion.
```suggestion
		exist, err := docker.DoesContainerExist(ctx, host.DClient, host.Address, CrtDownloaderContainer, false)
```

pki/deploy.go
```diff
@@ -331,7 +331,7 @@ func FetchFileFromHost(ctx context.Context, filePath, image string, host *hosts.
 		Binds:      Binds,
 		Privileged: true,
 	}
-	isRunning, err := docker.IsContainerRunning(ctx, host.DClient, host.Address, containerName, true)
+	isRunning, err := docker.DoesContainerExist(ctx, host.DClient, host.Address, containerName, true)
```
jiaqiluo: It is better to make the variable name matches the actual meaning to reduce confusion.
```suggestion
	exist, err := docker.DoesContainerExist(ctx, host.DClient, host.Address, containerName, true)
```

services/services.go
```diff
@@ -60,7 +60,7 @@ const (
 type RestartFunc func(context.Context, *hosts.Host) error
 
 func runSidekick(ctx context.Context, host *hosts.Host, prsMap map[string]v3.PrivateRegistry, sidecarProcess v3.Process, k8sVersion string) error {
-	isRunning, err := docker.IsContainerRunning(ctx, host.DClient, host.Address, SidekickContainerName, true)
+	isRunning, err := docker.DoesContainerExist(ctx, host.DClient, host.Address, SidekickContainerName, true)
```
jiaqiluo: It is better to make the variable name matches the actual meaning to reduce confusion.
```suggestion
	exist, err := docker.DoesContainerExist(ctx, host.DClient, host.Address, SidekickContainerName, true)
```

docker/docker.go
```diff
@@ -217,30 +217,52 @@ func DoRemoveContainer(ctx context.Context, dClient *client.Client, containerNam
...
+
+func IsContainerRunning(ctx context.Context, dClient *client.Client, hostname string, containerName string, all bool) (bool, error) {
+	if dClient == nil {
+		return false, fmt.Errorf("Failed to check if container is running: docker client is nil for container [%s] on host [%s]", containerName, hostname)
+	}
+	container, err := FindContainer(ctx, dClient, hostname, containerName, all)
+	if err != nil {
+		return false, fmt.Errorf("Error checking if container [%s] is running on host [%s]: %v", containerName, hostname, err)
+	}
+	return container.State == "running", nil
```
jiaqiluo: This could cause a runtime error in the current implementation because the `FindContainer` function could return "nil, nil"

services/etcd.go
```diff
@@ -449,18 +450,30 @@ func RunEtcdSnapshotSave(ctx context.Context, etcdHost *hosts.Host, prsMap map[s
...
+		}
+
 		if err := docker.DoRunContainer(ctx, etcdHost.DClient, imageCfg, hostCfg, EtcdSnapshotOnceContainerName, etcdHost.Address, ETCDRole, prsMap); err != nil {
 			return err
 		}
-		status, _, stderr, err := docker.GetContainerOutput(ctx, etcdHost.DClient, EtcdSnapshotOnceContainerName, etcdHost.Address)
+		status, _, stderr, err := docker.GetContainerOutput(ctx, etcdHost.DClient, EtcdSnapshotOnceContainerName, etcdHost.Address, false)
 		if status != 0 || err != nil {
 			if removeErr := docker.RemoveContainer(ctx, etcdHost.DClient, etcdHost.Address, EtcdSnapshotOnceContainerName); removeErr != nil {
 				log.Warnf(ctx, "[etcd] Failed to remove container [%s] on host [%s]: %v", removeErr, etcdHost.Address)
```
jiaqiluo: at the line start with "log.Warnf(", expect 3 arguments but only 2 provided.

jiaqiluo (changes requested):

docker/docker.go
jakefhyde: Good catch

services/etcd.go
jakefhyde: Very good catch, syntax highlighter doesn't even pick this up (probably because first parameter isn't a string)

cluster/etcd.go
```diff
@@ -17,16 +18,52 @@ import (
...
 	backupImage := c.getBackupImage()
+	containerTimeout := DefaultEtcdBackupConfigTimeout
+	if c.Services.Etcd.BackupConfig != nil && c.Services.Etcd.BackupConfig.Timeout > 0 {
+		containerTimeout = c.Services.Etcd.BackupConfig.Timeout
+	}
+
+	// store first error message
+	var snapshotErr error
+	snapshotFailures := 0
+	s3Failures := 0
```
a-blender: Nit: This specifies the count of failures to upload an etcd snapshot to s3. At first glance I thought, what kind of s3 failure is this? Rename to `s3UploadFailures`

services/etcd.go
```diff
@@ -449,18 +450,30 @@ func RunEtcdSnapshotSave(ctx context.Context, etcdHost *hosts.Host, prsMap map[s
 		if err := docker.DoRemoveContainer(ctx, etcdHost.DClient, EtcdSnapshotOnceContainerName, etcdHost.Address); err != nil {
 			return err
 		}
+
+		// If the etcd container is not running the snapshot will never succeed
+		log.Debugf(ctx, "[etcd] Checking if etcd is running on host [%s]", etcdHost.Address)
+		if running, err := docker.IsContainerRunning(ctx, etcdHost.DClient, etcdHost.Address, "etcd", true); err != nil {
```
a-blender, line 454: Update to

```
// If the etcd container does not exist the snapshot will never succeed
if exist, err := docker.doesContainerExist...
```

annablender (changes requested):

services/etcd.go
jakefhyde: This is actually intended. The etcd container has to be running or the snapshot will fail (that was actually the original code I had, that didn't work).

annablender:
Super nit: In the desc, `printing these logs for containers that have a potentially long run time floods the logs and isn't very useless by itself` did you mean helpful?

services/etcd.go
a-blender: This appears to be the only case where updating `IsContainerRunning` to `doesContainerExist` doesn't make sense. Maybe still update the var and func name, and leave the original comment about the container needing to be running? That's my 2c

jakefhyde: There may be some confusion around this so let me clarify: The original function named `IsContainerRunning` wasn't actually checking if the container was running, rather that it existed. In this case we want to make sure the container itself is running (not stopped, or killed, or starting, etc.) so I renamed the previous function to what it actually does, and introduced a new function `IsContainerRunning` that actually checks that the container does have the state `running`.

jiaqiluo (approved):

jiaqiluo:
LGTM. Please squash commits before merging

HarrisonWAffel (approved):

annablender (approved):

PR #2955, opened 2022-06-13
Add logic to fix coreDNS addon config issue.

rayandas:
According to https://github.com/rancher/rancher/issues/36985#issuecomment-1139800798, this PR is to revert https://github.com/rancher/rke/pull/2915
Part of: https://github.com/rancher/rancher/issues/36985


rayandas:
As per https://github.com/rancher/rancher/issues/37461#issuecomment-1155158664 we can say that the issue 'CoreDNS pod is not getting scheduled after the cluster upgrade' is not related to these (current PR) changes. So this can go in. 
@kinarashah @snasovich

kinarashah (approved):

rayandas:
Merging this.

PR #2986, opened 2022-07-19
always set kubelet node-ip to internal address

hensur:
Currently, there is no way to set the kubelet `node-ip` if `host.Address` is the same as `host.InternalAddress`.

Shouldn't it be possible to always set the kubelet `node-ip` to the internal address, if that one is set?

In my case, the node has two interfaces, one for the default route and one for the internal cluster address. However, the internal cluster address is the one used for rke provisioning as well. Therefore, the host address and the internal address are equal.
By default, the kubelet will select the address from the default route interface and there seems to be no way to override this one.


hensur:
I just noticed that this won't work, since InternalAddress gets set to host.Address (possibly a hostname) if it isn't set.
Once this happens, the kubelet will (thankfully) ignore the wrongly formatted `--node-ip` flag. This shouldn't happen, I'm working on a fix for this.

PR #3001, opened 2022-08-03
Add bind mount for `/var/lib/cri-dockerd`

Oats87:
https://github.com/rancher/rke/issues/2993

Issue #2993, opened 2022-07-27
`/var/lib/dockershim` and `/var/lib/cri-dockerd` should not be ephemeral


Oats87:
**RKE version:**
RKE <= 1.3.12? 
**Docker version: (`docker version`,`docker info` preferred)**
Doesn't matter
**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)**
Doesn't matter
**Type/provider of hosts: (VirtualBox/Bare-metal/AWS/GCE/DO)**
Doesn't matter
**cluster.yml file:**
Doesn't matter
**Steps to Reproduce:**
The `/var/lib/dockershim` (`/var/lib/cri-dockerd` for >= 1.24.x) directory for the `kubelet` container should not be ephemeral. The kubelet/dockershim relies on this directory to store runtime metadata such as `portMappings` for management of pods. 

We ran into this issue when debugging a case where hostport pods were not being properly cleaned up on tear down (with duplicate iptables rules being found) after Kubernetes version upgrades.

This issue was specifically a combination of multiple problems, namely one where we updated the CNI plugin versions from `v0.8.6` to `v1.0.0` which brought along a change (see: https://github.com/containernetworking/plugins/pull/509) that changed to performing a `NOOP` if `portMappings` were not specified. The issue we saw was that the kubelet/dockershim were sending an empty `portMappings` after a Kubernetes version upgrade, as the Kubelet would be replaced and the `/var/lib/dockershim` folder was empty, and could not properly construct a `portMappings` struct for teardown.

A workaround for this issue is to specify an `extra_bind` for the `kubelet` service, i.e. 
```
services:
  kubelet:
    extra_binds:
      - "/var/lib/dockershim:/var/lib/dockershim"
```
and rebooting the host after an `rke up` (if already dealing with duplicate iptables rules).

This will not be done automatically for `v1.23` and older versions of Kubernetes, but will be done for `v1.24` and newer versions of K8s.

SURE-4702

sowmyav27:
**Reproduced the issue on Rancher 2.6head commit id: `78017e64e`**

On an upgrade of RKE1 node driver Linode cluster (1 etcd/cp, 2 worker nodes) from `1.20.15-rancher2-1` to `1.21.14-rancher1-1` to `1.22.11-rancher1-1`, ingress after upgrading to `1.22.11-rancher1-1` doesn't work. Works after upgrade to `1.21.14-rancher1-1`

Other useases tested:
[...]

PR #3011, opened 2022-08-11
use new container runtime endpoint for cri-dockerd >=v0.2.4

kinarashah:
**Problem**
We bumped cri-dockerd to v0.2.4 here https://github.com/rancher/rke-tools/pull/155, which uses the new dockerd endpoint - https://github.com/Mirantis/cri-dockerd/blob/v0.2.4/cmd/cri/options/options.go#L45. 
```
   remoteRuntimeEndpoint = "unix:///var/run/cri-dockerd.sock"
```
This new cri-dockerd version is used not only by k8s 1.24, but also by latest patch k8s versions via latest `rke-tools:v0.1.86`: 
- v1.23.8-rancher1-1
- v1.22.11-rancher1-1
- v1.21.14-rancher1-1 

Kubelet on k8s versions currently fails to start because RKE sends old endpoint `"/var/run/dockershim.sock"`. 

**Solution**
Adding range checks for these versions and setting container runtime endpoint accordingly. 

**Issue** 
https://github.com/rancher/rancher/issues/38594

### DIFF
--- a/cluster/plan.go
+++ b/cluster/plan.go
@@ -509,8 +509,8 @@ func (c *Cluster) BuildKubeletProcess(host *hosts.Host, serviceOptions v3.Kubern
 		if err != nil {
 			logrus.Debugf("Error while parsing cluster version: %s", err)
 		}
-		// cri-dockerd must be enabled if the cluster version is 1.24 and higher
-		if parsedRangeAtLeast124(parsedVersion) {
+
+		if useNewCRIEndpoint(c.Version, parsedVersion) {
 			CommandArgs["container-runtime-endpoint"] = "unix:///var/run/cri-dockerd.sock"
 			Binds = []string{fmt.Sprintf("%s:/var/lib/cri-dockerd:z", path.Join(host.PrefixPath, "/var/lib/cri-dockerd"))}
 		}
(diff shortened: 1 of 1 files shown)


cluster/plan.go
```diff
@@ -509,8 +509,8 @@ func (c *Cluster) BuildKubeletProcess(host *hosts.Host, serviceOptions v3.Kubern
 		if err != nil {
 			logrus.Debugf("Error while parsing cluster version: %s", err)
 		}
-		// cri-dockerd must be enabled if the cluster version is 1.24 and higher
-		if parsedRangeAtLeast124(parsedVersion) {
+
+		if useNewCRIEndpoint(c.Version, parsedVersion) {
```
jiaqiluo, line 513: the variable `parsedVersion` is used only in this function; we can move lines 508-511 into the function `useNewCRIEndpoint`
```suggestion
		if useNewCRIEndpoint(c.Version) {
```

cluster/plan.go
```diff
@@ -1283,3 +1283,22 @@ func (c *Cluster) IsCRIDockerdEnabled() bool {
 	}
 	return false
 }
+
+func useNewCRIEndpoint(clusterVersion string, parsedVersion semver.Version) bool {
```
jiaqiluo, line 1287:
```suggestion
func useNewCRIEndpoint(clusterVersion string) bool {
		parsedVersion, err := getClusterVersion(c.Version)
		if err != nil {
			logrus.Debugf("Error while parsing cluster version: %s", err)
		}
```

cluster/plan.go
```diff
@@ -509,8 +509,8 @@ func (c *Cluster) BuildKubeletProcess(host *hosts.Host, serviceOptions v3.Kubern
 		if err != nil {
 			logrus.Debugf("Error while parsing cluster version: %s", err)
 		}
-		// cri-dockerd must be enabled if the cluster version is 1.24 and higher
-		if parsedRangeAtLeast124(parsedVersion) {
+
+		if useNewCRIEndpoint(c.Version, parsedVersion) {
 			CommandArgs["container-runtime-endpoint"] = "unix:///var/run/cri-dockerd.sock"
 			Binds = []string{fmt.Sprintf("%s:/var/lib/cri-dockerd:z", path.Join(host.PrefixPath, "/var/lib/cri-dockerd"))}
```
jiaqiluo, line 515: So the fix for the following issue will also be available on the latest 1.21 1.22 1.23, right? Do we want to do it? My concern is it will require extra validations from QA. 
https://github.com/rancher/rke/issues/2993 ( its PR https://github.com/rancher/rke/pull/3001)

jiaqiluo (commented):
Just notice the PR was closed right before submitting my comments. But anyway, since I have finished the typing, will still submit it.

PR #3038, opened 2022-09-15
[release/v1.3] Add new 1.19.16 template to bring stonex changes.

rayandas:


kinarashah (changes requested):
This PR shouldn't show anything else other than KDM changes, could you 
rebase off of latest`upstream/release/v1.3` and update `codegen/codegen.go` to change the URL so that data gets generated from `dev-v2.6` and run `go generate`? 

Also, it's removing the min version info for `v2.6.3-patch3` which is not correct. This is because the change from `release-v2.6` wasn't ported to `dev-v2.6`. So could you open a PR in KDM first to cherry-pick https://github.com/rancher/kontainer-driver-metadata/pull/932 changes and then update RKE PR to have the latest changes?

rayandas:
@kinarashah Sure. On it.

kinarashah (approved):

PR #3085, opened 2022-11-04
support K8S 1.25 and PSA

jiaqiluo:


cluster/cluster.go
```diff
@@ -374,11 +376,11 @@ func parseAuditLogConfig(clusterFile string, rkeConfig *v3.RancherKubernetesEngi
 	if services["kube-api"] == nil {
 		return nil
 	}
-	kubeapi := services["kube-api"].(map[string]interface{})
-	if kubeapi["audit_log"] == nil {
+	kubeAPI := services["kube-api"].(map[string]interface{})
```
sbstp, line 379: Not sure how the team feels about this, but since Go 1.18, a lot of people just use `any` instead of `interface{}`. Since the PR also sets the minimum Go version to 1.19 it should be safe to change it

a-blender: I'm fine with this. Just read up on it a bit, `any/interface{}` are interchangeable as generic types. So I guess this (and the other instances) can be updated to

```
kubeAPI := services["kube-api"].(map[string]any)
```

cluster/cluster.go
```diff
@@ -405,6 +407,39 @@ func parseAuditLogConfig(clusterFile string, rkeConfig *v3.RancherKubernetesEngi
 	return err
 }
 
+func parseEventRateLimit(clusterFile string, rkeConfig *v3.RancherKubernetesEngineConfig) error {
+	if rkeConfig.Services.KubeAPI.EventRateLimit == nil || !rkeConfig.Services.KubeAPI.EventRateLimit.Enabled {
+		return nil
+	}
+	logrus.Debugf("event rate limit is found in cluster.yml")
+	var r map[string]interface{}
```
a-blender: 1. Update to `map[string]any`
2. `r` should have a more descriptive name like `eventRateLimit` or `rateLimit`

cluster/hosts.go
```diff
@@ -270,21 +281,19 @@ func (c *Cluster) SetUpHosts(ctx context.Context, flags ExternalFlags) error {
...
-				bytes, err := yaml.Marshal(ac)
-				if err != nil {
-					return err
-				}
-				if err := deployFile(ctx, controlPlaneHosts, c.SystemImages.Alpine, c.PrivateRegistriesMap, DefaultKubeAPIArgAdmissionControlConfigFileValue, string(bytes), c.Version); err != nil {
-					return err
-				}
-				log.Infof(ctx, "[%s] Successfully deployed admission control config to Cluster control nodes", DefaultKubeAPIArgAdmissionControlConfigFileValue)
+			controlPlaneHosts := hosts.GetUniqueHostList(nil, c.ControlPlaneHosts, nil)
+			ac, err := c.getConsolidatedAdmissionConfiguration()
```
a-blender: Nit: I'd call this variable `admissionConfig` not ac, long as you aren't shadowing

cluster/hosts.go
```diff
@@ -270,21 +281,19 @@ func (c *Cluster) SetUpHosts(ctx context.Context, flags ExternalFlags) error {
...
+			if err != nil {
+				return fmt.Errorf("error getting consolidated admission configuration: %v", err)
 			}
+			bytes, err := yaml.Marshal(ac)
+			if err != nil {
+				return err
+			}
+			if err := deployFile(ctx, controlPlaneHosts, c.SystemImages.Alpine, c.PrivateRegistriesMap, DefaultKubeAPIArgAdmissionControlConfigFileValue, string(bytes), c.Version); err != nil {
```
a-blender: Probably more of a personal preference, but this syntax with the if at the front and `err != nil` as a secondary is confusing, I'd move the err check to a second line since the lines are so long but maybe that's just me

cluster/hosts.go
```diff
@@ -332,3 +341,18 @@ func removeFromRKENodes(nodeToRemove v3.RKEConfigNode, nodeList []v3.RKEConfigNo
 	}
 	return l
 }
+
+// setPluginConfiguration either adds the plugin configuration or replaces the existing one in the admission configuration
+func setPluginConfiguration(admissionConfig *apiserverv1.AdmissionConfiguration, pluginConfig apiserverv1.AdmissionPluginConfiguration) error {
+	if admissionConfig == nil {
+		return fmt.Errorf("admission configuarion does not exist")
+	}
+	for i, plugin := range admissionConfig.Plugins {
```
a-blender, line 351: Will there ever be no `admissionConfig.Plugins?` Nil check?

cluster/plan.go
```diff
@@ -705,7 +708,7 @@ func (c *Cluster) BuildKubeProxyProcess(host *hosts.Host, serviceOptions v3.Kube
 		services.SidekickContainerName,
 	}
 
-	//TODO: we should reevaluate if any of the bind mounts here should be using read-only mode
+	// TODO: we should reevaluate if any of the bind mounts here should be using read-only mode
```
a-blender, line 711: Is this TODO finished or should be included in this PR?

cluster/reconcile.go
```diff
@@ -488,3 +489,18 @@ func getTaintKey(taint v3.RKETaint) string {
 func getTaintValue(taint v3.RKETaint) string {
 	return fmt.Sprintf("%s=%s:%s", taint.Key, taint.Value, taint.Effect)
 }
+
+// RestartKubeAPIServerWhenConfigChanges restarts the kube-apiserver container on the control plan nodes
```
a-blender: Nit: control `plane`

cluster/validation.go
```diff
@@ -672,6 +677,29 @@ func validatePSP(c *Cluster) error {
...
 
+func validatePodSecurity(c *Cluster) error {
+	parsedVersion, err := getClusterVersion(c.Version)
+	if err != nil {
+		logrus.Warnf("Failed to parse semver range for validating Pod Security")
+		return err
+	}
+	logrus.Debugf("Checking PodSecurity for cluster version [%s]", c.Version)
+	level := c.Services.KubeAPI.PodSecurityConfiguration
+	if len(level) != 0 {
```
a-blender, line 692: Could you add a comment saying why checking the level of the PodSecurityConfiguration is important for those who are less familiar with PSP?

cluster/hosts.go
jiaqiluo: I can see your concern. 
There will be no error or panic happening when the `admissionConfig.Plugins` is nil, instead, the loop will do nothing silently.

cluster/plan.go
jiaqiluo: I did not add this TODO, it has been there for a long time. I will keep it for now, I may address it in another PR.

cluster/cluster.go
jiaqiluo: thank you for the advice 😄 
I will keep using the old fashion this time and leave the refactoring to another PR.

cluster/cluster.go
jiaqiluo: 1. I will keep using the old fashion this time and leave the refactoring to another PR.
2. the name is updated

cluster/hosts.go
jiaqiluo: fixed

cluster/hosts.go
jiaqiluo: fixed

cluster/reconcile.go
jiaqiluo: fixed

cluster/validation.go
jiaqiluo: sure, comment is added

cluster/plan.go
```diff
@@ -173,24 +177,25 @@ func (c *Cluster) BuildKubeAPIProcess(host *hosts.Host, serviceOptions v3.Kubern
...
-		"proxy-client-cert-file":       pki.GetCertPath(pki.APIProxyClientCertName),
-		"proxy-client-key-file":        pki.GetKeyPath(pki.APIProxyClientCertName),
-		"requestheader-allowed-names":  pki.APIProxyClientCertName,
-		"requestheader-client-ca-file": pki.GetCertPath(pki.RequestHeaderCACertName),
-		"service-account-key-file":     pki.GetKeyPath(pki.ServiceAccountTokenKeyName),
-		"service-cluster-ip-range":     c.Services.KubeAPI.ServiceClusterIPRange,
-		"service-node-port-range":      c.Services.KubeAPI.ServiceNodePortRange,
-		"tls-cert-file":                pki.GetCertPath(pki.KubeAPICertName),
-		"tls-private-key-file":         pki.GetKeyPath(pki.KubeAPICertName),
+		"admission-control-config-file": DefaultKubeAPIArgAdmissionControlConfigFileValue,
```
Oats87, line 180: Will this not cause a redeploy of the `kube-apiserver` pod anytime someone upgrades to a version of RKE that contains this change?

scripts/validate
```diff
@@ -6,6 +6,7 @@ cd $(dirname $0)/..
 echo Running validation
 
 echo Running: golangci-lint
+export "GOROOT=$(go env GOROOT)"
```
Oats87, line 9: Is this because of the go 1.19 bump?

Dockerfile.dapper
```diff
@@ -11,7 +11,7 @@ RUN apt-get update && \
 ENV GOLANG_ARCH_amd64=amd64 GOLANG_ARCH_arm=armv6l GOLANG_ARCH_arm64=arm64 GOLANG_ARCH=GOLANG_ARCH_${ARCH} \
     GOPATH=/go PATH=/go/bin:/usr/local/go/bin:${PATH} SHELL=/bin/bash
 
-RUN wget -O - https://storage.googleapis.com/golang/go1.17.6.linux-${!GOLANG_ARCH}.tar.gz | tar -xzf - -C /usr/local
+RUN wget -O - https://storage.googleapis.com/golang/go1.19.3.linux-${!GOLANG_ARCH}.tar.gz | tar -xzf - -C /usr/local
```
Oats87, line 14: Is this explicitly required for 1.25?

Oats87 (changes requested):

cluster/plan.go
Oats87: I am mostly concerned about this, as it has large implications for upgrades of Rancher.

scripts/validate
jiaqiluo: yep.

Dockerfile.dapper
jiaqiluo: yes, otherwise, Dapper fails to compile RKE.

cluster/plan.go
jiaqiluo: @Oats87 
Good point!
If  it is true that Rancher will run the reconciliation cycle against all downstream clusters after upgrading, I think, yes, the `kube-apiserve` pod will be redeployed. 
To me, it looks like something naturally happening based on how Rancher works today. I guess this is not the first we run into such a situation. Do you know any way to avoid redeployment? I also wonder if it is a good idea to add special logic to handle it in Rancher.

jiaqiluo: update:
This is confirmed offline that Rancher does NOT run the reconciliation cycle against all downstream RKE clusters after upgrading. However, Rancher updates the plan on the worker nodes after upgrading. I need to make sure that the downstream cluster is still working file in this case.

cluster/cluster.go
a-blender: Sure, that's fine -- thank you!

annablender (approved):
Updates LGTM thank you @jiaqiluo !

sbstp (approved):
LGTM but I feel like I'm not qualified yet to properly vet a big change like this

PR #3113, opened 2022-11-30
[2.6] added check for empty updatestrategy while setting dns addon defaults

vardhaman22:
fix for issue: https://github.com/rancher/rancher/issues/39729


snasovich (approved):
Same as https://github.com/rancher/rke/pull/3087, LGTM. One approval is enough for straight backports, so this is good to merge now.

PR #3132, opened 2022-12-28
add checks for PSP resources in the cluster

jiaqiluo:


cluster/validation.go
```diff
@@ -674,6 +676,30 @@ func validatePodSecurityPolicy(c *Cluster) error {
 			return errors.New("PodSecurityPolicy has been removed and can not be enabled since k8s v1.25")
 		}
 	}
+	// check if there is any PSP resource when upgrading a cluster to k8s v1.25 and above
+	if parsedRangeAtLeast125(parsedVersion) {
+		kubeClient, err := k8s.NewClient(c.LocalKubeConfigPath, c.K8sWrapTransport)
+		if err != nil {
+			// we can not tell this is invoked when creating a new cluster or updating an existing one, so skip this check
+			logrus.Debugf("Skip the check for PSP resource due to the failire of initializing the kubernetes client")
```
HarrisonWAffel: `failire` -> `failure`

cluster/validation.go
```diff
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/rancher/rke/k8s"
+
```
HarrisonWAffel: We should remove the space between imports on line 10

HarrisonWAffel (changes requested):
just two nits, otherwise lgtm

cluster/validation.go
```diff
@@ -674,6 +676,30 @@ func validatePodSecurityPolicy(c *Cluster) error {
...
 	}
+	// check if there is any PSP resource when upgrading a cluster to k8s v1.25 and above
+	if parsedRangeAtLeast125(parsedVersion) {
+		kubeClient, err := k8s.NewClient(c.LocalKubeConfigPath, c.K8sWrapTransport)
+		if err != nil {
+			// we can not tell this is invoked when creating a new cluster or updating an existing one, so skip this check
+			logrus.Warnf("Skip the check for PSP resource due to the failire of initializing the kubernetes client")
+			return nil
+		}
+		pspList, _ := k8s.GetPSPList(kubeClient)
```
a-blender, line 686: Could pspList be nil? Why are we not returning a possible error? I'd suggest,

```
pspList, err := k8s.GetPSPList(kubeClient)
if err != nil {
    return errors.New("Error getting PodSecurityPolicy list for cluster")
}
if pspList == nil {
    return errors.New("PodSecurityPolicy list for cluster resource is invalid")
}
```

cluster/validation.go
```diff
@@ -674,6 +676,30 @@ func validatePodSecurityPolicy(c *Cluster) error {
...
+		// ignore the error because the "no such resource type" error is definitely returned in k8s v1.25 and above
+		items := pspList.Items
+		if len(items) == 0 {
+			return nil
+		}
+		// a PSP "psp.flannel.unprivileged" from old Flannel templates is created when using Flannel as the network plugin
+		// we should ignore it if it is the only one PSP in the cluster
```
a-blender: Nit: `only one PSP in the cluster` -> `only PSP in the cluster`

annablender (changes requested):

cluster/validation.go
jiaqiluo: I also thought about this and checked  [the source code](https://github.com/kubernetes/client-go/blob/v0.25.4/kubernetes/typed/policy/v1beta1/podsecuritypolicy.go#L86), it looks like an "empty" pointer is always returned even when nothing is found.

cluster/validation.go
jiaqiluo: fixed.

cluster/validation.go
jiaqiluo: fixed.

cluster/validation.go
jiaqiluo: fixed.

HarrisonWAffel (approved):

annablender (approved):
Squash and merge!

PR #3169, opened 2023-02-14
[release/v1.3.20] Add missing calico logs volumeMount 

doflamingo721:
There was a missing volumeMount which was causing the calico/canal logs getting written to the container overlay filesystem causing the container to fill up. Adding the missing bind resolves this issue.

Linked issue: https://github.com/rancher/rancher/issues/40540
SURE: 5814


snasovich (dismissed):
LGTM on the surface, please add to the description on the PR that it also brings changes from https://github.com/rancher/kontainer-driver-metadata/pull/1077 to RKE.

Still need @kinarashah's approval before merging.

doflamingo721:
@kinarashah As discussed offline, I have added the k8s version constraints so that a existing cluster with a older k8s version than the latest patch won't be auto-upgraded as per the new plan after adding a worker node. Can you please review these changes?

PR #3174, opened 2023-02-20
[release/v1.4] Move to BCI

superseb:
- Moves to BCI for building
- Bumps docker-dind image used for testing
- Removes creating a Docker image for RKE as we are not publishing/using it
- Removes the race flag for testing as it requires an additional package in BCI and we don't use it anywhere else, if we would like to keep it, we can add the package (which then need to be kept up-to-date as well when we bump Go)

This also prepares the repository so we can keep it up-to-date automatically with Renovate in the future.

We can also postpone this PR to 1.5.x if wanted.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -9,7 +9,7 @@ platform:
 steps:
 - name: ci
   pull: default
-  image: rancher/dapper:1.11.2
+  image: rancher/dapper:v0.6.0
   commands:
   - dapper ci
   - ls -lR build/bin
@@ -23,18 +23,6 @@ steps:
     event:
     - pull_request
     - tag
-
-- name: stage-binaries
-  pull: default
-  image: rancher/dapper:1.11.2
-  commands:
-  - "cp -r ./bin/* ./package/"
-  when:
-    event:
-    - tag
-    ref:
-      include:
-      - "refs/tags/*"
 
 - name: github_binary_prerelease
   pull: default
@@ -117,7 +105,7 @@ platform:
 steps:
 - name: build
   pull: default
-  image: rancher/dapper:1.11.2
+  image: rancher/dapper:v0.6.0
   commands:
   - dapper build
   privileged: true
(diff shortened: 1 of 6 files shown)


Dockerfile.dapper
```diff
@@ -1,26 +1,11 @@
-FROM ubuntu:20.04
-# FROM arm=armhf/ubuntu:20.04 arm64=arm64v8/ubuntu:20.04
+FROM registry.suse.com/bci/golang:1.19-20.16
```
a-blender, line 1: What testing have you run to make sure this works? Local run of a `dapper make` or `dapper ci`? Does this update still give access to the same packages as ubuntu 20.04? I also wonder if BCI will be lighter weight too since it's not creating a Docker image. That's always ideal

cc @kinarashah

superseb: Yes, basically the same command Drone kicks off. The only package I needed to add was `awk` as its used in `scripts/integration`.

jiaqiluo (dismissed):
Good job! 
LGTM

a-blender (approved):

manuelbuil (dismissed):

thomasferrandiz (approved):

.drone.yml
```diff
@@ -26,7 +26,7 @@ steps:
 
 - name: stage-binaries
   pull: default
-  image: rancher/dapper:1.11.2
+  image: rancher/dapper:v0.6.0
   commands:
   - "cp -r ./bin/* ./package/"
```
mitulshah-suse: @superseb this needs to be fixed as well. In 1.5 where it is merged, CI for tags is failing due to this step. The package dir is deleted and its trying to copy files into that dir.

superseb: done, thanks!

manuelbuil (approved):

rbrtbnfgl (approved):

Issue #3083, opened 2022-11-03
[v1.4] Support certificate chains when using custom certs 


kinarashah:
This issue is to forward port https://github.com/rancher/rke/issues/1834. PR for this has been merged to `release/v1.3` branch - https://github.com/rancher/rke/pull/2742. Need to forward port it.

TheSp1der:
Can we re-open this? Having the ability to include intermediate certificates would be a huge benefit when using internal company issued certificates.

TheSp1der:
This is still relivant and would be resolved by merging [PR#3213](https://github.com/rancher/rke/pull/3213).

TheSp1der:
This is still relevant and would be resolved by merging https://github.com/rancher/rke/pull/3213.

TheSp1der:
This is still relevant and would be resolved by merging https://github.com/rancher/rke/pull/3213.

TheSp1der:
@kinarashah Thank you, I had given up hope for this PR.

PR #3213, opened 2023-04-06
[DNM] Support certificate chains when using custom certs.

TheSp1der:
This PR is a duplicate of PR [#2742](https://github.com/rancher/rke/pull/2742) but directed to the 1.4 release branch. This closes [#3083](https://github.com/rancher/rke/issues/3083).

I can not take credit for the original work for this pull request which was completed by [chaudhryfaisal](https://github.com/chaudhryfaisal).


TheSp1der:
This pull request is still valid.

jiaqiluo (approved):
LGTM

pki/util.go
```diff
@@ -636,18 +645,17 @@ func getOUName(certName string) string {
 	}
 }
 
-func getCertFromFile(certDir string, fileName string) (*x509.Certificate, error) {
-	var certificate *x509.Certificate
+func getCertsFromFile(certDir string, fileName string) ([]*x509.Certificate, error) {
```
bfbachmann, line 648: I'm guessing we're ignoring the error here because it's expected, but I'm wondering if we should actually check it in case it's something unexpected. Either that or we can remove the `error` return value since it's always `nil`. Sorry, I know this isn't your code and you're not really changing anything here, but I figure we might want to take this opportunity to fix it.

bfbachmann (approved):

pki/util.go
jiaqiluo: There is one case in the function where we return a real error (lines 653-655). I guess GH collapses those lines?

TheSp1der: I'm willing to support any changes that may be necessary, but from an initial glance this appears to be correct. As @jiaqiluo identified we do return an error when we can't parse the certificates from the file.
Are there any objections to me closing off this conversation?

jiaqiluo:
**Do not merge the PR until the REK v1.4.5 is released.**

TheSp1der:
Can this still be merged?

TheSp1der:
This is still relevant.

PR #3217, opened 2023-04-07
Add support for custom certificate life/expiration.

TheSp1der:
This pull request adds support to RKE to allow the configuration of certificate lifetime/expiration to support typical corporate requirements for shorter certificate expiration.

This change acknowledges that the default certificate lifetime for cluster certificates is ~10 years and makes no change to that default. Instead it adds support for a `cluster.yml` configuration option called `certificate_lifetime`. The `certificate_lifetime` option accepts an integer being a number of days from the current date for which certificates will be valid. This logic is respected during both the creation of a new cluster or during certificate rotation. This modification does not make any changes to the CA's created by RKE.


TheSp1der:
This pull request is still valid.

PR #3218, opened 2023-04-13
[DNM - for 2.7.x Q2 release] Add missing calico logs volumeMount

doflamingo721:
There was a missing volumeMount which was causing the calico/canal logs getting written to the container overlay filesystem causing the container to fill up. Adding the missing bind resolves this issue.

Linked issue: https://github.com/rancher/rancher/issues/40330

PR also vendors KDM `dev-v2.7-2023-04-patches` data.

SURE: 5814


doflamingo721:
@kinarashah I have updated k8s version constraints to latest patches getting released with April patch

cluster/plan.go
```diff
@@ -558,6 +560,26 @@ func (c *Cluster) BuildKubeletProcess(host *hosts.Host, serviceOptions v3.Kubern
 			fmt.Sprintf("%s:c:/host/run", path.Join(host.PrefixPath, "/run")),
 		}...)
 	} else {
+		parsedVersion, err := getClusterVersion(c.Version)
+		if err != nil {
+			logrus.Debugf("Error while parsing cluster version: %s", err)
```
snasovich, line 565: Not a fan of "shutting up" this error message, but I see it's an established pattern in this file.
So nothing to address, just sharing my thoughts here for completeness of the review.

cluster/plan.go
```diff
@@ -558,6 +560,26 @@ func (c *Cluster) BuildKubeletProcess(host *hosts.Host, serviceOptions v3.Kubern
...
+		}
+		parsedJanK8s123Version, err := getClusterVersion(JanK8s123Version)
```
snasovich, line 567: I'm missing context about the discussion with @kinarashah referenced [here](https://github.com/rancher/rke/pull/3169#issuecomment-1445654373), but can we actually include `v1.23.16-rancher1` here as it's already released? (Actual released version is `v1.23.16-rancher2-1` btw)

Note that we're planning to add `v1.23.16-rancher2-2` per https://github.com/rancher/kontainer-driver-metadata/pull/1110, so not sure if we can use that instead for checks.

cluster/plan.go
```diff
@@ -575,6 +597,7 @@ func (c *Cluster) BuildKubeletProcess(host *hosts.Host, serviceOptions v3.Kubern
 			"/dev:/host/dev:rprivate",
 			"/var/log/containers:/var/log/containers:z",
 			"/var/log/pods:/var/log/pods:z",
+			"/var/log/calico/cni:/var/log/calico/cni:z",
```
snasovich: We're conditionally adding what appears to be the same bind. Should it be removed from here then?

cluster/plan.go
```diff
@@ -52,13 +52,15 @@ const (
...
-
+	MaxEtcdOldEnvVersion             = "v3.2.99"
+	MaxK8s115Version                 = "v1.15"
+	MaxEtcdPort4001Version           = "v3.4.3-rancher99"
+	MaxEtcdNoStrictTLSVersion        = "v3.4.14-rancher99"
+	MaxK8s121Version                 = "v1.21.99-rancher99"
+	MaxK8s122Version                 = "v1.22.99-rancher99"
+	JanK8s123Version                 = "v1.23.16-rancher1"
+	AprilK8s124Version               = "v1.24.13-rancher1"
+	AprilK8s125Version               = "1.25.9-rancher1"
```
snasovich: Actual version added/to be relased are `v1.24.13-rancher1-1` / `v1.25.9-rancher1-1`

snasovich: For consistency we want to add `v` here as well.
```suggestion
	AprilK8s125Version               = "v1.25.9-rancher1"
```

cluster/plan.go
```diff
@@ -558,6 +560,26 @@ func (c *Cluster) BuildKubeletProcess(host *hosts.Host, serviceOptions v3.Kubern
...
+		}
+		parsedAprilK8s124Version, err := getClusterVersion(AprilK8s124Version)
+		if err != nil {
+			logrus.Debugf("Error while parsing cluster version: %s", err)
+		}
+		parsedAprilK8s125Version, err := getClusterVersion(AprilK8s125Version)
+		if err != nil {
+			logrus.Debugf("Error while parsing cluster version: %s", err)
+		}
+		if parsedVersion.GE(parsedJanK8s123Version) && parsedVersion.GE(parsedAprilK8s124Version) && parsedVersion.GE(parsedAprilK8s125Version) {
```
snasovich: I don't believe this logic is correct. For example, any `1.24.x` version will result in `false` returned by `parsedVersion.GE(parsedAprilK8s125Version)` and since all parts are AND-ed will result in `false` overall.
What we need here is to check ranges similar to how we do in KDM for applying different templates, e.g. checking against ranges like `>=1.24.13-rancher1-1 <1.25.0-rancher1-1` with all checks OR-ed.

snasovich (changes requested):
Added a few comments to address.

However, since the fix is in `rke` repo, is my understanding correct that the issue will be fixed for standalone RKE1 only and not for Rancher-provisioned downstream RKE1 clusters until we vendor in `rke` with this fix into `rancher/rancher`. If that's the case we can fix RKE1 in upcoming RKE1 release but the validation for Rancher side will require to be done for next Rancher release.
@kinarashah , please review and confirm.

cluster/plan.go
doflamingo721: Rectified the incorrect version.

doflamingo721: Added the actual versions `v1.23.16-rancher1-1`, `v1.24.13-rancher1-1` and `v1.25.9-rancher1-1`

cluster/plan.go
doflamingo721: Yes that is correct. Seems like it was added by mistake. Removed the bind.

cluster/plan.go
doflamingo721: Corrected the logic by adding OR operator in place of AND.

snasovich: @doflamingo721 , this is not sufficient to just "OR" the existing conditions as after this change any `1.24.x` version will return `true` because it's greater than `v1.23.16-rancher1-1`.
The logic we need is "any 1.23.x version that is greater or equal than v1.23.16-rancher1-1 OR any 1.24.x version that is greater or equal than v1.24.13-rancher1 OR any 1.25.x version that is greater or equal than v1.25.9-rancher1-1" - there are likely multiple ways to implement it with one being switching to matching **ranges** like `>=1.24.13-rancher1-1 <1.25.0-rancher1-1` as suggested in the original comment.

kinarashah:
@snasovich @doflamingo721 Because the new kubernetes versions are being delivered out of band, but RKE will be delivered only with new version of Rancher, I suggest holding off on this fix till we release Rancher. The reasoning is because right now the fix will work for RKE standalone to prevent auto-upgrades, and also out of band Kubernetes versions in Rancher. But in future when Rancher gets released with this fixed RKE code, the out of band Kubernetes versions would auto upgrade on Rancher upgrade because of condition checks for >=, which is expected but it doesn't work in this use case.

cluster/plan.go
doflamingo721: Okay. I understood it now. Will update accordingly. Thanks for the explanation.

skaven81:
Any update on which Rancher release this will make it into?  I presume it passed the deadline for 2.7.5.  But is there a chance it'll be in 2.7.6?

doflamingo721:
Closing this PR in favour of https://github.com/rancher/rke/pull/3303

PR #3230, opened 2023-04-25
[1.3] Update KDM for `1.3.20-rc4`

HarrisonWAffel:
Brings in changes from dev-v2.6 so they can be tested in rancher RC 2.6.12-rc4


rancherbot (approved):

kinarashah (dismissed):

PR #3231, opened 2023-04-26
[v1.3] update kdm branch for v1.3.20

kinarashah:


HarrisonWAffel (approved):
lgtm, I'm assuming that there are no changes to data.json as the last RKE RC was cut yesterday which updated datajson and there have been no KDM updates since then to include in data.json

PR #3241, opened 2023-05-04
[v1.4] update kdm for default v1.26

kinarashah:
https://github.com/rancher/rancher/issues/41113


rancherbot (approved):

PR #3254, opened 2023-06-06
[v2.6] Added ACI 5.2.7.1 variables

LavanyaVaddavalli:
https://github.com/rancher/rancher/issues/41761

### DIFF
--- a/cluster/defaults.go
+++ b/cluster/defaults.go
@@ -119,6 +119,11 @@ const (
 	DefaultAciSleepTimeSnatGlobalInfoSync       = "0"
 	DefaultAciOpflexAgentOpflexAsyncjsonEnabled = "false"
 	DefaultAciOpflexAgentOvsAsyncjsonEnabled    = "false"
+	DefaultAciOpflexAgentPolicyRetryDelayTimer  = "10"
+	DefaultAciAciMultipod                       = "false"
+	DefaultAciAciMultipodUbuntu                 = "false"
+	DefaultAciDhcpRenewMaxRetryCount            = "0"
+	DefaultAciDhcpDelay                         = "0"
 
 	KubeAPIArgAdmissionControlConfigFile             = "admission-control-config-file"
 	DefaultKubeAPIArgAdmissionControlConfigFileValue = "/etc/kubernetes/admission.yaml"
@@ -687,6 +692,11 @@ func (c *Cluster) setClusterNetworkDefaults() {
 			AciSleepTimeSnatGlobalInfoSync:       DefaultAciSleepTimeSnatGlobalInfoSync,
 			AciOpflexAgentOpflexAsyncjsonEnabled: DefaultAciOpflexAgentOpflexAsyncjsonEnabled,
 			AciOpflexAgentOvsAsyncjsonEnabled:    DefaultAciOpflexAgentOvsAsyncjsonEnabled,
+			AciOpflexAgentPolicyRetryDelayTimer:  DefaultAciOpflexAgentPolicyRetryDelayTimer,
+			AciAciMultipod:                       DefaultAciAciMultipod,
+			AciAciMultipodUbuntu:                 DefaultAciAciMultipodUbuntu,
+			AciDhcpRenewMaxRetryCount:            DefaultAciDhcpRenewMaxRetryCount,
+			AciDhcpDelay:                         DefaultAciDhcpDelay,
 		}
 	}
 	if c.Network.CalicoNetworkProvider != nil {
@@ -750,6 +760,11 @@ func (c *Cluster) setClusterNetworkDefaults() {
 		setDefaultIfEmpty(&c.Network.AciNetworkProvider.SleepTimeSnatGlobalInfoSync, DefaultAciSleepTimeSnatGlobalInfoSync)
 		setDefaultIfEmpty(&c.Network.AciNetworkProvider.OpflexAgentOpflexAsyncjsonEnabled, DefaultAciOpflexAgentOpflexAsyncjsonEnabled)
 		setDefaultIfEmpty(&c.Network.AciNetworkProvider.OpflexAgentOvsAsyncjsonEnabled, DefaultAciOpflexAgentOvsAsyncjsonEnabled)
+		setDefaultIfEmpty(&c.Network.AciNetworkProvider.OpflexAgentPolicyRetryDelayTimer, DefaultAciOpflexAgentPolicyRetryDelayTimer)
+		setDefaultIfEmpty(&c.Network.AciNetworkProvider.AciMultipod, DefaultAciAciMultipod)
+		setDefaultIfEmpty(&c.Network.AciNetworkProvider.AciMultipodUbuntu, DefaultAciAciMultipodUbuntu)
+		setDefaultIfEmpty(&c.Network.AciNetworkProvider.DhcpRenewMaxRetryCount, DefaultAciDhcpRenewMaxRetryCount)
+		setDefaultIfEmpty(&c.Network.AciNetworkProvider.DhcpDelay, DefaultAciDhcpDelay)
 		networkPluginConfigDefaultsMap[AciOVSMemoryLimit] = c.Network.AciNetworkProvider.OVSMemoryLimit
 		networkPluginConfigDefaultsMap[AciImagePullPolicy] = c.Network.AciNetworkProvider.ImagePullPolicy
 		networkPluginConfigDefaultsMap[AciPBRTrackingNonSnat] = c.Network.AciNetworkProvider.PBRTrackingNonSnat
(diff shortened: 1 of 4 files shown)


cluster/defaults.go
```diff
@@ -807,13 +823,10 @@ func (c *Cluster) setClusterNetworkDefaults() {
 		networkPluginConfigDefaultsMap[AciEncapType] = c.Network.AciNetworkProvider.EncapType
 		networkPluginConfigDefaultsMap[AciMcastRangeStart] = c.Network.AciNetworkProvider.McastRangeStart
 		networkPluginConfigDefaultsMap[AciMcastRangeEnd] = c.Network.AciNetworkProvider.McastRangeEnd
-		networkPluginConfigDefaultsMap[AciNodeSubnet] = c.Network.AciNetworkProvider.NodeSubnet
```
snasovich: Don't recall us removing variables before. Is it safe to do so? @kinarashah - FYI

cluster/network.go
```diff
@@ -485,26 +498,50 @@ func (c *Cluster) doWeaveDeploy(ctx context.Context, data map[string]interface{}
...
-	if err != nil {
-		return err
+	var podIPPool []IPPool
+	var podNetwork []PodIPNetwork
+	var podSubnet []string
+
+	clusterCIDRs := strings.Split(c.ClusterCIDR, ",")
+	for _, clusterCIDR := range clusterCIDRs {
+		podSubnet = append(podSubnet, fmt.Sprintf("\"%s\"", clusterCIDR))
+		_, ClusterCIDRs, err := net.ParseCIDR(clusterCIDR)
```
snasovich: Suggesting to rename variables to avoid confusion as we already have `clusterCIDRs` and now define `ClusterCIDRs`.

cluster/network.go
```diff
@@ -485,26 +498,50 @@ func (c *Cluster) doWeaveDeploy(ctx context.Context, data map[string]interface{}
...
+		if err != nil {
+			return err
+		}
+		podIPStart, podIPEnd := cidr.AddressRange(ClusterCIDRs)
+		podIPPool = append(podIPPool, IPPool{Start: cidr.Inc(cidr.Inc(podIPStart)), End: cidr.Dec(podIPEnd)})
```
snasovich, line 518: Not sure why we need to  increment `podIPStart` _twice_ and decrement `podIPEnd` (I mean I see a reason to increment/decrement once to exclude bottom/top of the range - but _twice_ increment is non-trivial so I believe a comment explaining the reason should be added).
(Reviewing the code further I see that's how it worked before - but still think it's a good idea to add a "why?" comment)

types/rke_types.go
```diff
@@ -599,16 +599,16 @@ type AciNetworkProvider struct {
 	VmmDomain                         string              `yaml:"vmm_domain,omitempty" json:"vmmDomain,omitempty"`
 	VmmController                     string              `yaml:"vmm_controller,omitempty" json:"vmmController,omitempty"`
 	EncapType                         string              `yaml:"encap_type,omitempty" json:"encapType,omitempty"`
-	NodeSubnet                        string              `yaml:"node_subnet,omitempty" json:"nodeSubnet,omitempty"`
+	NodeSubnet                        []string            `yaml:"node_subnet" json:"nodeSubnet,omitempty"`
```
snasovich: We're changing types on the existing type - I'm not sure how safe it is and how upgrade scenarios will be handled in such cases. @kinarashah , need your input.

cluster/network.go
```diff
@@ -485,26 +498,50 @@ func (c *Cluster) doWeaveDeploy(ctx context.Context, data map[string]interface{}
...
+		podNetwork = append(podNetwork, PodIPNetwork{Subnet: *ClusterCIDRs, Gateway: cidr.Inc(podIPStart)})
 	}
-	podIPStart, podIPEnd := cidr.AddressRange(clusterCIDR)
-	_, staticExternalSubnet, err := net.ParseCIDR(c.Network.Options[AciStaticExternalSubnet])
-	if err != nil {
-		return err
+
+	var staticServiceIPPool []IPPool
+	staticExternalSubnets := strings.Split(c.Network.AciNetworkProvider.StaticExternalSubnet[0], ",")
```
snasovich: Couple of questions/concerns.

1. Can the `c.Network.AciNetworkProvider.StaticExternalSubnet` be guaranteed non-empty? I see a check for `nil` but it would allow an empty slice.
2. It doesn't look like we access any elements past zero-index one. What's the point of it being a slice then? (And we split the only element on commas to get an array to boot)

cluster/network.go
```diff
@@ -485,26 +498,50 @@ func (c *Cluster) doWeaveDeploy(ctx context.Context, data map[string]interface{}
...
+	for _, staticExternalSubnet := range staticExternalSubnets {
+		_, externStatic, err := net.ParseCIDR(strings.Trim(staticExternalSubnet, "\\\""))
```
snasovich: Unclear why we need Trim here.

cluster/network.go
```diff
@@ -485,26 +498,50 @@ func (c *Cluster) doWeaveDeploy(ctx context.Context, data map[string]interface{}
...
 	if err != nil {
 		return err
 	}
 	nodeServiceIPStart, nodeServiceIPEnd := cidr.AddressRange(svcGraphSubnet)
-	_, dynamicExternalSubnet, err := net.ParseCIDR(c.Network.Options[AciDynamicExternalSubnet])
-	if err != nil {
-		return err
+
+	var serviceIPPool []IPPool
+	dynamicExternalSubnets := strings.Split(c.Network.AciNetworkProvider.DynamicExternalSubnet[0], ",")
```
snasovich: Same comment as above about using slice. Overall I feel like it's worth adding comments explaining the format expected in these fields.

And I'm now more worried about upgrade scenarios as it looks like expected input fields format is changed.

snasovich (changes requested):

cluster/defaults.go
kinarashah: We don't remove variables this way cuz it causes upgrades on RKE reconcile. But looks like the net result in template might be the same here because we're changing the logic to assign field from struct field instead. @LavanyaVaddavalli what is the reasoning behind moving it from Network.Options to a first class field under struct now? Users would be able to pass it both ways so I'm curious on why we're moving certain fields now.  

Same for DynamicExternalSubnet and StaticExternalSubnet.

types/rke_types.go
kinarashah: Yeah I don't like this change in type, even if we add validation on next reconcile to make sure users update cluster.yml - it'd probably make more sense to add a struct for ipv6 and have corresponding fields there.

LavanyaVaddavalli:
@snasovich Modified the code to not delete/modify existing variables, instead taking the dualstack variables in string datatype which has multiple IP subnets which are comma seperated. This could solve the issues you mentioned with upgrade.

cluster/network.go
LavanyaVaddavalli: when subnet IP is incremented once - it is to use it for gateway (l[ink](https://github.com/rancher/rke/pull/3254/files#diff-d4ecbdb0cf2e7bf588e89882093414c03122e317be2367865ce29796c57cf403R517))
when subnet IP is incremented twice - it is the start of the subnet

LavanyaVaddavalli:
The unit tests failed at a step, with error "2023-06-09T11:20:27Z [FAIL] v1.22.17-rancher1-2 (rke up)"

Since we did not add any ACI changes in v1.22.17-rancher1-2 version, I don't think this failure is because of our changes in this PR

cluster/network.go
```diff
@@ -323,7 +339,7 @@ var EtcdClientPortList = []string{
 }
 
 var CalicoNetworkLabels = []string{CalicoNodeLabel, CalicoControllerLabel}
-var IPv6CompatibleNetworkPlugins = []string{CalicoNetworkPlugin}
+var IPv6CompatibleNetworkPlugins = []string{CalicoNetworkPlugin, AciNetworkPlugin}
```
kinarashah, line 344: @LavanyaVaddavalli I think we need to scope the ipv6 compatibility only to `ACI >= 5.2.7.1` (or kubernetes versions that support this plugin version). If users update CIDR to comma separated values for ipv6 but don't upgrade to the new Kubernetes versions, this will break the ACI deploy because the older template renders it directly (same for all fields below) 
` "node-subnet": "{{.NodeSubnet}}",` 
vs 
`"node-subnet": [{{- range $index, $item := .NodeSubnet }}{{- if $index}},{{end}}{{$item}}{{end}}]`. 

Is this a valid scenario or something that's obvious to users by default? I am okay keeping it as is if you think users would be aware of which Kubernetes versions to run before trying to use ipv6.

kinarashah:
@LavanyaVaddavalli You're right, don't think your changes caused v1.22.x to fail. It was a flaky test, restarting build fixed the issue.

cluster/network.go
LavanyaVaddavalli: @kinarashah We mention to the user which kubernetes versions are supported for a particular aci release. But if the user still missed updating the kubernetes version, then this is a valid scenario. 

Since we do not take the ACI version as input anywhere, we would have to put a validation check based on kubernetes version { >=1.23.16-rancher2-3 <1.24.0-rancher0, >=1.24.13-rancher2-2 <1.25.0-rancher0, >=1.25.9-rancher2-2) The kubernetes version should satisfy any one of the conditions to be eligible. Is there any pointers on how to add this check for kubernetes versions?

kinarashah: @LavanyaVaddavalli Yeah scoping it with Kubernetes version checks will work. Here's an example on how to pass combination of ranges https://github.com/rancher/rke/blob/e0b3970591569b19b5be13be5a8febaa4ac9a163/cluster/defaults.go#L1154

LavanyaVaddavalli: Addressed in 4471ba579916db076fbace46d2e89a3d4277e7fa

snasovich (approved):

jiaqiluo (approved):

cluster/network.go
kinarashah: Thanks for updating, confirmed that the range restrictions for ipv6 look right to not break templates now 
```
">=1.23.16-rancher2-3 <1.24.0-rancher0":    aciv5271
">=1.24.13-rancher2-2":                     aciv5271
```

kinarashah (approved):

Issue #3247, opened 2023-05-23
 FATA[0000] Failed to validate cluster: cluster version must be at least v1.23 to use PodSecurity in RKE


GlitchTQ:
rke version v1.4.5 from git hub releases

Docker version 20.10.9

opensuse:leap:15.2

VmWare

**cluster.yaml file:**
``` yaml
nodes:
  - address: 10.10.10.1
    internal_address: 10.10.10.1
    user: root
    ssh_key_path: ~/.ssh/id_rsa
    role: [controlplane, worker, etcd]
  - address: 10.10.10.2
    internal_address: 10.10.10.2
    user: root
    ssh_key_path: ~/.ssh/id_rsa
    role: [controlplane, worker, etcd]
  - address: 10.10.10.2
    internal_address: 10.10.10.2
    user: root
    ssh_key_path: ~/.ssh/id_rsa
    role: [controlplane, worker, etcd]

services:
  etcd:
    snapshot: true
    creation: 6h
    retention: 24h

# network:
#   plugin: calico

# Required for external TLS termination with
# ingress-nginx v0.22+
ingress:
  provider: nginx
  options:
    use-forwarded-headers: "true"

# enable_cri_dockerd: true
cluster_name: local
# kubernetes_version: v1.21.14-rancher1-1
kubernetes_version: v1.23.16-rancher2-2
```

**Steps to Reproduce:**
rke -d up --config rancher-cluster.yml

**Results:**
`FATA[0000] Failed to validate cluster: cluster version must be at least v1.23 to use PodSecurity in RKE`

I want to update rke k8s cluster to v1.23 version, how can i avoid this error?

alanbchristie:
I had a similar issue and wonder whether your issue is caused by Pod Security Admissions (PSA) support that may have been added in RKE v1.4.4? Have a look at the **Workaround** suggestion in https://github.com/rancher/rancher/issues/41292#issuecomment-1528181762.

I suspect that starting with a pre-1.4.4 RKE that supports your existing kubernetes version will help.

_I'm not a contributor/developer of RKE, just a user who encountered a similar issue last week_

jiaqiluo:
waiting for the next RC 

---

### Root cause
  
[...]

PR #3261, opened 2023-06-14
set the default PodSecurityConfiguration value only if the cluster's k8s version is at least 1.23

jiaqiluo:
# Issue: 

https://github.com/rancher/rke/issues/3247
https://github.com/rancher/rancher/issues/41292

# Problem
 

 
The issue is reported in Rancher at first, but the root cause is in RKE. 

When RKE initializes the cluster, it sets the default value to the field `spec.Services.KubeAPI.PodSecurityConfiguration` when no value is provided in the cluster.yml file (see [code](https://github.com/rancher/rke/blob/5d360491203a3c1511aa6d12c496679dea05ba37/cluster/defaults.go#L398-L400)) **regardless of the cluster's Kubernetes version**.

But later on, when RKE validates the cluster's configuration before starting the upgrading, it expects to see the abovementioned field to be set only for clusters whose k8s version is **at least 1.23** (see [code](https://github.com/rancher/rke/blob/5d360491203a3c1511aa6d12c496679dea05ba37/cluster/validation.go#L691-L703)). Therefore the following error is returned: 
```
FATA[0000] Failed to validate cluster: cluster version must be at least v1.23 to use PodSecurity in RKE
```

To reproduce the bug, we can use RKE CLI v1.4.5 to provision a cluster with k8s version 1.22.17-rancher1-2, although the version is not officially supported by RKE CLI v1.4.5. 

```
// cluster.yaml
nodes:
  - address: xxx
    internal_address: xxx
    user: ubuntu
    role: [controlplane,worker,etcd]
kubernetes_version: v1.22.17-rancher1-2
```

# Solution
 

Check the cluster k8s version and set the default PodSecurityConfiguration value only if the cluster's k8s version is at least 1.23

After this PR is merged, a following PR need to be raised in Rancher to bump the RKE module. 

# Testing
 

Follow the same steps to provision a cluster with k8s version `1.22.17-rancher1-2` using the RKE binary compiled from this branch. The cluster is provisioned successfully. Also, if we ssh into the node to check the file `/etc/kubernetes/admission.yaml`, we can see that the default PodSecurity Configuration is not in the file, which is expected.

### DIFF
--- a/cluster/defaults.go
+++ b/cluster/defaults.go
@@ -399,8 +399,15 @@ func (c *Cluster) setClusterServicesDefaults() {
 			c.Services.KubeAPI.EventRateLimit.Configuration == nil {
 			c.Services.KubeAPI.EventRateLimit.Configuration = newDefaultEventRateLimitConfig()
 		}
-		if len(c.Services.KubeAPI.PodSecurityConfiguration) == 0 {
-			c.Services.KubeAPI.PodSecurityConfiguration = PodSecurityPrivileged
+		parsedVersion, err := getClusterVersion(c.Version)
+		if err != nil {
+			logrus.Warnf("Can not parse the cluster version [%s] to determine wether to set the default PodSecurityConfiguration: %v", c.Version, err)
+		} else {
+			if parsedRangeAtLeast123(parsedVersion) {
+				if len(c.Services.KubeAPI.PodSecurityConfiguration) == 0 {
+					c.Services.KubeAPI.PodSecurityConfiguration = PodSecurityPrivileged
+				}
+			}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


a-blender (approved):
@jiaqiluo Does this need to be backported because rke v1.3 versions are included in Rancher clusters that support kubernetes 1.23?

jiaqiluo:
hi @a-blender, nope, we do not need to backport it to the v1.3 line because we did not support Pod Security, ie. add our logic, in that release line.

a-blender:
@jiaqiluo Alright! LGTM

jakefhyde (approved):

atsai1220:
How come this was reverted before the 1.4.6 release?

Issue #3271, opened 2023-06-23
cri-dockerd needs IPv6 dual stack flag enabled when more than 1 CIDR is configured


superseb:
**RKE version:**
v1.4.6

**Docker version: (`docker version`,`docker info` preferred)**
Tested with 23.0.6

**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)**
Ubuntu 22.04, `5.19.0-1025-aws`

**Type/provider of hosts: (VirtualBox/Bare-metal/AWS/GCE/DO)**
AWS EC2

**cluster.yml file:**
```
services:
  kube-api:
    service_cluster_ip_range: 10.43.0.0/16,fd98::/108
  kube-controller:
    service_cluster_ip_range: 10.43.0.0/16,fd98::/108
    cluster_cidr: 10.42.0.0/16,fd01::/64

network:
  plugin: calico
```

**Steps to Reproduce:**
Run `rke up`, deploy pod(s) and check `.status.podIPs`

**Results:**
It will only show IPv4 address, not the IPv6 address.


After looking at the `cri-dockerd` flags, I found this flag which needs to be enabled to make it work. After running with a custom `rke-tools` with this flag added, the IPv6 connectivity worked.

```
      --ipv6-dual-stack                         Enable IPv6 dual stack support                                                                                                                                                                                                                                                
```

terraboops:
@superseb - I'd like to take this on, I've been investigating this issue for our RKE clusters and I think I can have a fix for it in the next few days.

superseb:
@terramauthe thanks for the help! I opened 2 PRs for the issue as linked above, if you are able to test it if it resolves the issue for your cluster(s), that would be great. If not or if you come up with a better solution, let me know.

terraboops:
> @terramauthe thanks for the help! I opened 2 PRs for the issue as linked above, if you are able to test it if it resolves the issue for your cluster(s), that would be great. If not or if you come up with a better solution, let me know.

Great! Thanks! I should be able to take a look tomorrow at testing and PR feedback.

superseb:
You should be able to test it using RKE CLI built from the PR, you need to build the `rke-tools` image and make it available (publicly or locally) and then override the `rke-tools` image in the configuration (probably only `kubernetes_services_sidecar` but automation changes them all when I test):

```
system_images:                   
  alpine: your_name/rke-tools:ipv6
  nginx_proxy: your_name/rke-tools:ipv6
[...]

PR #3272, opened 2023-06-27
[release/v1.4] Add env var to enable dual-stack in cri-dockerd

superseb:
https://github.com/rancher/rke/issues/3271

The flag to enable cri-dockerd to work properly with dual-stack was missing, this PR adds another environment variable to the container to instruct the entrypoint in rke-tools (https://github.com/rancher/rke-tools/pull/163) to configure the needed flag.

As there is no good way to detect multiple CIDR/dual-stack inside the container itself, we need to instruct the container via an environment variable after checking the cluster configuration (`cluster-cidr` flag containing more than one CIDR)

I wanted to add IPv6 tests to `scripts/integration` but because the DinD setup, this needs a lot more work to enable IPv6 and make the test pass.

### DIFF
--- a/cluster/plan.go
+++ b/cluster/plan.go
@@ -22,6 +22,7 @@ import (
 	v3 "github.com/rancher/rke/types"
 	"github.com/rancher/rke/util"
 	"github.com/sirupsen/logrus"
+	netutils "k8s.io/utils/net"
 	"sigs.k8s.io/yaml"
 )
 
@@ -62,6 +63,7 @@ const (
 	EncryptionProviderConfigArgument = "encryption-provider-config"
 
 	KubeletCRIDockerdNameEnv = "RKE_KUBELET_CRIDOCKERD"
+	KubeletDualStackNameEnv  = "RKE_KUBELET_CRIDOCKERD_DUALSTACK"
 )
 
 var (
@@ -602,6 +604,11 @@ func (c *Cluster) BuildKubeletProcess(host *hosts.Host, serviceOptions v3.Kubern
 		Env = append(Env,
 			// Enable running cri-dockerd
 			fmt.Sprintf("%s=%s", KubeletCRIDockerdNameEnv, "true"))
+		if c.multipleCIDRsConfigured() {
+			Env = append(Env,
+				// Enable cri-dockerd flag for dual-stack
+				fmt.Sprintf("%s=%s", KubeletDualStackNameEnv, "true"))
+		}
 	}
 
 	if len(c.CloudProvider.Name) > 0 {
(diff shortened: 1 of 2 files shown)


a-blender (dismissed):
Thanks @superseb, how was this tested? Do we just need to make sure the env var is set when multiple CIDRs are configured or is there more functionality that needs to be verified?

cluster/plan.go
```diff
@@ -602,6 +603,11 @@ func (c *Cluster) BuildKubeletProcess(host *hosts.Host, serviceOptions v3.Kubern
 		Env = append(Env,
 			// Enable running cri-dockerd
 			fmt.Sprintf("%s=%s", KubeletCRIDockerdNameEnv, "true"))
+		if c.multipleCIDRsConfigured() {
```
manuelbuil, line 607: What about using the library https://pkg.go.dev/k8s.io/utils/net with functions such as IsDualStackCIDRStrings. That will verify that the CIDRs include IPv4 and IPv6

superseb: Nice, going to check this.

cluster/plan.go
```diff
@@ -1286,3 +1293,26 @@ func (c *Cluster) IsCRIDockerdEnabled() bool {
...
+	if c == nil {
+		logrus.Debug("multipleCIDRsConfigured: Returning false, cluster object is nil")
+		return false
+	}
+	if c.Services.KubeController.ClusterCIDR == "" {
+		logrus.Debug("multipleCIDRsConfigured: Returning false, ClusterCIDR is empty")
+		return false
+	}
+	CIDRsSplitted := strings.Split(c.Services.KubeController.ClusterCIDR, ",")
+	multipleCIDRs, err := netutils.IsDualStackCIDRStrings(CIDRsSplitted)
```
manuelbuil, line 1307: nit: in rke2/k3s, we would combine these two into one. I see benefits in both options so, I don't really mind

cluster/plan.go
```diff
@@ -1286,3 +1293,26 @@ func (c *Cluster) IsCRIDockerdEnabled() bool {
 	}
 	return false
 }
+
+func (c *Cluster) multipleCIDRsConfigured() bool {
+	if c == nil {
+		logrus.Debug("multipleCIDRsConfigured: Returning false, cluster object is nil")
+		return false
+	}
```
manuelbuil, line 1301: Hold on, if `c == nil` we would never reach this code, right? This check is probably not needed

manuelbuil (changes requested):

cluster/plan.go
superseb: It is aligned with the other function that checks for c being `nil` (https://github.com/rancher/rke/blob/release/v1.4/cluster/plan.go#L1280), that's why I added it.

manuelbuil: Same problem there. It is impossible to reach that code because the struct does not exist, so we could never reach its methods

manuelbuil: I have just run a quick test and there might be a way to reach the method of a struct even if the struct is nil. Let me quickly check this tomorrow, I might be wrong after all

jakefhyde: @manuelbuil I can confirm that calling a method on a nil struct is fine, it will only panic if you try to access it in some way, like line 1302. Source: https://stackoverflow.com/questions/42238624/calling-a-method-on-a-nil-struct-pointer-doesnt-panic-why-not

manuelbuil: thanks @jakefhyde. I had never seen that check in go code, that's why it confused me

manuelbuil (approved):

probably-autopilot:
Hey. Will this get merged?
We are running Rancher 2.7.10 and we are updating an older cluster running Kubernetes version 1.23.16 and we are stopped from updating to v1.24.17 due to IPv6 breaking. 
I'm looking into manually moving to a newer rke-tools image so that we can get these upgrades completed, but if this were to get merged and a newer rke-tools image get an official release, that would be preferable.

PR #3312, opened 2023-07-31
Add monthly issue metrics

superseb:
Start using the new [issue-metrics](https://github.com/github/issue-metrics) Action to get insights into issue handling, we probably want to tweak the config along the way.

See https://github.blog/2023-07-19-metrics-for-issues-pull-requests-and-discussions/ for intro blog

### DIFF
--- a/.github/workflows/issue-metrics.yml
+++ b/.github/workflows/issue-metrics.yml
@@ -1,0 +1,41 @@
+name: Monthly issue metrics
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '3 2 1 * *'
+
+permissions:
+  issues: write
+  pull-requests: read
+
+jobs:
+  build:
+    name: issue metrics
+    runs-on: ubuntu-latest
+    
+    steps:
+    - name: Get dates for last month
+      shell: bash
+      run: |
+        # Calculate the first day of the previous month
+        first_day=$(date -d "last month" +%Y-%m-01)
+
+        # Calculate the last day of the previous month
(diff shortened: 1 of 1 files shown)


.github/workflows/issue-metrics.yml
```diff
@@ -0,0 +1,41 @@
...
+
+        # Calculate the last day of the previous month
+        last_day=$(date -d "$first_day +1 month -1 day" +%Y-%m-%d)
+        
+        #Set an environment variable with the date range
+        echo "$first_day..$last_day"
+        echo "last_month=$first_day..$last_day" >> "$GITHUB_ENV"
+
+    - name: Run issue-metrics tool
+      uses: github/issue-metrics@cf3dda21393dcaf2bbb50f9bac0505f897339122 # v2.3.0
```
manuelbuil, line 31: ugh, this seems ugly. github/issue-metrics@v2.3.0 does not work? Note there is already a `2.4.0`

superseb: GitHub Actions tags are not immutable, meaning that if someone takes over the repository and overwrites the tag, you are running code into your repository that you don't want. Digest pinning (which is done here), means you get the exact code you want.

This also prepares the repository for enabling Renovate, which can update the workflows digests based on tags (which is the format used as shown here). See https://github.com/rancher/renovate-config/commit/264123d1496cac20bca5852a42e031b3acd731d4 as an example

superseb: @manuelbuil please review

thomasferrandiz (approved):

PR #3337, opened 2023-08-18
Add warning log about weave deprecation

manuelbuil:
This PR verifies if:
* k8s version is greater than 1.27
* weave was selected as the CNI plugin

In that case, it prints a warning in the logs: "Weave CNI plugin is deprecated and will be removed in v1.30 kubernetes release"


cluster/validation.go
```diff
@@ -731,3 +737,19 @@ func getClusterVersion(version string) (semver.Version, error) {
 	}
 	return parsedVersion, nil
 }
+
+// warnWeaveDeprecation prints a deprecation warning if version is higher than 1.27 and weave was selected
+func warnWeaveDeprecation(k8sVersion, cniPlugin string) error {
+	version, err := util.StrToSemVer(k8sVersion)
+	if err != nil {
+		return fmt.Errorf("Error in the version: %v", version)
```
snasovich: Should probably reference `k8sVersion` (string to parse) instead of returned semver as I doubt it will return anything usable in case of error. Also would like to include/wrap the actual error returned.

```suggestion
		return fmt.Errorf("Error parsing K8s version '%s': %w", k8sVersion, err)
```

cluster/validation.go
```diff
@@ -731,3 +737,19 @@ func getClusterVersion(version string) (semver.Version, error) {
...
+	}
+	version127, err := util.StrToSemVer("v1.27.0")
+	if err != nil {
+		return fmt.Errorf("Error in the version: %v", version127)
```
snasovich: Same as the above comment on wrapping this error.

snasovich (commented):
Couple comments to address.
Note: Responding as "Comment" instead of "Request changes" not to block merging if I'm slow to circle back on the review after these comments are addressed.

cluster/validation.go
```diff
@@ -217,6 +217,12 @@ func validateNetworkOptions(c *Cluster) error {
 	if c.Network.Plugin == FlannelNetworkPlugin && c.Network.MTU != 0 {
 		return fmt.Errorf("Network plugin [%s] does not support configuring MTU", FlannelNetworkPlugin)
 	}
+
+	err := warnWeaveDeprecation(c.Version, c.Network.Plugin)
```
kinarashah: check for `cniPlugin == WeaveNetworkPlugin` before calling so only version checks belong to the weave func itself

manuelbuil: I was checking this inside the function but I can do it outside too

cluster/validation.go
manuelbuil: good point

cluster/validation.go
```diff
@@ -217,6 +217,14 @@ func validateNetworkOptions(c *Cluster) error {
 	if c.Network.Plugin == FlannelNetworkPlugin && c.Network.MTU != 0 {
 		return fmt.Errorf("Network plugin [%s] does not support configuring MTU", FlannelNetworkPlugin)
 	}
+
+	if c.Network.Plugin == WeaveNetworkPlugin {
+		err := warnWeaveDeprecation(c.Version)
+		if err != nil {
+			return fmt.Errorf("Error while printing Weave deprecation message: %v", err)
+		}
```
snasovich: Nit: Combine the assignment and the if check
```suggestion
		if err := warnWeaveDeprecation(c.Version); err != nil {
			return fmt.Errorf("Error while printing Weave deprecation message: %w", err)
		}
```

cluster/validation.go
```diff
@@ -731,3 +739,19 @@ func getClusterVersion(version string) (semver.Version, error) {
 	}
 	return parsedVersion, nil
 }
+
+// warnWeaveDeprecation prints a deprecation warning if version is higher than 1.27
+func warnWeaveDeprecation(k8sVersion string) error {
+	version, err := util.StrToSemVer(k8sVersion)
+	if err != nil {
+		return fmt.Errorf("failed to translate: %s to semver notation: %v", k8sVersion, err)
```
snasovich: Nit: Any reason we can't use error wrapping? https://stackoverflow.com/a/61287626
```suggestion
		return fmt.Errorf("failed to translate: %s to semver notation: %w", k8sVersion, err)
```

cluster/validation.go
```diff
@@ -731,3 +739,19 @@ func getClusterVersion(version string) (semver.Version, error) {
...
+	}
+	version127, err := util.StrToSemVer("v1.27.0")
+	if err != nil {
+		return fmt.Errorf("fialed to tranlate v1.27.0 to semver notation: %v", err)
```
snasovich: Minor typo and error wrapping (if we can do it):
```suggestion
		return fmt.Errorf("failed to tranlate v1.27.0 to semver notation: %w", err)
```

jiaqiluo (approved):
LGTM

cluster/validation.go
```diff
@@ -731,3 +738,19 @@ func getClusterVersion(version string) (semver.Version, error) {
...
+
+// warnWeaveDeprecation prints a deprecation warning if version is higher than 1.27
+func warnWeaveDeprecation(k8sVersion string) error {
+	version, err := util.StrToSemVer(k8sVersion)
+	if err != nil {
+		return fmt.Errorf("failed to translate: %s to semver notation: %w", k8sVersion, err)
+	}
+	version127, err := util.StrToSemVer("v1.27.0")
+	if err != nil {
+		return fmt.Errorf("failed to tranlate v1.27.0 to semver notation: %w", err)
```
superseb: typo `tranlate`

cluster/validation.go
```diff
@@ -731,3 +738,19 @@ func getClusterVersion(version string) (semver.Version, error) {
 	}
 	return parsedVersion, nil
 }
+
+// warnWeaveDeprecation prints a deprecation warning if version is higher than 1.27
+func warnWeaveDeprecation(k8sVersion string) error {
+	version, err := util.StrToSemVer(k8sVersion)
+	if err != nil {
+		return fmt.Errorf("failed to translate: %s to semver notation: %w", k8sVersion, err)
```
superseb: Probably want to make this more uniform, see https://github.com/rancher/rke/blob/c2395d3451087f864ec46af8226604ceb7caaf07/hosts/tunnel.go#L70

cluster/validation.go
```diff
@@ -731,3 +738,19 @@ func getClusterVersion(version string) (semver.Version, error) {
...
+	}
+	version127, err := util.StrToSemVer("v1.27.0")
+	if err != nil {
+		return fmt.Errorf("failed to tranlate v1.27.0 to semver notation: %w", err)
+	}
+	if !version.LessThan(*version127) {
```
superseb, line 752: We scope this to 1.27 and up because we only support and expect users to first upgrade to new minor (so 1.25->1.26->1.27) to see this message? Is there a problem to warn every user of Weave that it will be removed in v1.30.x?

cluster/validation.go
```diff
@@ -731,3 +738,19 @@ func getClusterVersion(version string) (semver.Version, error) {
...
+		logrus.Warn("Weave CNI plugin is deprecated and will be removed in v1.30 kubernetes release")
```
superseb: I would make this as extended as possible, something like:

```
Weave CNI plugin is deprecated starting with RKE v1.5.0 and Kubernetes v1.27.x and will be removed in Kubernetes v1.30.x
```

Or something along those lines.

cluster/validation.go
manuelbuil: No problem but this is what was agreed with the PM after a few conversations

cluster/validation.go
manuelbuil: ok

cluster/validation.go
superseb: Can we include those agreements in the issue/PR? We can't review things if the agreements or decisions are not recorded which can be matched with the implementation. I am not here to override PM decisions so we can leave it for now.

manuelbuil: I think the issue already describes it but I can add even more details about it. I don't think PM will be against adding the message to all the different branches, so if you think that adds a value, we can do it

superseb (approved):

PR #3344, opened 2023-08-24
[release/v1.5] Add K8s v1.27.5

rayandas:
Related: https://github.com/rancher/rancher/issues/41840


cluster/plan.go
```diff
@@ -505,7 +505,7 @@ func (c *Cluster) BuildKubeletProcess(host *hosts.Host, serviceOptions v3.Kubern
 	var Binds []string
 
 	if c.IsCRIDockerdEnabled() {
-		CommandArgs["container-runtime"] = "remote"
+		//CommandArgs["container-runtime"] = "remote" https://v1-26.docs.kubernetes.io/docs/reference/command-line-tools-reference/kubelet/
```
kinarashah: Does `container-runtime` default to `remote` for <v1.27 versions? If not, we should add a separate check for >=v1.27. Need to make sure we don't break older versions by removing this field.

docker/docker.go
```diff
@@ -400,11 +400,11 @@ func StopContainer(ctx context.Context, dClient *client.Client, hostname string,
 	}
 	var err error
 	// define the stop timeout
-	stopTimeoutDuration := StopTimeout * time.Second
+	stopTimeoutDuration := StopTimeout
```
kinarashah: don't need this anymore as `StopTimeout` can be used directly, same for `RestartTimeout` above.

cluster/plan.go
rayandas: No I don't think it defaults to remote for <v1.27 versions. Will add a separate check here.

docker/docker.go
rayandas: Yeah right. Updating it.

kinarashah (approved):

Issue #3274, opened 2023-06-28
How to check applications health state before rke upgrade


nbisson:
**RKE version:** 1.4.1

**Docker version:**  20.10.23

**Operating system and kernel: Rocky 8

**Type/provider of hosts:** VM/Bare-metal

Hi, I would like to know if it is possible to exclude certain nodes to the upgrade process.

Some applications (ex: Ceph, Kafka, etc) have their own logic et need to be in a healthy state to be upgraded. This state is not visible by the kubernetes node status but by intern commands (ex: `ceph status`).

To upgrade without take the risk to lost data, it would be nice if we can choose which node to upgrade or if we can give this upgrade logic to rke (ex : rke check if cluster is healthy before upgrade each node).

I have found a ticket #1906 that add a labels to ignore upgrade node, but theses labels seems to not work in the latest versions.

The only method that i have found today is to edit ssh key for certain nodes which allows to not upgrade theses nodes.

Thanks.

nbisson:
Still relevant

nbisson:
Still relevant

nbisson:
Still relevant

PR #3348, opened 2023-08-25
Add a label which allows to avoid the update

nbisson:
Hi,

In the continuation of ticket #3274, I added a filter which makes it possible not to update all the nodes at once.

I added it by using the #1906 informations. 

I don't understand the reasons but the filters were added by pull request #1912 and removed by pull request #1952.

In order for a node to ignore an update, the label must be added to it: `user.cattle.io/upgrade-policy=prevent`


nbisson:
@superseb can you please look at this pull request ? Thanks

superseb:
@nbisson I added it to be triaged for next release

PR #3351, opened 2023-08-30
Add tigera operator to Calico CNI installation

thomasferrandiz:
This PR adds support for the Tigera Operator for the Calico CNI as described in: https://github.com/rancher/rancher/issues/42731


go.mod
```diff
@@ -11,7 +11,7 @@ replace (
 require (
 	github.com/Masterminds/sprig/v3 v3.2.2
 	github.com/apparentlymart/go-cidr v1.0.1
-	github.com/aws/aws-sdk-go v1.38.65
+	github.com/aws/aws-sdk-go v1.44.106
```
manuelbuil, line 14: did you update this? I don't think it has anything to do with tigera operator or?

cluster/tigera-operator.go
```diff
@@ -0,0 +1,49 @@
...
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/kubectl/pkg/scheme"
+)
+
+func tigeraClient(config *rest.Config) (*rest.RESTClient, error) {
+	crdConfig := *config
+	crdConfig.ContentConfig.GroupVersion = &schema.GroupVersion{Group: operator.GroupVersion.Group, Version: operator.GroupVersion.Version}
+	crdConfig.APIPath = "/apis"
+	crdConfig.NegotiatedSerializer = serializer.NewCodecFactory(scheme.Scheme)
+	crdConfig.UserAgent = rest.DefaultKubernetesUserAgent()
```
manuelbuil, line 19: I'm surprised so much stuff needed to create a kubectl client :O

cluster/network.go
```diff
@@ -677,7 +695,7 @@ func (c *Cluster) doAciDeploy(ctx context.Context, data map[string]interface{})
 
 func (c *Cluster) getNetworkPluginManifest(pluginConfig, data map[string]interface{}) (string, error) {
 	switch c.Network.Plugin {
-	case CanalNetworkPlugin, FlannelNetworkPlugin, CalicoNetworkPlugin, WeaveNetworkPlugin, AciNetworkPlugin:
+	case CanalNetworkPlugin, CalicoNetworkPlugin, FlannelNetworkPlugin, WeaveNetworkPlugin, AciNetworkPlugin:
```
manuelbuil, line 698: there is no need for this swap, or?

cmd/up.go
```diff
@@ -242,6 +242,7 @@ func ClusterUp(ctx context.Context, dialersOptions hosts.DialersOptions, flags c
 		return APIURL, caCrt, clientCert, clientKey, nil, err
 	}
 
+	data["calico-upgrade-cluster"] = reconcileCluster
```
manuelbuil, line 245: Could you elaborate a bit on what this mean? I saw we are checking it somewhere else

manuelbuil (commented):
Looks good in general but I am missing where we are checking `calicoConfig["CalicoUseOperator"] = true` to decide that we deploy calico-operator instead of calico-manifest

thomasferrandiz:
> Looks good in general but I am missing where we are checking `calicoConfig["CalicoUseOperator"] = true` to decide that we deploy calico-operator instead of calico-manifest

It's used in the Tigera Operator template itself which is in the KDM repo.

go.mod
thomasferrandiz: I didn't update it manually but I will check why `go mod tidy` changed it.

cluster/tigera-operator.go
thomasferrandiz: Me too, usually it's hidden when there is an auto-generated clientset but it's not provided for the operator API, only for projectcalicoV3, which isn't what we need.

cluster/network.go
thomasferrandiz: indeed, I forgot to restore it.

cmd/up.go
thomasferrandiz: `reconcileCluster` is a boolean which is set to true when rke is upgrading an existing cluster, false when creating a new one. I added the value to `data` as it was the simplest way to access it in `deployCalico`.

PR #3384, opened 2023-09-21
[release/v1.5] Proper description of deleting pods

superseb:
Debugging another issue I saw this seems to be outdated.

- `network, ingress, and metrics` has been extended since the function was created
- There is no such thing as restart cluster pods, they are deleted so renamed function
- Add some extra debug logging

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -1104,12 +1104,12 @@ func ConfigureCluster(
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


superseb:
@a-blender Not sure why it would be `RedeployClusterPods`, the function does not redeploy, it deletes pods. The logic that creates new pods is in Kubernetes (controller and kubelet). 

And the second question is answered in the issue: 

```
network, ingress, and metrics has been extended since the function was created
```

PR #3387, opened 2023-09-21
Fix "unknown revision v0.0.0" error caused by the k8s.io/kubernetes module

jiaqiluo:
Replace some k8s modules with specific versions to get rid of the "unknown revision v0.0.0" error caused by the k8s.io/kubernetes module

Upstream Issue: https://github.com/kubernetes/kubernetes/issues/79384


rancherbot (approved):

superseb (approved):

Issue #3382, opened 2023-09-21
etcd snapshot-restore with etcd v3.5.7 fails due to missing `/bin/sh` in etcd container image


superseb:
**RKE version:**
v1.5.x but any k8s version that is using etcd v3.5.7 (currently k8s v1.27.x)

**Steps to Reproduce:**
- run `rke etcd snapshot-save --name yolo`
- run `rke etcd snapshot-restore --name yolo`

**Results:**
```
INFO[0033] Starting container [etcd-restore] on host [54.203.164.32], try #1                                                                                                                                                                                                                                                  
WARN[0034] Can't start Docker container [etcd-restore] on host [54.203.164.32]: Error response from daemon: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "sh": executable file not found in $PATH: unknown
```

Cause is etcd moving to distroless base image https://github.com/etcd-io/etcd/pull/15016

vardhaman22:
since etcd is moved to distroless base image from v3.5.7 so /bin/sh, mv and rm tools will not be present in etcd Image. which is required by the restore command. [ref](https://github.com/rancher/rke/blob/d2b01281f33970e5980d307935474dac6a9833cd/services/etcd.go#L652-L670)
One solution for that is to use a custom image which has **etcdctl** as well as these other required tools.
Since in rke we use rke-tools image to create etcd snapshots, i have changed ([PR](https://github.com/rancher/rke/pull/3390)) restore image to rke-tools as well. 
since rke-tools has etcdctl with the other required deps for that command. 

Tested this solution by building rke binary with these changes on a rke standalone rke cluster v1.27.5. and the snapshot restore works fine with this change.
Step 1: built a custom rke binary.
Step 2: created a rke v1.27.5 standalone cluster.
Step 3: created an nginx deployment.
Step 4: create a snapshot using command:
```
rke etcd snapshot-save --name yolo
```
Step 5: deleted the nginx deployment
Step 6: restored the snapshot using command:
```
rke etcd snapshot-restore --name yolo
```
Result: snapshot was restored successfully and the nginx deployment was there in the cluster after snapshot restore.

jiaqiluo:
can be validated in RKE v1.5.0-rc8 https://github.com/rancher/rke/releases/tag/v1.5.0-rc8

rishabhmsra:
Validated on rancher v2.8-head

Followed below steps on RKE1 v1.25, v1.26 and v1.27 clusters:
[...]

PR #3390, opened 2023-09-25
changed etcd restore image to rke-tools for etcd >=3.5.7

vardhaman22:
issue: https://github.com/rancher/rke/issues/3382

since etcd is moved to distroless base image from v3.5.7  so **/bin/sh, mv and rm**  tools will not be present in etcd Image. which is required by the restore command. [ref](https://github.com/rancher/rke/blob/d2b01281f33970e5980d307935474dac6a9833cd/services/etcd.go#L652-L670) 
One solution for that is to use a custom image which has etcdctl as well as these other required tools.
Since in rke we use rke-tools image to create etcd snapshots, i have changed restore image to rke-tools as well. since rke-tools has etcdctl with the other required deps for that command.

### DIFF
--- a/cluster/etcd.go
+++ b/cluster/etcd.go
@@ -16,6 +16,8 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+const MinEtcdVersionWithDistrolessImage = "v3.5.7"
+
 func (c *Cluster) SnapshotEtcd(ctx context.Context, snapshotName string) error {
 	backupImage := c.getBackupImage()
 	containerTimeout := DefaultEtcdBackupConfigTimeout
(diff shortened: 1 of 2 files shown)


jiaqiluo (approved):
LGTM.

But I have one concern:

There is a possibility that in some future KDM release, we bump the `c.SystemImages.Etcd`  image tag but not the etcd version in the `rke-tools` image, which results in using an old version of the `etcdcli` to manage a newer version of etcd instance. It may cause any compatibility issues and break the etcd backup/restore feature.  

But it looks like not an issue because we have used the `rke-tools` image to create etcd backups, which uses `etcdctl` internally,  in the `RunEtcdSnapshotSave` function, which proves it worked.

jakefhyde (approved):

superseb (changes requested):
Can we scope the change to etcd v3.5.7 and up to limit possible impact? We don't have resources to test all the variations that are hit by this change, but by limiting the image change to etcd v3.5.7 we can ensure that everything before that works the same as before.

Also, by scoping the change, you ensure that the process works the same for etcd <v3.5.7 as you don't really want anything in existing setups.

One other option could be to move only the non-etcd commands to `rke-tools` and still use the etcd image to run `etcdctl` but this will require more code changes.

vardhaman22:
> Can we scope the change to etcd v3.5.7 and up to limit possible impact? We don't have resources to test all the variations that are hit by this change, but by limiting the image change to etcd v3.5.7 we can ensure that everything before that works the same as before.
> 
> Also, by scoping the change, you ensure that the process works the same for etcd <v3.5.7 as you don't really want anything in existing setups.
> 
> One other option could be to move only the non-etcd commands to `rke-tools` and still use the etcd image to run `etcdctl` but this will require more code changes.

updated the PR to scope the changes to etcd v3.5.7 and up.

jiaqiluo (approved):

cluster/etcd.go
```diff
@@ -281,3 +284,35 @@ func (c *Cluster) getBackupImage() string {
...
+func (c *Cluster) getRestoreImage() string {
+
+	// use etcd image for restore in case of custom system image
+	if !strings.Contains(c.SystemImages.Etcd, "rancher/mirrored-coreos-etcd") {
+		return c.SystemImages.Etcd
+	}
+
+	etcdImageTag, err := util.GetImageTagFromImage(c.SystemImages.Etcd)
+	if err != nil {
+		logrus.Errorf("[etcd] error getting restore image %v", err)
```
superseb: Please specify the correct error for each as there is no way to know where it failed exactly (besides the `err` obviously). This can be specified better, i.e. `error getting image tag from restore image` or something, so it can be looked up more easily when encountered.

vardhaman22: updated.

superseb (approved):

Issue #3388, opened 2023-09-21
CoreDNS pods are not rechable for resolution from other node


maslo64:
**RKE version:**
```rke version v1.4.8```
**Docker version: (`docker version`,`docker info` preferred)**
```
#docker version
Client: Docker Engine - Community
 Version:           20.10.24
 API version:       1.41
 Go version:        go1.19.7
 Git commit:        297e128
 Built:             Tue Apr  4 18:20:23 2023
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.24
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.19.7
  Git commit:       5d6db84
  Built:            Tue Apr  4 18:18:29 2023
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.22
  GitCommit:        8165feabfdfe38c65b599c4993d227328c231fca
 runc:
  Version:          1.1.8
  GitCommit:        v1.1.8-0-g82f18fe
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
server.subdom.company.com:(/docker/rke)(root)#docker info
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Docker Buildx (Docker Inc., v0.10.4-docker)
  compose: Docker Compose (Docker Inc., v2.20.2)
  scan: Docker Scan (Docker Inc., v0.23.0)

Server:
 Containers: 43
  Running: 23
  Paused: 0
  Stopped: 20
 Images: 20
 Server Version: 20.10.24
 Storage Driver: overlay2
  Backing Filesystem: xfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 8165feabfdfe38c65b599c4993d227328c231fca
 runc version: v1.1.8-0-g82f18fe
 init version: de40ad0
[...]

PR #3406, opened 2023-10-05
Add support for FelixConfiguration in CalicoNetworkProvider

thomasferrandiz:
This PR implements support for Calico FelixConfiguration in rke `cluster.yml`.

The Felix configuration is parsed and propagated to the calico template so that it can be applied to the cluster by calico.
In particular this allows the user to specify:
```
network:
  plugin: calico
  calicoNetworkProvider:
    felixConfiguration:
      featureDetectOverride: "ChecksumOffloadBroken=true"
```

as a way to fix the linked issue.

Related issue: https://github.com/rancher/rke/issues/3388

PR based on the calico-fix-cni PR because otherwise I could't test.


rbrtbnfgl (approved):

cluster/cluster.go
```diff
@@ -732,6 +743,15 @@ func ParseConfig(clusterFile string) (*v3.RancherKubernetesEngineConfig, error)
 	if err := parseAddonConfig(clusterFile, &rkeConfig); err != nil {
 		return &rkeConfig, fmt.Errorf("error parsing addon config: %v", err)
 	}
+
+	// inject manually parsed FelixConfigurationSpec for calico network plugin
+	if rkeConfig.Network.Plugin == "calico" {
+		if rkeConfig.Network.CalicoNetworkProvider == nil {
+			var calicoNetworkProvider v3.CalicoNetworkProvider
```
manuelbuil, line 750: you are importing the calico library as `calico` not as `v3`

cluster/cluster.go
```diff
@@ -707,6 +708,16 @@ func ParseConfig(clusterFile string) (*v3.RancherKubernetesEngineConfig, error)
...
+
+	// Calico FelixConfigurationSpec is a 3rd party type that doesn't unmarshall properly
+	// so we need to parse it manually and inject the parsed struct
+	// into the final rkeConfig
+	var felixConfig calico.FelixConfigurationSpec
+	clusterFile, err = resolveFelixConfiguration(clusterFile, &felixConfig)
+	if err != nil {
+		return nil, err
+	}
+
```
manuelbuil, line 720: I would prefer this to be hidden in calico.go and only execute it if `if rkeConfig.Network.Plugin == "calico" {`, that way we don't need to import calico in `cluster.go`. This is personal taste :)

cluster/cluster.go
```diff
@@ -732,6 +743,15 @@ func ParseConfig(clusterFile string) (*v3.RancherKubernetesEngineConfig, error)
...
 	}
+
+	// inject manually parsed FelixConfigurationSpec for calico network plugin
+	if rkeConfig.Network.Plugin == "calico" {
+		if rkeConfig.Network.CalicoNetworkProvider == nil {
+			var calicoNetworkProvider v3.CalicoNetworkProvider
+			rkeConfig.Network.CalicoNetworkProvider = &calicoNetworkProvider
+		}
+		rkeConfig.Network.CalicoNetworkProvider.FelixConfiguration = felixConfig
+	}
```
manuelbuil, line 754: Like in my other comment, I would prefer if we have a function in `calico.go` that does all this when `rkeConfig.Network.Plugin == "calico"`, so that we don't have to import calico in `cluster.go`

cluster/plan.go
```diff
@@ -563,6 +568,29 @@ func (c *Cluster) BuildKubeletProcess(host *hosts.Host, serviceOptions v3.Kubern
...
+			logrus.Debugf("Error while parsing cluster version: %s", err)
+		}
+		parsedSeptK8s127Version, err := getClusterVersion(SeptK8s127Version)
+		if err != nil {
+			logrus.Debugf("Error while parsing cluster version: %s", err)
+		}
+		if (parsedVersion.Major == parsedSeptK8s125Version.Major && parsedVersion.Minor == parsedSeptK8s125Version.Minor && parsedVersion.GE(parsedSeptK8s125Version)) ||
+			(parsedVersion.Major == parsedSeptK8s126Version.Major && parsedVersion.Minor == parsedSeptK8s126Version.Minor && parsedVersion.GE(parsedSeptK8s126Version)) ||
+			parsedVersion.GE(parsedSeptK8s127Version) {
+			Binds = append(Binds, "/var/log/calico/cni:/var/log/calico/cni:z")
```
manuelbuil, line 591: Just for clarification, adding calico logs have nothing to do with felix config right? This is an extra thing, or?

go.mod
```diff
@@ -36,6 +36,7 @@ require (
 	github.com/mattn/go-colorable v0.1.8
 	github.com/mcuadros/go-version v0.0.0-20180611085657-6d5863ca60fa
 	github.com/pkg/errors v0.9.1
+	github.com/projectcalico/api v0.0.0-20220722155641-439a754a988b
```
manuelbuil, line 39: Why did you choose such an old commit?

manuelbuil (changes requested):
good job! It seems difficult to handle it!

PR #3413, opened 2023-10-16
add logic for external aws cloud provider

kinarashah:
**Issue:** https://github.com/rancher/rancher/issues/43175 
One of the prerequisites of external (out-of-tree) aws cloud provider is to follow node naming conventions.

> When IP-based naming is used, the nodes must be named after the instance followed by the regional domain name (ip-xxx-xxx-xxx-xxx.ec2.<region>.internal). `https://cloud-provider-aws.sigs.k8s.io/prerequisites/`

This can be managed by setting `hostname-override` to kubelet and kube-proxy for custom clusters via `--node-name`. But for node driver clusters, Rancher assigns hostname based on the node name generated with prefix from node pool names. This results into cloud controller manager unable to find the node by name: `Error getting instance metadata for node addresses: error fetching node by provider ID:`

**Fix:**

- Introduce a new cloud provider field name `external-aws`
- When cloud provider is set to `external-aws` and `useInstanceMetadataHostname` is enabled, rke-tools will fetch the hostname by querying ec2 metadata service `http://169.254.169.254/latest/meta-data/hostname`. rke-tools will use `hostname -f` if metadata service returns empty. This name will be set as hostname-override to kubelet and kube-proxy.
- `useInstanceMetadataHostname` is disabled by default. It'll need to be enabled by users to get the above mentioned behavior.

**Note:** 
- RKE looks up a node by matching `hostname-override` and `"kubernetes.io/hostname"`. This needed to be updated with this fix because node lookup would fail otherwise. It is updated only when cloud provider is `external-aws` to avoid regressions.


kinarashah:
Need to merge https://github.com/rancher/rke-tools/pull/177 first and update KDM to use the latest `rke-tools` before merging this PR.

k8s/node.go
```diff
@@ -55,6 +56,13 @@ func GetNode(k8sClient *kubernetes.Clientset, nodeName string) (*v1.Node, error)
 			if strings.ToLower(node.Labels[HostnameLabel]) == strings.ToLower(nodeName) {
 				return &node, nil
 			}
+			if cloudProviderName == ExternalAWSCloudProviderName {
```
superseb, line 59: Please add debug logs as we are lacking these already a lot

superseb:
I assume the taint function(s) are unused as reason for removal

k8s/node.go
```diff
@@ -15,18 +15,19 @@ import (
...
-	InternalAddressAnnotation = "rke.cattle.io/internal-ip"
-	ExternalAddressAnnotation = "rke.cattle.io/external-ip"
-	AWSCloudProvider          = "aws"
-	MaxRetries                = 5
-	RetryInterval             = 5
+	HostnameLabel                = "kubernetes.io/hostname"
+	InternalAddressAnnotation    = "rke.cattle.io/internal-ip"
+	ExternalAddressAnnotation    = "rke.cattle.io/external-ip"
+	AWSCloudProvider             = "aws"
+	ExternalAWSCloudProviderName = "external-aws"
```
jiaqiluo, line 22: Why can not we use the existing AWSCloudProvider? Can you explain the difference between those two?

cluster/plan.go
```diff
@@ -466,7 +467,7 @@ func (c *Cluster) BuildKubeletProcess(host *hosts.Host, serviceOptions v3.Kubern
 	Command := c.getRKEToolsEntryPoint(host.OS(), "kubelet")
 	CommandArgs := map[string]string{
 		"client-ca-file": pki.GetCertPath(pki.CACertName),
-		"cloud-provider": c.CloudProvider.Name,
+		"cloud-provider": getCloudProviderName(c.CloudProvider.Name),
```
jiaqiluo, line 470: [K8s docs](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/#:~:text=%2D%2Dcloud%2Dprovider%20string) say the option will be removed from kubelet in k8s 1.24 or later.   
How will the removal affect us, especially when providing a cluster with k8s 1.24 and later?

jiaqiluo (changes requested):

k8s/node.go
kinarashah: @jiaqiluo `external-aws` is same as `external` for Kubernetes. For RKE, when cloud provider is set to `external-aws` and `useInstanceMetadataHostname` is enabled, `rke-tools` will fetch the hostname by querying ec2 metadata service `http://169.254.169.254/latest/meta-data/hostname`. This name will be set as hostname-override to `kubelet` and `kube-proxy`.

We cannot make this amazon specific change under `external` as it's generic for all cloud providers. But it's required for node driver clusters to work (error in issue description), so `external-aws` is introduced.

cluster/plan.go
kinarashah: @jiaqiluo It still isn't removed and continues to work with all cloud providers. For external providers, Kubernetes still expects this flag to be set. 

> `kubelet`, `kube-apiserver` and `kube-controller-manager` must be set according to the user's usage of external CCM. If the user has an external CCM (not the internal cloud controller loops in the Kubernetes Controller Manager), then `--cloud-provider=external` must be specified. Otherwise, it should not be specified.

https://kubernetes.io/docs/tasks/administer-cluster/running-cloud-controller/

k8s/node.go
kinarashah: Added `logrus.Debugf("Checking hostname address for node [%v], cloud provider: %v", nodeName, cloudProviderName)`

k8s/node.go
jiaqiluo: @kinarashah, what is the use case for the existing `aws`?  Why can we not use `aws` and `useInstanceMetadataHostname` to achieve the goal?

superseb (dismissed):

jiaqiluo (approved):

k8s/node.go
kinarashah: @jiaqiluo That's a good idea and I think it'd have worked if we had kept it as `aws` and updated code to consider it as `external` for k8s>=1.27 + documented the behavior. Though from regression point of view and timing of the release, right now the change is all scoped under `external-aws` vs updating existing behavior for `aws` so it seems safer. 

Also this is not too critical but it would mean that existing k8s 1.26 clusters wouldn't be able to use this change because of k8s version checks with `aws` but they'd be able to migrate using `external-aws`.

cluster/defaults.go
```diff
@@ -1022,11 +1023,25 @@ func (c *Cluster) setCloudProvider() error {
...
+		}
+		if c.CloudProvider.Name == aws.AWSCloudProviderName {
+			clusterVersion, err := getClusterVersion(c.Version)
+			if err != nil {
+				return fmt.Errorf("failed to get cluster version for checking cloud provider: %v", err)
+			}
+			// cloud provider must be external or external-aws for >=1.27
+			defaultExternalAwsRange, err := semver.ParseRange(">=1.27.0-rancher0")
+			if err != nil {
+				logrus.Warnf("Failed to parse semver range for checking cloud provider")
```
jiaqiluo: Should we return an error here?

cluster/defaults.go
```diff
@@ -1022,11 +1023,25 @@ func (c *Cluster) setCloudProvider() error {
...
 		if err != nil {
-			return fmt.Errorf("Failed to parse cloud config file: %v", err)
+			return fmt.Errorf("failed to parse cloud config file: %v", err)
 		}
 		c.CloudProvider.Name = p.GetName()
 		if c.CloudProvider.Name == "" {
-			return fmt.Errorf("Name of the cloud provider is not defined for custom provider")
+			return fmt.Errorf("name of the cloud provider is not defined for custom provider")
+		}
+		if c.CloudProvider.Name == aws.AWSCloudProviderName {
```
jiaqiluo, line 1032: Is `external-aws` valid when the cluster version is < 1.27? Do we need to add a check similar to the existing one?

jiaqiluo (changes requested):

cluster/defaults.go
kinarashah: Yeah, it's valid as out-of-tree is supported for <1.27 but wasn't mandatory. There could be use cases where users might migrate to external first before upgrading to 1.27 so only validating for aws in-tree.

cluster/defaults.go
kinarashah: thanks, updated

cluster/defaults.go
jiaqiluo: Right, it makes sense to allow it for migration.

jiaqiluo (approved):

jiaqiluo (approved):

Issue #3396, opened 2023-09-27
Reduce Drone CI time/resources


superseb:
We need to optimize Drone CI for this repository as it takes too long to do certain things:

- When there are no code changes or anything changes that Drone needs to check, we need to skip/complete the run immediately.
- On release, we don't need to run the full CI suite as it should already have been checked after the last merge/RC. This should speed up making releases.

cwayne18:
We have some prior art that we can re-use here from `rancher/rke2`:
https://github.com/rancher/rke2/blob/0f71c9a47b030bf152d541db52e85f05e2473ced/.drone.yml#L14
https://github.com/rancher/rke2/blob/master/.droneignore

jiaqiluo:
The linked issue PR has been merged, and a few releases have been done since then, so it is safe to call it done and close the issue.

PR #3417, opened 2023-10-24
Clean/speed up Drone

superseb:
https://github.com/rancher/rke/issues/3396

- Adds a `skipfiles` step taken from [rke2](https://github.com/rancher/rke2/blob/0f71c9a47b030bf152d541db52e85f05e2473ced/.drone.yml#L14), which will skip the pipeline if only non code changes are present (exitcode 78 skips the whole pipeline in Drone)
- Removes running `script/integration` in `scripts/ci`, adds `scripts/integration-ci` to be run as a separate step. This makes it possible to have it run on PR but not on tag.
- Removes all the other integration steps on tag as well, tagging is confirming an existing good state so we don't have to waste time and resources when cutting a tag.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -7,6 +7,24 @@ platform:
   arch: amd64
 
 steps:
+- name: skipfiles
+  image: plugins/git
+  commands:
+  - export NAME=$(test $DRONE_BUILD_EVENT = pull_request && echo remotes/origin/${DRONE_COMMIT_BRANCH} || echo ${DRONE_COMMIT_SHA}~)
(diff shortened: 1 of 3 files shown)


manuelbuil (approved):
great!

jiaqiluo (approved):
LGTM

jiaqiluo:
still valid

manuelbuil:
@jiaqiluo @thomasferrandiz What do you think about merging?

jiaqiluo:
@manuelbuil, be free to merge it!  Can you also handle the backport if it is necessary?

PR #3443, opened 2023-11-27
Fix node lookup for external aws

kinarashah:
Problem: 
Current logic of matching `Hostname` with node name doesn't work, it is true for all nodes and we won't have always the correct hostname (set by cloud provider) under `hostname-override` in the cluster config.

Fix: 
Searching node by `InternalIP` because `ExternalIP` is populated by cloud provider in `node.Status.Addresses` after aws cloud controller manager is deployed. 

Issue: 
https://github.com/rancher/rancher/issues/43624

### DIFF
--- a/cluster/remove.go
+++ b/cluster/remove.go
@@ -11,6 +11,7 @@ import (
 	v3 "github.com/rancher/rke/types"
 	"github.com/rancher/rke/util"
 	"golang.org/x/sync/errgroup"
+	v1 "k8s.io/api/core/v1"
 )
 
 func (c *Cluster) ClusterRemove(ctx context.Context) error {
@@ -92,7 +93,13 @@ func (c *Cluster) RemoveOldNodes(ctx context.Context) error {
 		host := &hosts.Host{}
 		host.HostnameOverride = node.Name
 		if !hosts.IsNodeInList(host, uniqueHosts) {
-			if err := k8s.DeleteNode(kubeClient, node.Name, c.CloudProvider.Name); err != nil {
+			nodeAddress := ""
+			for _, addr := range node.Status.Addresses {
+				if addr.Type == v1.NodeInternalIP {
+					nodeAddress = addr.Address
+				}
+			}
+			if err := k8s.DeleteNode(kubeClient, node.Name, nodeAddress, c.CloudProvider.Name); err != nil {
 				log.Warnf(ctx, "Failed to delete old node [%s] from kubernetes")
 			}
 		}
--- a/k8s/node.go
+++ b/k8s/node.go
@@ -24,10 +24,10 @@ const (
 	RetryInterval                = 5
 )
 
-func DeleteNode(k8sClient *kubernetes.Clientset, nodeName, cloudProviderName string) error {
+func DeleteNode(k8sClient *kubernetes.Clientset, nodeName string, nodeAddress string, cloudProviderName string) error {
 	// If cloud provider is configured, the node name can be set by the cloud provider, which can be different from the original node name
 	if cloudProviderName != "" {
-		node, err := GetNode(k8sClient, nodeName, cloudProviderName)
+		node, err := GetNode(k8sClient, nodeName, nodeAddress, cloudProviderName)
 		if err != nil {
 			return err
 		}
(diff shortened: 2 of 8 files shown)


cluster/remove.go
```diff
@@ -92,7 +93,13 @@ func (c *Cluster) RemoveOldNodes(ctx context.Context) error {
...
 		host.HostnameOverride = node.Name
 		if !hosts.IsNodeInList(host, uniqueHosts) {
-			if err := k8s.DeleteNode(kubeClient, node.Name, c.CloudProvider.Name); err != nil {
+			nodeAddress := ""
+			for _, addr := range node.Status.Addresses {
+				if addr.Type == v1.NodeInternalIP {
+					nodeAddress = addr.Address
+				}
+			}
+			if err := k8s.DeleteNode(kubeClient, node.Name, nodeAddress, c.CloudProvider.Name); err != nil {
```
jiaqiluo, line 102: Is it possible there is no internal IP on the node?  I think the answer is no, but I would like to get your confirmation.

k8s/node.go
```diff
@@ -57,9 +57,10 @@ func GetNode(k8sClient *kubernetes.Clientset, nodeName, cloudProviderName string
 				return &node, nil
 			}
 			if cloudProviderName == ExternalAWSCloudProviderName {
-				logrus.Debugf("Checking hostname address for node [%v], cloud provider: %v", nodeName, cloudProviderName)
+				logrus.Debugf("Checking internal address for node [%v], cloud provider: %v", nodeName, cloudProviderName)
 				for _, addr := range node.Status.Addresses {
-					if addr.Type == v1.NodeHostName && strings.ToLower(node.Labels[HostnameLabel]) == addr.Address {
+					if addr.Type == v1.NodeInternalIP && nodeAddress == addr.Address {
```
jiaqiluo, line 65: Continuing with my preview comment, even it will never happen, but to avoid any possible regression,  I suggest adding a check in the `GetNode` function to return error if `nodeAddress` is (somehow) an empty string.

jiaqiluo (changes requested):
Overall LGTM, with a few comments.

superseb (dismissed):

k8s/node.go
a-blender: I agree with this. Logic in the caller function should output the err already if you return it from here. Otherwise lgtm

a-blender (changes requested):

k8s/node.go
kinarashah: @jiaqiluo Addressing both comments here since they're related. 
`nodeAddress` or `internalAddress` won't be empty because rke defaults to `address`/`externalIP` when it's empty in cluster/defaults. 
https://github.com/rancher/rke/blob/c56fb8c42a079d569dd402c1243a5ae16b4004c9/cluster/defaults.go#L225

rke sets `node-ip` for kubelet and kubelet kinda uses it to set the `internalIP` (after some other checks) so `internalIP` shouldn't be empty either. 
https://github.com/rancher/rke/blob/c56fb8c42a079d569dd402c1243a5ae16b4004c9/cluster/plan.go#L495
https://github.com/kubernetes/kubernetes/blob/ad9b60e2c9ddb21e8b00cabbe27e639638a0ea88/pkg/kubelet/nodestatus/setters.go#L231 

I can still add check for empty fields check, but I don't have fallback field to find the node instead with. Other addresses like private dns and external ip get set by aws cloud controller manager, but since that's deployed after cluster provisions we only have internal ip and hostname to find the node against. So in this case, it'll error out with `NotFound` error. What do you think?

jiaqiluo: @kinarashah, thank you for the explanation. I think we should return an error saying that the provided IP can not be empty, instead of NotFound.

kinarashah: @jiaqiluo the provided IP or nodeAddress could come from user but would also be set by RKE if it's empty, but since it's crucial to finding node, I'll add a check to be safe. 
cc @a-blender I think your comment was also around the same block, added check for empty nodeAddress and return error.

jiaqiluo (approved):
LGTM

k8s/node.go
a-blender: @kinarashah Thanks for explaining! Since it's so important for provisioning, we want to make sure we handle bad data invalid/empty str for node IP set by the user. Lgtm now

a-blender (approved):

PR #3448, opened 2023-12-05
Add ADR about RKE1-KDM decouple

manuelbuil:


rancher-max (approved):
I really like this. There's nothing glaring that I can see from it and I'm happy to try it!

adr/RKE-KDM-decouple.md
```diff
@@ -0,0 +1,185 @@
...
+
+When merged in KDM, the default URL in RKE1 is changed, pointing at the previously described branch. Here is [an example](https://github.com/rancher/rke/commit/ca8b578f55ffd46c5ff9485a4fcedcb9107f9425).
+
+Once the default URL PR is merged, RKE1 releases an RC. This RC is used by QA (infracloud team) to run its RKE1/KDM tests. If a regression is found, KDM or RKE1 code is updated and a new RC is released. Once QA signs off an RC, that RC can be considered for an "official" RKE1 release. However, we release RKE1 until KDM does an final release. KDM is consumed by other projects, such as Rancher Manager or RKE2, so this final KDM release can be delayed.
+
+When KDM releases, it merges the changes of `dev-vX.Y-YEAR-MONTH-patches` into `release-vX.Y` and the default URL in RKE1 is changed again. Here is [an example](https://github.com/rancher/rke/commit/6e8ec25f09f2ff88f66b0fe643f9d40b84f43756). After that, RKE1 is officially released.
+
+Note that, as of today, RKE1 does not intend to quickly release upstream Kubernetes releases. By the time the KDM branch is created, it is normal that upstream Kubernetes releases are already out.
```
kinarashah: > Note that, as of today, RKE1 does not intend to quickly release upstream Kubernetes releases 

This line sounds a bit off. We can probably just say KDM branches are created after upstream Kubernetes releases are out.

adr/RKE-KDM-decouple.md
```diff
@@ -0,0 +1,185 @@
...
+Then, a PR is created updating that KDM branch:
+* The new images that will be used (both Kubernetes and addons)
+* The new manifests if needed (e.g. if there is a new addon version that changes the manifest)
+* The logic to pick up manifests if needed (e.g. if there is a new addon manifest)
+
+Here is [an example](https://github.com/rancher/kontainer-driver-metadata/pull/1257/files)
+
+When merged in KDM, the default URL in RKE1 is changed, pointing at the previously described branch. Here is [an example](https://github.com/rancher/rke/commit/ca8b578f55ffd46c5ff9485a4fcedcb9107f9425).
+
+Once the default URL PR is merged, RKE1 releases an RC. This RC is used by QA (infracloud team) to run its RKE1/KDM tests. If a regression is found, KDM or RKE1 code is updated and a new RC is released. Once QA signs off an RC, that RC can be considered for an "official" RKE1 release. However, we release RKE1 until KDM does an final release. KDM is consumed by other projects, such as Rancher Manager or RKE2, so this final KDM release can be delayed.
```
kinarashah: > However, we release RKE1 until KDM does an final release 

However we don't release RKE1 until KDM officially releases

adr/RKE-KDM-decouple.md
```diff
@@ -0,0 +1,185 @@
...
+* Same as with the current release, RKE1 releases an RC and gets tested by QA
+* If QA signs off, that RC becomes the official release
+
+In other words, RKE1 will not wait until Rancher Manager is thorough with its testing to make KDM release official
+
+### Potential problems:
+
+1 - After RKE1 release, `dev-vX.Y-YEAR-MONTH-patches` gets a new PR because rke2 (or other project) requires something else (e.g. bugfix)
+
+RKE1 would have been released using an old `dev-vX.Y-YEAR-MONTH-patches`. However, the difference would be rke2 related code. Therefore, the important RKE1 code in KDM will still be the same. The difference will exist for a short span of time, as it will be removed in the next monthly release `dev-vX.Y-YEAR-MONTH+1-patches` which will include the rke2 bits
```
kinarashah, line 151: Yeah, RKE1 don't technically care about the the changes in KDM related to rke2, but as of now, we have build logic that checks if the data is the latest or it fails. Could you call this out? We'd need to fix this as a part of changing our process.

manuelbuil: thanks for this, I was not aware of it. Could you point me to where are we doing that check? Is it automatic?

kinarashah: It's in our CI, https://github.com/rancher/rke/blob/267a46d917409dc9f037c3283a079ea48f64af3b/scripts/validate#L19

kinarashah: we should separate rke1 and rke2 in generate logic so data.json changes only if rke1 info changes

manuelbuil: Sorry Kinara, I still don't understand :(.

If the RKE1's defaultURL points at KDM's branch: `dev-vX.Y-YEAR-MONTH-patches`, `go generate` will always pick up data.json from that branch, or? I was expecting to see in that `validate` script a comparison between:
* data.json from `defaultURL`
* data.json from KDM's official release

but it just calls `go generate`, so I assume it will still use defaultURL, right? In that case it should work with my proposal.

In my proposal, I am not saying that RKE1 keeps its own data.json, unsync from KDM. What I am saying is that it uses KDM's branch: `dev-vX.Y-YEAR-MONTH-patches` to release RKE1 (after RKE1 validates it). So RKE1's data.json will always be in sync with KDM's `dev-vX.Y-YEAR-MONTH-patches` branch

caroline-suse-rancher:
Let's keep this open until we have a chance to revisit this :)

PR #3472, opened 2024-01-10
[v1.28] Bump k8s dependencies to support k8s v1.28

chiukapoor:
### Update
* Bump k8s dependencies to 0.28.6

### Related
* https://github.com/rancher/rancher/issues/43109


kinarashah (changes requested):
@chiukapoor Could you point kdm url to your KDM branch in codegen/codegen.go and run `go generate` so we can see tests passing for 1.28?

HarrisonWAffel (approved):

kinarashah (approved):

PR #3476, opened 2024-01-17
change certificate and key file permission to 0600

vardhaman22:
related issue: https://github.com/rancher/rancher/issues/42012


rbrtbnfgl (approved):

vardhaman22:
the permission related change needed to fix the cis 4.1.7 test, needs to go in rancher/rke-tools [cert-deployer](https://github.com/rancher/rke-tools/blob/master/cert-deployer) script, changing permission in rke will not be sufficient.

Issue #3485, opened 2024-01-25
[Backport] [v1.4] etcd snapshot-restore with etcd v3.5.9 fails due to missing /bin/sh in etcd container image


vardhaman22:
backport: https://github.com/rancher/rke/issues/3382


RKE version:
v1.4.13 but any k8s version that is using etcd v3.5.7 and above (currently k8s v1.26.11)

Steps to Reproduce:

run rke etcd snapshot-save --name yolo
run rke etcd snapshot-restore --name yolo
**Results:**
```
INFO[0033] Starting container [etcd-restore] on host [54.203.164.32], try #1                                                                                                                                                                                                                                                  
WARN[0034] Can't start Docker container [etcd-restore] on host [54.203.164.32]: Error response from daemon: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "sh": executable file not found in $PATH: unknown
```
Cause is etcd moving to distroless base image https://github.com/etcd-io/etcd/pull/15016

rishabhmsra:
Using standalone RKE v1.4.14-rc2 - fresh install
- Created v1.26.13-rancher1-1 (1-cp, 1-e, 1-w), performed backup after creating workloads - it was successful.
- Performed restore and it was successful:
```
INFO[0203] Restarting network, ingress, and metrics pods 
INFO[0205] Finished restoring snapshot [new_bkp1] on all etcd hosts
```
Using standalone RKE v1.4.14-rc2 - upgrade check:
- Created v1.25 cluster
- Created workloads.
- Perform backup
- Upgraded cluster to v1.26
- Created workloads and then performed backup and it was successful

Both scenarios passed, hence closing it.

PR #3486, opened 2024-01-25
[v1.4] changed etcd restoreImage to rke-tools for etcd >=3.5.7

vardhaman22:
issue: https://github.com/rancher/rke/issues/3485 / https://github.com/rancher/rancher/issues/44218
since etcd is moved to distroless base image from v3.5.7 so /bin/sh, mv and rm tools will not be present in etcd Image. which is required by the restore command. [ref](https://github.com/rancher/rke/blob/d2b01281f33970e5980d307935474dac6a9833cd/services/etcd.go#L652-L670)
One solution for that is to use a custom image which has etcdctl as well as these other required tools.
Since in rke we use rke-tools image to create etcd snapshots, i have changed restore image to rke-tools as well. since rke-tools has etcdctl with the other required deps for that command.

cherry-pick of https://github.com/rancher/rke/pull/3390

### DIFF
--- a/cluster/etcd.go
+++ b/cluster/etcd.go
@@ -16,6 +16,8 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+const MinEtcdVersionWithDistrolessImage = "v3.5.7"
+
 func (c *Cluster) SnapshotEtcd(ctx context.Context, snapshotName string) error {
 	backupImage := c.getBackupImage()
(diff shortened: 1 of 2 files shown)


snasovich (approved):
Matches the changes from https://github.com/rancher/rke/pull/3486#issue-2099640563

jiaqiluo (approved):
LGTM

PR #3495, opened 2024-02-02
[v2.8] updated containerd and runc dependencies

vardhaman22:
https://github.com/rancher/rancher/issues/44287


snasovich (dismissed):
LGTM - there are some additional dependency bumps but assuming they are due to runc/containerd bumps and are at the same or lower versions than same deps bumped or already existing per similar `rancher/rancher` change - https://github.com/rancher/rancher/pull/44268

kinarashah (dismissed):

kinarashah (approved):

PR #3531, opened 2024-03-13
Restart kube-apiserver in all CP nodes after changing the Pod Security Admission Configuration

jiaqiluo:
# Problem
 
In Rancher, enabling/changing PSACT on an RKE1 node-driver cluster doesn't restart the `kube-apiserver` container in all Control Plane Nodes.

The root cause is that RKE does not handle the restarting of the `kube-apiserver` container when the admission configuration file is changed properly: the current implementation restarts the container in only one of the CP nodes. 

# Solution

This PR fixes the issue that kube-apiserver does not restart in all CP nodes after changing the Pod Security Admission Configuration.

Now, a new environment variable RKE_ADMISSION_CONFIG_CHECKSUM whose value is the checksum of the content of the admission configuration file is added to the env list that is set in the `kube-apiserver` container configuration, so any changes in the admission configuration file will result in a change in the container's configuration. RKE will detect the changes during reconciliation and therefore restart the kube-apiserver container on all CP nodes. The cadence is controlled by the upgrade strategy in the cluster. 

This PR also drops the unnecessary appending of env var to the cluster object which shows in the cluster.rkestate file as the following:  
Screenshot 2024-03-13 at 4 37 49 PM

 
# Testing

To prepare for the test, I used the RKE binary built from this branch to provision a 3-node all-role cluster. 

Then I performed the following tests by editing the cluster.yml and running `rke up`; after each test case, I verified that  1) the kube-apiserver container is restarted in all three CP nodes and 2) the file `/etc/kubernetes/admission.yaml` or `/etc/kubernetes/audit-policy.yaml` is changed to match the cluster.yml in all three CP nodes:

case 1. add the admission configuration under `.services.kube-api.admission_configuration` 
case 2. change the above admission configuration 
case 3. drop the above admission configuration 
case 4. set `services.kube-api.pod_security_configuration` to `privileged`
case 5. change `services.kube-api.pod_security_configuration` to `restricted` 
case 6. add an audit log configuration under `.services.kube-api.audit_log`
case 7. change the above audit log configuration

### DIFF
--- a/cluster/plan.go
+++ b/cluster/plan.go
@@ -40,10 +40,11 @@ const (
 
 	NetworkConfigurationEnv = "RKE_NETWORK_CONFIGURATION"
 
-	EtcdPathPrefix       = "/registry"
-	CloudConfigSumEnv    = "RKE_CLOUD_CONFIG_CHECKSUM"
-	CloudProviderNameEnv = "RKE_CLOUD_PROVIDER_NAME"
-	AuditLogConfigSumEnv = "RKE_AUDITLOG_CONFIG_CHECKSUM"
+	EtcdPathPrefix        = "/registry"
+	CloudConfigSumEnv     = "RKE_CLOUD_CONFIG_CHECKSUM"
+	CloudProviderNameEnv  = "RKE_CLOUD_PROVIDER_NAME"
+	AuditLogConfigSumEnv  = "RKE_AUDITLOG_CONFIG_CHECKSUM"
+	AdmissionConfigSumEnv = "RKE_ADMISSION_CONFIG_CHECKSUM"
 
 	DefaultToolsEntrypoint        = "/opt/rke-tools/entrypoint.sh"
 	DefaultToolsEntrypointVersion = "0.1.13"
@@ -202,6 +203,7 @@ func (c *Cluster) BuildKubeAPIProcess(host *hosts.Host, serviceOptions v3.Kubern
 		"tls-private-key-file":          pki.GetKeyPath(pki.KubeAPICertName),
 	}
 	CommandArrayArgs := make(map[string][]string, len(c.Services.KubeAPI.ExtraArgsArray))
+	Env := make([]string, len(c.Services.KubeAPI.ExtraEnv))
 
 	if len(c.CloudProvider.Name) > 0 {
 		CommandArgs["cloud-config"] = cloudConfigFileName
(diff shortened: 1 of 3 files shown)


cmd/up.go
```diff
@@ -200,10 +200,6 @@ func ClusterUp(ctx context.Context, dialersOptions hosts.DialersOptions, flags c
 		return APIURL, caCrt, clientCert, clientKey, nil, err
 	}
 
-	if err := cluster.RestartKubeAPIServerWhenConfigChanges(ctx, kubeCluster, currentCluster); err != nil {
```
kinarashah, line 203: @jiaqiluo The main reason for just one api server to restart is cuz we return here, right? So wouldn't it work if we restarted kube api for all hosts here? Or the problem is only DeepEqual comparison isn't sufficient to decide when config changes?

jiaqiluo: @kinarashah,  this approach breaks the rolling upgrade (or zero-downtime upgrade) for control plane nodes: it will restart the kube-apiserver container on all CP nodes at once( if I remove the return). That is the main reason why I decided to use the "adding an env var to kube-apiserver's docker configuration" approach.

kinarashah (approved):

cluster/plan.go
```diff
@@ -328,12 +336,14 @@ func (c *Cluster) BuildKubeAPIProcess(host *hosts.Host, serviceOptions v3.Kubern
 	}
 	registryAuthConfig, _, _ := docker.GetImageRegistryConfig(c.Services.KubeAPI.Image, c.PrivateRegistriesMap)
 
+	Env = append(Env, c.Services.KubeAPI.ExtraEnv...)
```
jakefhyde, line 339: I'm assuming order doesn't matter here (since we are reordering the environment variable putting `extraEnv` at the end instead of the beginning), correct?

kinarashah: yep should be ok, upgrade check uses `sliceEqualsIgnoreOrder` under the hood https://github.com/rancher/rke/blob/1951add95d62585b8e21e191aea5ef0de516343f/docker/docker.go#L583 

will let Jack double check and confirm tho

jiaqiluo: It is correct. The order here does not matter.  I wrote it this way at first because I planned to modify the getUniqStringList function to split each item into (key, value) pair and keep only the first occurrence of the key, but then I realized that it was unnecessary.

jakefhyde (approved):

PR #3536, opened 2024-03-20
[v1.5] add timeout to etcd member add

kinarashah:
# Issue: https://github.com/rancher/rancher/issues/45539

- Etcd client v3 kept retrying on member add vs v2 client used to return errors so we didn't have timeout earlier 
- DialTimeout alone doesn't suffice, also need to specify client request timeout with context
- Updated to close all etcd clients as v3 client uses grpc to connect to etcd, could leave us with leaks if not closed

### DIFF
--- a/services/etcd.go
+++ b/services/etcd.go
@@ -20,6 +20,7 @@ import (
 	"github.com/rancher/rke/util"
 	"github.com/sirupsen/logrus"
 	etcdclientv2 "go.etcd.io/etcd/client/v2"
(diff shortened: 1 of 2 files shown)


zmedico:
Hi, I've also experienced etcd scaling issues with rancher-2.8.x that I did not experience with rancher-2.7.x, and this timeout for the v3 etcd client looks very promising, thanks!

jakefhyde (approved):

HarrisonWAffel (approved):

Issue #3570, opened 2024-04-30
Errors while removing an etcd member can cause RKE to try to remove the member indefinitely


jakefhyde:
**RKE version:** v1.4.10

**Steps to Reproduce:**
- Provision a cluster with 3 etcd members
- Using `etcdctl`, remove a member and then attempt to remove the etcd node from RKE

**Results:**
RKE will continuously attempt to remove the etcd member with the following errors:

```
{"level":"warn","ts":"2024-02-07T09:42:15.024Z","logger":"etcd-client","caller":"v3@v3.5.4/retry_interceptor.go:62","msg":"retrying of unary invoker failed","target":"etcd-endpoints://0x<hex>/<ip>:2379","attempt":0,"error":"rpc error: code = NotFound desc = etcdserver: member not found"}

[ERROR] cluster [c-xxxxx] provisioning: [reconcile] Failed to delete etcd member [hostname] from etcd cluster
```

This can also happen if when an etcd member remove is issued via RKE, which succeeds however an error may be returned if the request is not completed in time, the connection is dropped, or some other issue occurs.

RKE1 should confirm the etcd member exists before aatempting to delete it.

SURE-7601

jakefhyde:
Blocked, waiting for RKE1 1.6 rc.

snasovich:
Closing as tracking `rancher/rancher` issue got validated - https://github.com/rancher/rancher/issues/45309#issuecomment-2181502432

PR #3573, opened 2024-05-01
check if etcd member in list before removal

jakefhyde:
https://github.com/rancher/rke/issues/3570
https://github.com/rancher/rancher/issues/45309

If a member has been removed from etcd (either externally, or internally but an error was returned), RKE will continuously attempt to remove the member from etcd despite it no longer being a member.

### DIFF
--- a/services/etcd.go
+++ b/services/etcd.go
@@ -222,76 +222,90 @@ func etcdClientV3Range(k8sVersion string) bool {
 	return etcdv3Range(toMatch)
 }
 
+func removeEtcdMemberV3(ctx context.Context, toDeleteEtcdHost *hosts.Host, host *hosts.Host, localConnDialerFactory hosts.DialerFactory, cert, key []byte) bool {
+	etcdClient, err := getEtcdClientV3(ctx, host, localConnDialerFactory, cert, key)
+	if err != nil {
+		logrus.Debugf("Failed to create etcd client for host [%s]: %v", host.Address, err)
+		return false
+	}
+	members, err := etcdClient.MemberList(ctx)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


kinarashah (approved):

services/etcd.go
```diff
@@ -242,14 +242,19 @@ func RemoveEtcdMember(ctx context.Context, toDeleteEtcdHost *hosts.Host, etcdHos
...
 			for _, member := range members.Members {
 				if member.Name == fmt.Sprintf("etcd-%s", toDeleteEtcdHost.HostnameOverride) {
-					mIDv3 = member.ID
+					mIDv3 = &member.ID
 					break
 				}
 			}
-			if _, err := etcdClient.MemberRemove(ctx, mIDv3); err != nil {
-				logrus.Debugf("Failed to list etcd members from host [%s]: %v", host.Address, err)
+			if mIDv3 == nil {
```
jiaqiluo: `mIDv3` is declared outside the for loop, if the 1st interaction ( on the 1st host) fails but the etcd node is removed either by an external tool or the failure is due to a network delay, then when the 2nd interaction runs it will not find the member but the `mIDv3` is not nill so it invokes `MemberRemove` which will fail. Will we run into the same issue as the PR tries to fix?  We can avoid the case by declaring the variable inside the for loop.  The same to `mIDv2`.

jiaqiluo (changes requested):

services/etcd.go
jakefhyde: Refactored it a bit so this won't be an issue.

jiaqiluo (approved):

jiaqiluo:
LGTM. It would be great if you could squash commits into one before merging.

kinarashah (approved):

PR #3591, opened 2024-05-24
Move from drone to Github Actions for 1.6

mitulshah-suse:
ref issue: https://github.com/rancher/rancher/issues/45377

- setup runners to use custom runners provided by EIO, since using ubuntu-latest is causing diskpressure issues https://github.com/rancherlabs/eio/issues/2338
- removed dapper containers, since it is causing failures in pulling images from docker hub. `resolv.conf` file gets mounted with incorrect entries for dns which cannot be overridden directly. ref: https://github.com/rancher/rke/blob/release/v1.5/dind/dind.go#L77-L78
- `rke remove` kept failing with `directory not empty` errors in docker. Updated the code to first stop the container and then remove it with a pause in between. ref: https://github.com/moby/moby/issues/8203#issuecomment-56625506


HarrisonWAffel (dismissed):

.github/workflows/fossa.yaml
```diff
@@ -0,0 +1,29 @@
+name: Fossa scan
+
+on:
+    push: 
+      branches:
+        - '*'
```
jiaqiluo:
```suggestion
        - 'release/v*'
```

.github/workflows/workflow.yaml
```diff
@@ -0,0 +1,145 @@
...
+          name: rke-binaries-${{ github.run_number }}-${{ github.run_attempt }}
+          path: build/bin
+      - name: Create pre-release
+        run: |
+          gh release create ${{ github.ref_name }} -p --verify-tag --title "Pre-release ${{ github.ref_name }}" --notes-file build/bin/rke-k8sversions.txt build/bin/rke*
+
+  github-release:
+    permissions:
+      contents: write
+      id-token: write
```
jiaqiluo: Do we need this permission id-token here?  Maybe you forgot to drop it?

jiaqiluo (changes requested):

.github/workflows/workflow.yaml
mitulshah-suse: yep.. forgot to drop it.. updated..

.github/workflows/fossa.yaml
mitulshah-suse: updated to '**' which should cover all branches

HarrisonWAffel (approved):

jiaqiluo (approved):

kinarashah (approved):

Issue #3587, opened 2024-05-15
Provisionning fails with RKE 1.5.9 when `extra_env` is used for `kube-api` service


tmsdce:
**Description**

I'm trying to create a cluster with extra_env specified for the kube-api service. This works fine with RKE 1.5.8 but fails with RKE 1.5.9. I think the following commit might be involved : https://github.com/rancher/rke/commit/2e767c8c28fac0b0de7fc1d75a06e956090f7fba

**RKE version: 1.5.9**

**Docker version: (`docker version`,`docker info` preferred)**

```
Client: Docker Engine - Community
 Version:    25.0.5
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.14.0
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.27.0
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 3
  Running: 2
  Paused: 0
  Stopped: 1
 Images: 10
 Server Version: 25.0.5
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: e377cd56a71523140ca6ae87e30244719194a521
 runc version: v1.1.12-0-g51d5e94
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.5.0-35-generic
 Operating System: Ubuntu 22.04.4 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 4
 Total Memory: 5.778GiB
 Name: ubu22
 ID: 7bc42587-26ab-4f66-83ec-56004d44d654
 Docker Root Dir: /var/lib/docker
 Debug Mode: true
  File Descriptors: 33
  Goroutines: 53
  System Time: 2024-05-15T15:03:08.602809828Z
  EventsListeners: 0
[...]

PR #3597, opened 2024-06-04
Avoid allocating empty strings in the Env slice

jiaqiluo:
https://github.com/rancher/rke/issues/3587
https://github.com/rancher/rancher/issues/45645

Problem:
Previously, we declared the Env slice with the size of the `c.Services.KubeAPI.ExtraEnv` field, which leads to empty strings at the beginning of the Env slice because we use Golang's append function to add new elements, and Docker is not happy with that. 

Fix:
We fix this issue by declaring the Env variable as an empty slice. 
We also enhance the `getUniqStringList` function to properly trim leading and trailing spaces in each element and to ignore empty strings.  
We add unit tests for  the `getUniqStringList` function  and update the integration tests. 

Dev validation:
Compile the code locally and run `rke up` succeeds with the following section in cluster.yaml file which failed before the fixing:
 
```
services:
  kube-api:
    extra_env:
    - TEST_VAR=test
```


jakefhyde (approved):
LGTM, nice job!

HarrisonWAffel (approved):

PR #3602, opened 2024-06-10
Sync `release/v1.4` with `release/v1.4.19`

bfbachmann:


codegen/codegen.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 const (
-	defaultURL = "https://releases.rancher.com/kontainer-driver-metadata/dev-v2.7/data.json"
```
jiaqiluo, line 13: there is no need to change the url to the release branch, unless you plan to have a following PR to change it back to dev (as a part of the post-release cleaning up).

jiaqiluo (changes requested):

PR #3603, opened 2024-06-10
Sync `release/v1.5` with `release/v1.5.10`

bfbachmann:


codegen/codegen.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 const (
-	defaultURL = "https://releases.rancher.com/kontainer-driver-metadata/dev-v2.8/data.json"
+	defaultURL = "https://releases.rancher.com/kontainer-driver-metadata/release-v2.8/data.json"
```
jiaqiluo, line 13: Same as the other PR, there is no need to change the url to the release branch, unless you plan to have a following PR to change it back to dev (as a part of the post-release cleaning up).

jiaqiluo (changes requested):

PR #3608, opened 2024-06-11
[release v1.6] K8s 1.30

vardhaman22:
issue: https://github.com/rancher/rancher/issues/45089

- removed gengo dependency
- inpur dirs flag removed and output-file-base cli option removed. so updated it to output-file option
https://github.com/kubernetes/code-generator/commit/4d6b3bd5ca9ac09408e6e37d251d14ef46c76d0e#diff-efe143f58bc289328d0d56ad4c6e64487f358f5c6407b43ab6c1a902e4149255
just appending these dirs at the end of the command worked fine.
- deepcopy gen moved to code-generator so updated the import in code gen
https://github.com/kubernetes/kubernetes/commit/cb7f0593aead2d9e5748d95e52e983e7eef93c62
- apiserver types moved from one dir to another so updated the imports accordingly
https://github.com/kubernetes/apiserver/commit/e7eedd15ec9dd1ab5349b05ddbb1530655a28388#diff-179ab44de4a16de69e23c8a3fee5a0d6ef22755b90c182744f05ca9e1eca4381


kinarashah (approved):

codegen/codegen.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 const (
-	defaultURL = "https://releases.rancher.com/kontainer-driver-metadata/dev-v2.9/data.json"
+	defaultURL = "https://raw.githubusercontent.com/vardhaman22/rancher-kontainer-driver-metadata/k8s-1.30/data/data.json"
```
jiaqiluo: if we can get https://github.com/rancher/kontainer-driver-metadata/pull/1406 to be merged first, then do not need to change the URL here, right?

go.mod
```diff
@@ -1,24 +1,26 @@
 module github.com/rancher/rke
 
-go 1.22
+go 1.22.0
+
+toolchain go1.22.3
```
jiaqiluo, line 5: `Toolchain` is new to me, go1.22.3 seems to be for Go 1.22.3, not 1.22.0, so could you explain why it was chosen?

jiaqiluo (changes requested):

codegen/codegen.go
kinarashah: yeah, once kdm is merged, this url change would be removed. I asked for using the custom branch url to verify build is passing beforehand

go.mod
kinarashah: Afaik, toolchain version is required >= go version so v1.22.3 should be ok but good question in case there was something specific that needed 1.22.3 instead of latest 1.22.4

jiaqiluo: @kinarashah Thank you for the explanation. 

> A module’s go line must declare a version greater than or equal to the go version declared by each of the modules listed in require statements. 

Based on the above statement from the [Golang doc](https://go.dev/doc/toolchain), and the fact that rancher/rancher declares go1.22.3 ([code](https://github.com/rancher/rancher/blob/release/v2.9/go.mod#L5C11-L5C19)), it makes seems to use go1.22.3 here.

jiaqiluo (dismissed):

kinarashah (approved):

PR #3615, opened 2024-06-21
Move from drone to GHA for rke 1.5

mitulshah-suse:
ref: https://github.com/rancher/rancher/issues/45377


.github/workflows/workflow.yaml
```diff
@@ -0,0 +1,149 @@
...
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: 'go.mod'
+      - name: Run golangci-lint
+        uses: golangci/golangci-lint-action@v6
+        with:
+          version: v1.58
+      - name: Run CI
+        run: |
+          ./scripts/ci
```
jiaqiluo, line 42: Just want to mention that `./scripts/ci` runs `validate` which also runs `golangci-lint run`, so the same validate is executed twice.

.github/workflows/workflow.yaml
```diff
@@ -0,0 +1,149 @@
...
+      - '**.md'
+      - 'CODEOWNERS'
+      - 'LICENSE'
+      - 'docs/**'
+
+jobs:
+  ci:
+    permissions:
+      contents: read
+    runs-on: ubuntu-latest
```
jiaqiluo, line 27: This job fails on the release/v1.4 branch due to the error "run out of disk space".  Although it does pass here, can we change it to run on the non-containerized runner to avoid the potential degradation?
```
 runs-on: runs-on,runner=4cpu-linux-x64,run-id=${{ github.run_id }}
```

jiaqiluo (changes requested):

.github/workflows/workflow.yaml
mitulshah-suse: Yes i do know that.
The `/scripts/ci` lint run fails if the setup is not present which is installed by the github action `golangci/golangci-lint-action@v6`, so needed to add the step.
And i left the lint check in the `ci` script, in case someone is running it locally.
It doesnt take more than 1 minute, so not really an issue i belv.

.github/workflows/workflow.yaml
mitulshah-suse: 1.4 was running integration checks which have been moved to a separate job here. Have fixed it for 1.4 and it is passing there as well. Have added more details in the thread there https://github.com/rancher/rke/pull/3616#discussion_r1650635832
Have not faced this issue here or on 1.6 on multiple runs, so i belv it should be fine. I believe its better to stay on `ubuntu-latest` where possible since that is free.

.github/workflows/workflow.yaml
jiaqiluo: SGTM.

.github/workflows/workflow.yaml
```diff
@@ -0,0 +1,149 @@
...
+        env:
+          CROSS: 1
+      - name: Upload rke bin artifacts
+        if: github.event_name == 'push' && github.ref_type == 'tag'
+        uses: actions/upload-artifact@v4
+        with:
+          name: rke-binaries-${{ github.run_number }}-${{ github.run_attempt }}
+          path: build/bin/rke*
+          if-no-files-found: error
+          retention-days: 1
```
jakefhyde, line 53: Is `retention-days` 1 correct here? If so, why? Maybe we need a comment as to why this is necessary because it's not immediately obvious to me.

dind/dind.go
```diff
@@ -120,6 +121,15 @@ func RmoveDindContainer(ctx context.Context, dindAddress string) error {
 			return nil
 		}
 	}
+
+	timeout := 2 * time.Minute
```
jakefhyde, line 125: This could do with a comment as to why the change was necessary, perhaps even a separate commit.

jakefhyde (changes requested):

jakefhyde:
Assume all my review comments apply to https://github.com/rancher/rke/pull/3616 as well.

.github/workflows/workflow.yaml
mitulshah-suse: The default `retention-days` is 90 days for github actions and we dont need the artifacts stored beyond the life of the pipeline as whatever artifacts are needed beyond that are already published in the github release. 
So 1 was the minimum setting available, which is why i set it to that.
https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#artifact-and-log-retention-policy

dind/dind.go
mitulshah-suse: This PR is a backport of the the 1.6 issue where all the details are already added. You can refer to that for additional details of why the above was done.
https://github.com/rancher/rke/pull/3591#issue-2314585717

jakefhyde (approved):

PR #3616, opened 2024-06-21
Move from drone to GHA for rke 1.4

mitulshah-suse:
ref: https://github.com/rancher/rancher/issues/45377


.github/workflows/workflow.yaml
```diff
@@ -0,0 +1,149 @@
...
+      - '**.md'
+      - 'CODEOWNERS'
+      - 'LICENSE'
+      - 'docs/**'
+
+jobs:
+  ci:
+    permissions:
+      contents: read
+    runs-on: ubuntu-latest
```
jiaqiluo, line 27: I noticed this job fails because it runs out of disk space on [this run](https://github.com/rancher/rke/actions/runs/9615575617?pr=3616).  I am wondering if it is better to run on Rancher's own runner?

jiaqiluo (changes requested):
Overall LGMT, but the workflow  "CI workflow / ci" fails, so will not approve the PR until the CI passes.

.github/workflows/workflow.yaml
jiaqiluo: Note that the same error does not happen on the run on the release/v1.5 branch: https://github.com/rancher/rke/actions/runs/9609448943/job/26503951786?pr=3615

jiaqiluo: OK. I got confirmation from the EIO team that we should use non-containerized runners to avoid this error. 
```
 runs-on: runs-on,runner=4cpu-linux-x64,run-id=${{ github.run_id }}
```

mitulshah-suse: @jiaqiluo So i did a little further digging into why this issue was specific to 1.4 and found that `./integration` was being called from the `ci` script here which was not the case with 1.5 and 1.6.
Seems like they were part of some improvement changes which were done on 1.5 but not done for 1.4.
So, did similar changes here and that fixed the run. It does not need the additional space there anymore.

On a side note, yes the non-containerized runners have 125 GB space. Had gotten that added specifically for the integration checks that were failing here. https://github.com/rancherlabs/eio/issues/2338#issuecomment-2140952584

jiaqiluo (approved):

jakefhyde (approved):

PR #3657, opened 2024-08-20
[v1.6] Fixes for generating data.json

kinarashah:
Fixing two common issues around RKE1 release process: 

1. KDM generates the `data.json` file and includes RKE2/K3s version information, which is not used by RKE1 but causes a problem for the CI. The script validates if the `data.json` is the latest version from KDM. So we end up doing additional work on the RKE1 side even if changes are made to KDM only for RKE2/K3s versions. This is especially annoying during releases, as RKE1 validation might be complete, but we could still need to tag another RC due to updates in `data.json`. We could modify the script to ignore RKE2/K3s changes but ideally, this data shouldn’t be included at all.

2. We manually update the KDM URL to `../release-v2.x/data.json` for final releases, and then switch it back to `../dev-v2.x/data.json` after release. We could instead use the RKE version from the git tag to automatically decide the URL. Note that we might still need to manually update `defaultDevURL`, depending on the custom branch created in KDM for a release, but there is no way to know that beforehand. 

**Dev Testing:** 
- When CI fails requiring `go generate`, verify it's only for RKE1 related changes.
- Verify KDM url when tagging RCs and releases, CI will show the URL in logs: 
`Reading data from https://releases.rancher.com/kontainer-driver-metadata/dev-v2.9/data.json` 

**QA:** 
- Test RKE1 and RKE2 provisioning works in both normal and airgap Rancher setups.

### DIFF
--- a/codegen/codegen.go
+++ b/codegen/codegen.go
@@ -1,25 +1,27 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
 	"os"
+	"strings"
 
+	"github.com/coreos/go-semver/semver"
 	"github.com/rancher/rke/metadata"
 )
 
 const (
-	defaultURL = "https://releases.rancher.com/kontainer-driver-metadata/dev-v2.9/data.json"
-	dataFile   = "data/data.json"
+	defaultDevURL     = "https://releases.rancher.com/kontainer-driver-metadata/dev-v2.9/data.json"
+	defaultReleaseURL = "https://releases.rancher.com/kontainer-driver-metadata/release-v2.9/data.json"
+	dataFile          = "data/data.json"
 )
 
-// Codegen fetch data.json from https://releases.rancher.com/kontainer-driver-metadata/release-v2.7/data.json and generates bindata
+// Codegen fetch data.json from defaultURL or defaultReleaseURL and generates bindata
 func main() {
-	u := os.Getenv(metadata.RancherMetadataURLEnv)
-	if u == "" {
-		u = defaultURL
-	}
+	u := getURL()
+	fmt.Printf("Reading data from %s \n", u)
 	data, err := http.Get(u)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


codegen/codegen.go
```diff
@@ -31,8 +32,36 @@ func main() {
...
 	if err := os.WriteFile(dataFile, b, 0755); err != nil {
 		return
 	}
 }
+
+func getURL() string {
+	u := os.Getenv(metadata.RancherMetadataURLEnv)
+	if u == "" {
+		u = defaultDevURL
+		tag := os.Getenv("TAG")
```
vardhaman22, line 60: Do we need to make any changes in CI workflows so that this tag is correctly set here ?

vardhaman22 (commented):
just have a question about TAG env variable. Otherwise LGTM.

codegen/codegen.go
kinarashah: The tag already exists so I'm hoping we don't need any other changes: https://github.com/rancher/rke/blob/2e0a4a74647262497ad72e30d7e9fadb499ab846/Dockerfile.dapper#L14

vardhaman22: actually we are not using dapper in the CI for rke. And i checked the action workflows, the TAG is not being set there so we need to make some changes in that i guess.
https://github.com/rancher/rke/blob/2e0a4a74647262497ad72e30d7e9fadb499ab846/.github/workflows/workflow.yaml#L40-L43
we did not use dapper in CI because there were some network related issues using dapper for integration tests (dind) is used in that.

kinarashah: Thank you! Fixed to set `TAG` as env var in gha. 

I also noticed that `semver.NewVersion` errors out if tag starts with `v` so updated it and codegen_test.go to reflect this change. Could you re-review?

vardhaman22 (approved):
LGTM.
Just need to be rebased with release/v1.6

jiaqiluo (approved):

Issue #3656, opened 2024-08-14
use fixed stream server bind address for cri-dockerd


kinarashah:
Currently RKE doesn't set `streaming-bind-addr` when starting cri-dockerd: 
https://github.com/rancher/rke-tools/blob/f90a8996c5f2143b7b62a07444eb47f962354bad/entrypoint.sh#L118 

Fix it to set it to a known address, preferably to the same address `127.0.0.1:10010` as other runtimes do (for reference: https://github.com/k3s-io/k3s/pull/9975) 

Will need a change in RKE too if users can configure `streaming-bind-addr`, else the fix in rke-tools will suffice. Note that the change would be applied on kubelet restarts. 

SURE-8866

vardhaman22:
testing scenarios:
1. Creation from rancher and rke should set the default port (10010)
2. Upgrading (from july to august) on rancher should also set the default port(10010)
3. And setting the custom host port on an installed version should update them in the cluster
@kinarashah please suggest if anything else needs to be tested for this.

vivek-shilimkar:
Tested the scenario 1 and 3 with rancher v2.9-head and rke version 1.6.2-rc1. Both the scenarios set the port as expected.

rishabhmsra:
Validation is completed as part of below issues, hence closing it.
- https://github.com/rancher/rancher/issues/46918
- https://github.com/rancher/rancher/issues/46871

PR #3658, opened 2024-08-22
added streaming server addr configuration for cri-dockerd

vardhaman22:
issues:
* https://github.com/rancher/rke/issues/3656
* https://github.com/rancher/rancher/issues/46871

- added stream_server_address and stream_server_port in RancherKubernetesEngineConfig
- also added env variables to pass to kubelet container for setting stream-server-addr for cri-dockerd.

### DIFF
--- a/cluster/plan.go
+++ b/cluster/plan.go
@@ -64,8 +64,10 @@ const (
 
 	EncryptionProviderConfigArgument = "encryption-provider-config"
 
-	KubeletCRIDockerdNameEnv = "RKE_KUBELET_CRIDOCKERD"
-	KubeletDualStackNameEnv  = "RKE_KUBELET_CRIDOCKERD_DUALSTACK"
+	KubeletCRIDockerdNameEnv         = "RKE_KUBELET_CRIDOCKERD"
+	KubeletDualStackNameEnv          = "RKE_KUBELET_CRIDOCKERD_DUALSTACK"
+	CRIDockerdStreamServerAddressEnv = "CRIDOCKERD_STREAM_SERVER_ADDRESS"
+	CRIDockerdStreamServerPortEnv    = "CRIDOCKERD_STREAM_SERVER_PORT"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


cluster/plan.go
```diff
@@ -66,6 +66,8 @@ const (
 
 	KubeletCRIDockerdNameEnv = "RKE_KUBELET_CRIDOCKERD"
 	KubeletDualStackNameEnv  = "RKE_KUBELET_CRIDOCKERD_DUALSTACK"
+	StreamServerAddressEnv   = "STREAM_SERVER_ADDRESS"
+	StreamServerPortEnv      = "STREAM_SERVER_PORT"
```
kinarashah: we need to rename this to include CRIDOCKERD since it's a cluster level field

cluster/plan.go
```diff
@@ -617,7 +619,9 @@ func (c *Cluster) BuildKubeletProcess(host *hosts.Host, serviceOptions v3.Kubern
 	if c.IsCRIDockerdEnabled() {
 		Env = append(Env,
 			// Enable running cri-dockerd
-			fmt.Sprintf("%s=%s", KubeletCRIDockerdNameEnv, "true"))
+			fmt.Sprintf("%s=%s", KubeletCRIDockerdNameEnv, "true"),
+			fmt.Sprintf("%s=%s", CRIDockerdStreamServerAddressEnv, c.CRIDockerdStreamServerAddress),
+			fmt.Sprintf("%s=%s", CRIDockerdStreamServerPortEnv, c.CRIDockerdStreamServerPort))
```
kinarashah: Sorry for not catching this in the previous review, but this will cause all kubelets to restart since there'll be a change in node plan even when users haven't set a value. Could you add a check so that we add these env vars only when k8s version >= august patches?

kinarashah: The good thing is rke-tools will still change the port to default values when kubelet restarts, so there'll be a partial fix for older k8s versions. But I'm concerned about force restarting kubelets without users triggering cluster upgrade.

kinarashah (changes requested):

cluster/plan.go
vardhaman22: updated

cluster/plan.go
vardhaman22: updated

kinarashah (approved):

PR #3796, opened 2025-02-21
[release/v1.7] use sha256 for checksumming when k8s version >= 1.31.6

thatmidwesterncoder:
## Issue: https://github.com/rancher/rancher/issues/49373

tl:dr; we use md5 for checksumming part of plans, so in order to minimize unnecessary cluster upgrades we decided to just gate the `getStringChecksum(string)` function to require a kubernetes version and then use a better hashing algorithm over time. This way as time goes on the "insecure" hashing algorithm will not be used.

### DIFF
--- a/cluster/plan.go
+++ b/cluster/plan.go
@@ -3,9 +3,13 @@ package cluster
 import (
 	"context"
 	"crypto/md5"
+	"crypto/sha256"
 	b64 "encoding/base64"
+	"encoding/hex"
 	"encoding/json"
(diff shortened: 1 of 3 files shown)


jiaqiluo (approved):

HarrisonWAffel (approved):

kinarashah:
@thatmidwesterncoder Could you also forward port this to `release/v1.8` for Rancher v2.11?

Issue #3847, opened 2025-05-05
Error on adding a new node with etcd and controlplane flags


sp3ak3r:
Description
When attempting to add a new node to a Rancher cluster using the rancher/rancher-agent:v2.9.3 image with --etcd --controlplane --worker flags, the rke-etcd-port-listener container fails to start.

The command used to add the node (tokens and hosts hidden for security reasons):

```
sudo docker run -d --privileged --restart=unless-stopped --net=host \
-v /etc/kubernetes:/etc/kubernetes \
-v /var/run:/var/run \
rancher/rancher-agent:v2.9.3 \
--server https://rancher.net \
--token <TOKEN> \
--ca-checksum <CHECKSUM> \
--etcd --controlplane --worker
```
Error Message
In the Rancher UI, the following error appears:

```
Failed to start [rke-etcd-port-listener] container on host [10.50.0.2]: 
Error response from daemon: failed to set up container networking: 
driver failed programming external connectivity on endpoint rke-etcd-port-listener (0f48b341647b2ff34fa61cc8f67713105): 
failed to bind host port for 0.0.0.0:2379:172.17.0.2:1337/tcp: address already in use
Meaning of the error
```
The container fails to start because port 2379 is already in use on the host.

Docker Version

```
Docker Engine - Community
Version: 28.1.1
API version: 1.49
Go version: go1.23.8
Git commit: 4eba377
Built: Fri Apr 18 09:52:14 2025
OS/Arch: linux/amd64
```

dustin-groh-dev:
I mean, you kind of identified the issue already. Something else on the host is using port 2379, which is required to be available for etcd. Try running `sudo lsof -i :2379` to see the PID and the name of the process using that port.

sp3ak3r:
> I mean, you kind of identified the issue already. Something else on the host is using port 2379, which is required to be available for etcd. Try running `sudo lsof -i :2379` to see the PID and the name of the process using that port.

this is the etcd service on the current master nodes (

```
etcd      1468     root    8u  IPv6  11383      0t0  TCP *:2379 (LISTEN)
etcd      1468     root   14u  IPv6 161405      0t0  TCP MASTER-01:2379->MASTER-01:52176 (ESTABLISHED)
etcd      1468     root   19u  IPv6  18938      0t0  TCP MASTER-01:2379->MASTER-01:53846 (ESTABLISHED)
[...]

PR #3865, opened 2025-07-03
Updates deployListener() to contain docker output check compatable with Docker Enginer 28 and below

Faulkner-Ned:
For:
* https://github.com/rancher/rke/issues/3847
* https://github.com/rancher/rancher/issues/50352

**Docker Port Binding Error in Version 28.x Breaks ETCD and Control Plane Node Creation**

In version 28, Docker Engine changed the output format for port binding errors when an address is already in use.

**Docker version 27.1.2:**

`failed to bind port 0.0.0.0:22/tcp: Error starting userland proxy: listen tcp4 0.0.0.0:22: bind: address already in use.`

**Docker version 28.2.2:**

`failed to bind host port for 0.0.0.0:22:172.17.0.3:80/tcp: address already in use`

However, this breaks the creation of ETCD and Control Plane nodes due to [this line of code](https://github.com/rancher/rke/blob/release/v1.8/cluster/network.go#L908), which uses Docker’s error output to determine whether a service is already running. The new bind message format no longer starts with bind:, so the check fails.

This causes provisioning failures when attempting to spin up Control Plane or ETCD nodes on hosts running Docker 28.x:

`[Failed to start [rke-cp-port-listener] container on host [X.X.X.X]: Error response from daemon: failed to set up container networking: driver failed programming external connectivity on endpoint rke-etcd-port-listener `

`[Failed to start [rke-cp-port-listener] container on host [X.X.X.X]: Error response from daemon: failed to set up container networking: driver failed programming external connectivity on endpoint rke-etcd-port-listener `

**Currently, the only workarounds are**:

- Running rke up --disable-port-check (which doesn’t seem supported for Rancher-managed downstream clusters)
- Downgrading Docker Engine to version 27.x

✅ Proposed Fix

I've removed the "bind" prefix requirement in the error check, which makes the logic compatible with both Docker 28.x and earlier versions.

### DIFF
--- a/cluster/network.go
+++ b/cluster/network.go
@@ -905,7 +905,7 @@ func (c *Cluster) deployListener(ctx context.Context, host *hosts.Host, portList
 
 	logrus.Debugf("[network] Starting deployListener [%s] on host [%s]", containerName, host.Address)
 	if err := docker.DoRunContainer(ctx, host.DClient, imageCfg, hostCfg, containerName, host.Address, "network", c.PrivateRegistriesMap); err != nil {
-		if strings.Contains(err.Error(), "bind: address already in use") {
+		if strings.Contains(err.Error(), ": address already in use") {
 			logrus.Debugf("[network] Service is already up on host [%s]", host.Address)
 			return nil
 		}


cluster/network.go
```diff
@@ -905,7 +905,7 @@ func (c *Cluster) deployListener(ctx context.Context, host *hosts.Host, portList
 
 	logrus.Debugf("[network] Starting deployListener [%s] on host [%s]", containerName, host.Address)
 	if err := docker.DoRunContainer(ctx, host.DClient, imageCfg, hostCfg, containerName, host.Address, "network", c.PrivateRegistriesMap); err != nil {
-		if strings.Contains(err.Error(), "bind: address already in use") {
+		if strings.Contains(err.Error(), ": address already in use") {
```
snasovich, line 908: Is leaving `: ` significant here? It doesn't necessarily hurt but looks quite off.

```suggestion
		if strings.Contains(err.Error(), "address already in use") {
```

snasovich (commented):
LGTM, one minor nit that would be great to address if the comment is not off.

cluster/network.go
Faulkner-Ned: No significant impact & happy to remove colon if team prefers/think it's best!

I am still fairly new to Docker Engine v28 Spec, so wanted to be a bit cautious incase docker had decided to change another error message to now include "address already in use", so keep the `: ` to try and ensure we only catch the expected error and not a random one.

However just had another quick scan at the Docker Engine Code and I can't see any other ways for error messages using "address already in use" 👍🏻

snasovich: @Faulkner-Ned , agree - it's questionable what is the best option. I'll let other reviewers chime in on that.

jiaqiluo (approved):

PR #3888, opened 2025-09-24
[v1.8] upload rke extended images list on tag

kinarashah:
Artifacts under build/bin are already uploaded as artifacts: https://github.com/rancher/rke/blob/release/v1.8/.github/workflows/workflow.yaml#L109
https://github.com/rancher/rke/blob/release/v1.8/.github/workflows/workflow.yaml#L130

Creating new file rke-extended-life-images.txt to avoid uploading it manually. See https://github.com/rancher/rke/releases/download/v1.8.7/rke-extended-life-images.txt for reference.

### DIFF
--- a/scripts/create-imageslist.sh
+++ b/scripts/create-imageslist.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+# This script will create a txt file with RKE extended life images.
+set -e -x
+
+IMAGES_LIST_FILE="./build/bin/rke-extended-life-images.txt"
+
+# Ensure the build directory exists
+mkdir -p ./build/bin
+
+echo "Creating ${IMAGES_LIST_FILE}"
+
+# The command to generate a unique list of sorted images after removing noiro/aci images. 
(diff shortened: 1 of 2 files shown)


scripts/create-imageslist.sh
```diff
@@ -0,0 +1,24 @@
...
+# Ensure the build directory exists
+mkdir -p ./build/bin
+
+echo "Creating ${IMAGES_LIST_FILE}"
+
+# The command to generate a unique list of sorted images after removing noiro/aci images. 
+./bin/rke config --system-images --all | grep -Eo '^[^ ]+/.+' | grep -v '^noiro/' | sort | uniq > "$IMAGES_LIST_FILE"
+
+# Check if the command was successful
+if [ $? -ne 0 ]; then
```
jakefhyde: Won't `set -e` cause the above command to exit anyway?

kinarashah: Yep, looks redundant with `set -e` in the beginning. Picked the template from https://github.com/rancher/rke/pull/2843/files#diff-711d44b1e728c3a61cf4123cc4ccd56bdeadb212c8faf443913b3d72ccd0305a. Removed this check

mitulshah-suse (approved):
LGTM

